PR #18, opened 2020-10-26
Upgrading to version 0.4.0

BDonnot:
Adding the Gauss Seidel powerflow algorithm, and improve the documentation


BDonnot (commented):
This looks rather good, and tests pass

Issue #34, opened 2021-09-23
Compute the parameters of lines trafo (r, x, b, coefficients to put in the jacobian etc) only once 


BDonnot:
## Is your feature request related to a problem? Please describe.

This would allow for faster execution. 

Today these coefficients are computed each time a powerflow is ran. Actually they are even computed twice per powerflow. Once to create Ybus and another to compute flows on powerline. 

## Describe the solution you'd like

 See title 

## Describe alternatives you've considered

There is no alternative at the moment. We lose time to recompte them each time.

Issue #33, opened 2021-09-21
DC powerflow


BDonnot:
Make (and especially test) the DC powerflow in lightsim2grid

Issue #32, opened 2021-09-09
Being able to perform a "N-k" security analysis


BDonnot:
## Is your feature request related to a problem? Please describe.

It should be faster to compute directly the complete N-k security criterion "at once", or even in parallel c++ side, with lightsim2grid. 

## Describe the solution you'd like

Another interface that digest a list of "actions" - new topologies and / or injections - that would return the powerflow for each of them. 

## Describe alternatives you've considered

For now, the only thing to do is to do it python side.

## Additional context

   
This kind of computation is always performed in power system operation

Issue #26, opened 2021-04-14
Release the gil when computing the powerflow


BDonnot:
## Is your feature request related to a problem? Please describe.

To allow use of threading in python it would be nice to "release the gil" when making powerflows. This could be done rather easily with the ` py::call_guard<py::gil_scoped_release>()` 

Consider doing some benchmark for both single threaded and multi threaded example to see if the performances are worth it.

PR #39, opened 2021-12-17
Upgrade to version 0.6.0

BDonnot:
Breaking changes
--------------------------

- change the interface of the `newton_pf` function to reflect pandapower change in their
  latest version (arguments `ref` has been added). You can still use the old `newton_pf` function, with the
  old signature by importing `newtonpf_old` instead.
- `SecurityAnalysis` now also returns the active flows when calling `security_analysis.get_flows()`
- change the file names (python side) to be compliant with pep 8. You can no longer
  do things like `from lightsim2grid.LightSimBackend import LightSimBackend` change it to
  `from lightsim2grid import LightSimBackend` (preferred method)
- change the file names (python side) to be compliant with pep 8. You can no longer
  do things like `from lightsim2grid.initGridModel import init` change it to
  `from lightsim2grid.gridmodel import init` (preferred method) (same for `GridModel` class)

Bug Fixes
--------------

- a bug that lead to the wrong computation of the dc powerflow in case of `sn_mva != 1.` and phase shifters.
- bug preventing to use the NICSLU linear solver in the `GridModel`
- compilation warnings on clang (missing virtual destructor, unused variables, etc.)
- a bug in the `SecurityAnalysisCPP`: when it diverges for some contingencies, the others were not simulated properly.
- `LightSimBackend` now contains members for `shunts` and `***_theta` as it does for the other quantities. This improves the consistency, but most importantly
  fixes some bugs when used in earlier grid2op versions

New features
------------------
- possibility to compute the active flows using the `BaseMultiplePower` 
- possibility to change linear solver used when performing a DC solver
- possibility to make powerflow with distributed slack bus (only for newton raphson at the moment)
- access (read only) to the element of a lightsim2grid grid with the `get_XXX` (*eg* `get_loads()`) methods (see documentation)
- direct access to the solver used in the grid model python side
- unittest in circleci.
- [all kind of solvers based on different linear solvers (Eigen sparse LU, KLU or NICSLU) for Newton Raphson and
  DC approximation (9 solvers in total)

General improvments
--------------------------------
- use of `steady_clock` to retrieve the ellapse time c++ side
- refactoring of the c++ part to use template mecanism instead of inheritance for the
  Newton Raphson and DC solvers.
- `GridModel` now contains two different solvers, one for AC powerflow and one for DC powerflow.
- error message in the solver are now embedded in an Enum instead of being integers, for better readibility.
- error message when the powerflow diverge (error are read from c++ now)

Github issues resolved
-------------------------------
- #26 
- #32 
- #33 
- #34


BDonnot (commented):
File change looks good and tests pass

PR #41, opened 2022-02-01
newtonpf.py: choose solver based on distributed_slack parameter

rbolgaryn:
+Some small changes

* to make the behavior in lightsim2grid and pandapower consistent in terms of the J matrix, we choose the "standard" vs "single slack" version of the solver bnased on the parameter "distributed_slack".
* We transform the Ybus to csc if it is not yet csc. TODO for pandapower: define Ybus in csc in the first place if lightsim2grid is used (right now implemented via to_csc()).

### DIFF
--- a/lightsim2grid/newtonpf/newtonpf.py
+++ b/lightsim2grid/newtonpf/newtonpf.py
@@ -10,15 +10,16 @@ import warnings
 import numpy as np
 from scipy import sparse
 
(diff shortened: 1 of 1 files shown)


BDonnot:
Hello,

Can you do the pull request to the `bd-dev` branch instead ?

Thanks

Benjamin

PR #42, opened 2022-02-01
newtonpf.py: choose solver based on distributed_slack parameter

rbolgaryn:
+Some small changes

 * to make the behavior in lightsim2grid and pandapower consistent in terms of the J matrix, we choose the "standard" vs "single slack" version of the solver bnased on the parameter "distributed_slack".

 * We transform the Ybus to csc if it is not yet csc. TODO for pandapower: define Ybus in csc in the first place if lightsim2grid is used (right now implemented via to_csc()).

### DIFF
--- a/lightsim2grid/newtonpf/newtonpf.py
+++ b/lightsim2grid/newtonpf/newtonpf.py
@@ -307,7 +307,7 @@ def newtonpf_new(Ybus, Sbus, V0, ref, pv, pq, ppci, options):
 
     # initialize the solver
     # TODO have that in options maybe (can use GaussSeidel, and NR with KLU -faster- or SparseLU)
-    if options["distributed_slack"]:
+    if options.get("distributed_slack", False):
         solver = KLUSolver() if KLU_solver_available else SparseLUSolver()
     else:
         solver = KLUSolverSingleSlack() if KLU_solver_available else SparseLUSolverSingleSlack()
--- a/lightsim2grid/newtonpf/newtonpf.py
+++ b/lightsim2grid/newtonpf/newtonpf.py
@@ -10,15 +10,16 @@ import warnings
 import numpy as np
 from scipy import sparse
 
-from lightsim2grid_cpp import SparseLUSolver
-KLU_solver_available = False
+from lightsim2grid_cpp import SparseLUSolver, SparseLUSolverSingleSlack
+
 try:
-    from lightsim2grid_cpp import KLUSolver
+    from lightsim2grid_cpp import KLUSolver, KLUSolverSingleSlack
     KLU_solver_available = True
 except ImportError:
-    pass
+    KLU_solver_available = False
 
 _PP_VERSION_MAX = "2.7.0"
+
 
 def _isolate_slack_ids(Sbus, pv, pq):
     # extract the slack bus
@@ -290,14 +291,14 @@ def newtonpf_new(Ybus, Sbus, V0, ref, pv, pq, ppci, options):
         By default (and this cannot be changed at the moment), all buses in `ref` will be pv buses except the first one.
 
     """
-    max_it = options["max_iteration"]
-    tol = options['tolerance_mva']
-    bus = ppci['bus']
+    max_iteration = options["max_iteration"]
+    tolerance_pu = options['tolerance_mva']  # / ppci["baseMVA"]
 
     try:
-        from pandapower.pypower.idx_bus import SL_FAC  # lazy import for earlier pandapower version (without distributed slack)
-        slack_weights = bus[:, SL_FAC].astype(float)  ## contribution factors for distributed slack
-        slack_weights /= slack_weights.sum()
+        # lazy import for earlier pandapower version (without distributed slack):
+        from pandapower.pypower.idx_bus import SL_FAC
+        # contribution factors for distributed slack:
+        slack_weights = ppci['bus'][:, SL_FAC].astype(np.float64)
     except ImportError:
         # earlier version of pandapower
         warnings.warn("You are using a pandapower version that does not support distributed slack. We will attempt to "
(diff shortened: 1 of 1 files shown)


lightsim2grid/newtonpf/newtonpf.py
```diff
@@ -290,14 +291,14 @@ def newtonpf_new(Ybus, Sbus, V0, ref, pv, pq, ppci, options):
...
+    tolerance_pu = options['tolerance_mva']  # / ppci["baseMVA"]
 
     try:
-        from pandapower.pypower.idx_bus import SL_FAC  # lazy import for earlier pandapower version (without distributed slack)
-        slack_weights = bus[:, SL_FAC].astype(float)  ## contribution factors for distributed slack
-        slack_weights /= slack_weights.sum()
+        # lazy import for earlier pandapower version (without distributed slack):
+        from pandapower.pypower.idx_bus import SL_FAC
+        # contribution factors for distributed slack:
+        slack_weights = ppci['bus'][:, SL_FAC].astype(np.float64)
```
BDonnot, line 301: Is this a guarantee from pandapower that the slack_weights sum to 1 here ?
(it is expected that it does in the lightsim2grid solver)

lightsim2grid/newtonpf/newtonpf.py
```diff
@@ -306,14 +307,16 @@ def newtonpf_new(Ybus, Sbus, V0, ref, pv, pq, ppci, options):
...
     # initialize the solver
     # TODO have that in options maybe (can use GaussSeidel, and NR with KLU -faster- or SparseLU)
-    if KLU_solver_available:
-        solver = KLUSolver()
+    if options["distributed_slack"]:
+        solver = KLUSolver() if KLU_solver_available else SparseLUSolver()
     else:
-        solver = SparseLUSolver()
-    Ybus = sparse.csc_matrix(Ybus)
+        solver = KLUSolverSingleSlack() if KLU_solver_available else SparseLUSolverSingleSlack()
```
BDonnot, line 313: Looks good ! It's more consistent :-)

Will it work with earlier version of pandapower ? (maybe a check that `"distributed_slack" in options` should be added I think

lightsim2grid/newtonpf/newtonpf.py
```diff
@@ -306,14 +307,16 @@ def newtonpf_new(Ybus, Sbus, V0, ref, pv, pq, ppci, options):
...
+
+    if ~sparse.isspmatrix_csc(Ybus):
+        Ybus = sparse.csc_matrix(Ybus)
```
BDonnot, line 316: We had also a problem with the "canonical form" if I remember correctly. Is it checked here ?

BDonnot (commented):
Changes look really nice and will be a plus for lightsim2grid. Thanks

I had a few questions in some comments. Once answered i will merge this asap

Benjamin

lightsim2grid/newtonpf/newtonpf.py
rbolgaryn: Yes, we already divide by the sum in pandapower:

(build_gen.py) 

```python
    for subnet in subnets:
        subnet_gen_mask = np.isin(gen_buses, subnet)
        sum_slack_weights = np.sum(slack_weights_gen[subnet_gen_mask])
        if np.isclose(sum_slack_weights, 0):
            # ppc['gen'][subnet_gen_mask, SL_FAC] = 0
            raise ValueError("Distributed slack contribution factors in "
                             "island '%s' sum to zero." % str(subnet))
        elif sum_slack_weights < 0:
            raise ValueError("Distributed slack contribution factors in island '%s'" % str(subnet) +
                             " sum to negative value. Please correct the data.")
        else:
            # ppc['gen'][subnet_gen_mask, SL_FAC] /= sum_slack_weights
            slack_weights_gen /= sum_slack_weights
            buses, slack_weights_bus, _ = _sum_by_group(gen_buses[subnet_gen_mask], slack_weights_gen[subnet_gen_mask], slack_weights_gen[subnet_gen_mask])
            ppc['bus'][buses, SL_FAC_BUS] = slack_weights_bus

    # raise NotImplementedError if there are several separate zones for distributed slack:
    if not np.isclose(sum(ppc['bus'][:, SL_FAC_BUS]), 1):
        raise NotImplementedError("Distributed slack calculation is not implemented for several separate zones at once, "
                                  "please calculate the zones separately.")
```

lightsim2grid/newtonpf/newtonpf.py
rbolgaryn: I will modify the code to account for when distributed_slack is not in options

lightsim2grid/newtonpf/newtonpf.py
rbolgaryn: I will add the check

rbolgaryn: Do you think this is adequate, or do you expect that Ybus can also come from a different software?

rbolgaryn:
I noticed that if I uncomment the division by baseMVA then a test fails in pandapower, due to tolerances.
Do you accept tolerance in MVA in newton and translate it in pu internally?
Or do you consider the tolerance in a different way?

In pandapower tolerance is considered like this:

```python
def _check_for_convergence(F, tol):
    # calc infinity norm
    return linalg.norm(F, Inf) < tol
```

Can you please double-check?

lightsim2grid/newtonpf/newtonpf.py
BDonnot: I will add some warnings in the documentation and a check here that will raise an error if it's not in canonical form so it's fine like this

BDonnot:
> Can you please double-check?

I can confirm that inside the c++ solver nothing is done with respect to the tolerance. 

The c++ code looks like:

```cpp
bool BaseSolver::_check_for_convergence(const RealVect & F,
                                        real_type tol)
{
    const auto norm_inf = F.lpNorm<Eigen::Infinity>();
    bool res =  norm_inf  < tol;
    return res;
}
```

So it's exactly the same as the one you mentionned

PR #57, opened 2023-01-11
Upgrade to version 0.7.1

BDonnot:
Breaking changes
-------------------------
- drop support for numpy version < 1.20 (to be consistent with grid2op). If you are using numpy < 1.20 you should really consider upgrading for most recent versions.
- it does not appear to compile when using clang10 and python 3.7 on linux.

Bug fix
----------
- [FIXED] a compatibility issue with grid2op 1.7.2 (missing another backend attribute
  when the environment is copied) see https://github.com/rte-france/Grid2Op/issues/360
- [FIXED] now an error if thrown if the bus indexes in the pandapower grid are not contiguous
  or do not start at 0 (thanks @rbolgaryn for spotting this issue)

New features
-------------------
- [ADDED] automatic build for python 3.11
- [ADDED] support for numpy >= 1.24 (some deprecation *eg** np.str and np.bool are removed)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -175,7 +175,7 @@ jobs:
                     git submodule update
                     CC=clang make
                     CC=clang python setup.py build
-                    CC=clang python -m pip install -U .
+                    CC=clang python -m pip install .
     compile_clang11:
         executor: clang11
         resource_class: small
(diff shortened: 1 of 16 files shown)


BDonnot (commented):
looks good

Issue #92, opened 2024-11-04
Support for Shedding


DEUCE1957:
PandaPower can run a powerflow is 1 or more elements is not in service (in_service = False). This can for instance be used to represent load shedding as an action.

However, if we attempt to set the bus of a load to -1 with LightSim2Grid, this results in a GameOver under all circumstances. In the latest Pull Request for Grid2Op: [https://github.com/Grid2op/grid2op/pull/655](https://github.com/Grid2op/grid2op/pull/655) I have added support for the PandaPowerBackend to still run the powerflow if the user explicitly initiates the environment with gridop.make(..., allow_shedding=True). 

In order to use this with LightSim2Grid, the backend would need to be modified to make use of the new attribute 'self.allow_shedding'. It would need this method:
```python
set_shedding(self, allow_shedding:bool):
   self.shedding = allow_shedding
```

Edit: Digging some more it seems LightSim2Grid already has some methods for removing objects from the GridModel that aren't in_service. Not sure why it does not converge then though while PandaPower does.

PR #93, opened 2024-11-04
Add: set_shedding method to support load/generator shedding

DEUCE1957:
See Issue [92](https://github.com/Grid2op/lightsim2grid/issues/92), this PR add a method `set_shedding()` to the class lightsimBackend. It also changes 2 lines of code to not throw a BackendError if allow_shedding is True and a load or generator is disconnected. 

Note: May need to wait for main PR request to Grid2Op before approving it here.


BDonnot:
Hello,

I included a more generic ways to do it that would also work with legacy grid2op versions. So I will close this PR.

Thanks though :-)

PR #95, opened 2024-11-29
Fix issue with theta and add build for python 3.13

BDonnot:


BDonnot (commented):
looks good, let's see if all files are in the archive now

Issue #96, opened 2024-12-04
Solver benchmarking and pandapower connection for DC power flow


SteffenMeinecke:
Dear @BDonnot,
many thanks for all you implementations!

## Documentation issue description
I don't think that the values referenced below the two tables at https://lightsim2grid.readthedocs.io/en/latest/benchmarks.html#computation-time can be found in the tables, e.g.

* no solver with `~1200 steps each second`
* `“only” 70 for the default PandaPower Backend` (I find 46.3 in the table)
* ...

Consequently, I find it hard to follow the discussion on the timings of the solvers.
One more question: Is there a parameter, such as `tolerance_mva` in pandapower, to adjust the stopping criteria of the iteration algorithms? At the tables `Δ aor (amps)`, `Δ gen_p (MW)`, `Δ gen_q (MVAr)` are documented but it seems that these are output and not the parameter for the stopping criteria. Are the documented three values absolute or in difference to pandapower (since pandapower always has 0 and not something like 1e-8)?
One major point to add to the timing comparison, I think, is to mention the initialization applied. This effects the convergence bahavior and speed.

## Additional context
Did you compare the calculation time for DC power flows, as well? If there is again a massive speed-up, do you think about making it available for pandapower, as well?
While pandapower runs `Va = dcpf(B, Pbus, Va0, ref, pv, pq)`, cf. [link](https://github.com/e2nIEE/pandapower/blob/5f49bb0cbeea23bbbf172e6fd97bc026ecdb46a5/pandapower/pf/run_dc_pf.py#L79), lightsim2grid documentation types `converged = dc_solver.solve(Ybus, V0, Sbus, ref, slack_weights, pv, pq, max_it, tol)` which should be relatively compatible, I'd say.

BDonnot:
Hello,

Thanks for spotting this.

You are correct, there might be some inconsistencies between the written text (that has been written a while ago) and the tables, that I try to maintain as up to date as possible.
I'll try to fix thee text and maybe come up with something to "automatically" update the text with correct data

The impact of the tolerance is discussed in another benchmark that I did not upload apparently. For this one the tolerance is the "default" one which should be 1e-7 or 1e-9 pu I don't remember. I'll try to put it somewhere in the text too. Because once again you are correct this makes a difference 😊

The delta's a_or and others are the absolute error compared to the baseline (in this case pandapower) which totally explains why pandapower has 0 there, it's the baseline. It has nothing to do with stopping criteria.

[...]

PR #99, opened 2024-12-19
Update for pandapower v3.0

panos-xenos:
Hi @BDonnot, we are updating pandapower to v3.0 in which we are renaming/re-purposing the tap_phase_shifter parameter to tap_changer_type. We picked up an error when running our tests which is linked to your _aux_add_trafo.py lib. We propose a simple change in the code that works with both old and new implementation. We hope it's correct to merge in the master branch. Please let us know if you have any comments or suggestions.


BDonnot:
Hello,

Thanks for raising this issue. I am currently on holiday and did not had time to look at this before leaving. I'll check this when I'll get back :-)

Thanks

Benjamin

panos-xenos:
Hey @BDonnot, any updates on this pull request? Thanks!

BDonnot:
Hello,

I have been working on #96, which is also related to pandapower :-)

I'll try to have to have a look ASAP (hopefully tomorrow)

Best

Benjamin

panos-xenos:
Hi @BDonnot, have you had the chance to proceed with the review and approval of this PR? pandapower version 3.0 is about to be officially release and there is a dependence on this change. Otherwise, an error is raised when running a load flow using lightsim2grid solver due to the parameter name change from this version onwards.

BDonnot:
Hello,

I need to adress some bugs before I can make a release. This feature will be present in the next release. I will see if I can quickly do another release not to block you too much.

BDonnot:
I just made the release, it should be on pypi soon.

But I think this check should not happen, lightsim2grid can perfectly handle tap_changer.

Can you send me some grid files where lightsim2grid and pandapower does not have the same results with tap_changer ?

Thanks

panos-xenos:
> I just made the release, it should be on pypi soon.
> 
> But I think this check should not happen, lightsim2grid can perfectly handle tap_changer.
> 
> Can you send me some grid files where lightsim2grid and pandapower does not have the same results with tap_changer ?
> 
> Thanks

Many thanks for releasing!

Not sure where this limitation comes from to be honest and if it's still actually a limitation. Note however that this is not relevant to simple tap changers (changing voltage ratio), but to phase shifting transformers. Is there still a limitation in lightsim2grid when there is an angle shift in the output voltage vector of two or three-winding transformers?

Since I'm working with TSO data, I cannot share these network files for confidentiality reasons, so we would need to use generic test network cases.

BDonnot:
Hello

Lightsim2grid does not handle 3w trafos at the moment. It's more a "lack of time" rather than a true limitation. Adding it would be rather straightforward and not cause any major issues. But... I don't have the time :-/

For the voltage angle shift I am pretty sure lightsim2grid can handle it internally. If there is a problem it's maybe because the data is not read when the pypowsybl grid is processed and sent to lightsim2grid.
If you know of a public (I don't want data from other tso in this open source project 😉) test case with phase shifter then I can run it and with both pandapower and lightsim2grid and see 1) if there are issues and 2) whether these issues can be easily fixed

Thanks 😊

PR #107, opened 2025-02-21
add a mode to keep only the main cc when performing steps

BDonnot:
Refacto + some new options in the backend + improved detachment


BDonnot (commented):
ok

Issue #97, opened 2024-12-06
Can not copy a GymEnv environment when using a LightSim2Grid backend


EBoguslawski:
## Environment

   - Python version: `3.12`
   - Grid2op version: `1.10.5`
   - LightSim2Grid version: `0.9.2.post2`
   - System: `ubuntu`

## Bug description

Hello,

I create a `GymEnv` environment from a grid2op environment. If I use a `LightSimBackend` backend when creating the grid2op environment, then I obtain an error if I try to copy the `GymEnv` environment with `copy.deepcopy`.
It works on an older environment with older versions (python `3.8`, Grid2op `1.9.8`, LightSim2Grid `0.7.0.post1`)

Notes:
- In the l2rpn_baselines.PPO_SB3.utils.SB3Agent class, a copy.deepcopy is performed on the nn_kwargs argument during initialization, the latter containing a GymEnv in its keys / values (needed to train the agent). Consequently, it now raises this error.
- I cannot find any copy method for the `GymEnv` class
- I firstly put this issue on the Grid2op repo, but was asked to copy it here. So this issue is a replicate of this one: Grid2op/grid2op#672

### Code snippet

```python
import grid2op
from grid2op.gym_compat import GymEnv
from lightsim2grid import LightSimBackend
import copy

env = grid2op.make("l2rpn_case14_sandbox", backend=LightSimBackend())
env_gym = GymEnv(env)

# This line raises an error
copy.deepcopy(env_gym)
```

## Current output

```
/data/boguslawskieva/virtual_environments/venv_multi_agent/lib/python3.12/site-packages/lightsim2grid/gridmodel/from_pandapower/_aux_add_trafo.py:38: UserWarning: There were some Nan in the pp_net.trafo["tap_neutral"], they have been replaced by 0
  warnings.warn("There were some Nan in the pp_net.trafo[\"tap_neutral\"], they have been replaced by 0")
/data/boguslawskieva/virtual_environments/venv_multi_agent/lib/python3.12/site-packages/lightsim2grid/gridmodel/from_pandapower/_aux_add_trafo.py:46: UserWarning: There were some Nan in the pp_net.trafo["tap_step_percent"], they have been replaced by 0
  warnings.warn("There were some Nan in the pp_net.trafo[\"tap_step_percent\"], they have been replaced by 0")
[...]

Issue #36, opened 2021-10-13
Problem when deep copying a grid2op environment with lightsim2grid backend


BDonnot:
## Environment
 - Grid2op version: `1.6.2` (**eg** 1.4.0)
 - lightsim version: `0.5.4` (**eg** 0.5.0)
 - System: `ubuntu 20.04

## Bug description

The powerflow diverge when the environment is deep copied.

## How to reproduce

It can be reproduced by executing:

```python
import copy
from lightsim2grid import LightSimBackend
import grid2op
env = grid2op.make("l2rpn_case14_sandbox", test=True, backend=LightSimBackend())
obs = env.reset()
env_cpy = copy.deepcopy(env)
new_obs, reward, done, info = env.step(env.action_space())
assert not done
new_obs_2, reward_2, done_2, info_2 = env_cpy.step(env.action_space())
assert not done_2, f'powerflow diverge, error was: \n{info_2["exception"]}'
```

## Current output

```
/home/benjamin/Documents/grid2op_dev/grid2op/MakeEnv/Make.py:317: UserWarning: You are using a development environment. This environment is not intended for training agents. It might not be up to date and its primary use if for tests (hence the "test=True" you passed as argument). Use at your own risk.
  warnings.warn(_MAKE_DEV_ENV_WARN)
Traceback (most recent call last):
  File "test_deepcopy_env.py", line 11, in <module>
    assert not done_2, f'powerflow diverge, error was: \n{info_2["exception"]}'
AssertionError: powerflow diverge, error was: 
[RuntimeError('The slack bus is disconnected.')]

```

## Expected output

Nothing shoud be raised and the environment should work correctly

BDonnot:
The complete implementation of "deepcopy" is not really implemented at the moment, in the mean time, please use `env.copy()`  which is working as expected

PR #111, opened 2025-04-17
Ready for version 0.10.3

BDonnot:
WIth changelog

- [FIXED] remove deprecated use of numpy<2 function in LightSimBackend
- [FIXED] the "copy.deepcopy" of a lightsim2grid backend does not crash anymore 
  (see issues #36 and #97)
- [FIXED] an issue that could lead to a "segfault" if the "runpf" method of
  `LightSimBackend` was called first with is_dc=True and then with is_dc=False
- [IMPROVED] compat with grid2op 1.11.0
- [IMPROVED] now test proper compilation on clang 20 (was limited to clang 18 before)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,61 +5,28 @@ orbs:
 
 
 executors:
-    gcc_14:
+    gcc_latest:
         docker:
             - image: gcc:14
-    gcc_13:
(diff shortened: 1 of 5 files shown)


BDonnot (commented):
looks good

PR #139, opened 2026-06-19
Refacto nr 2

BDonnot:
Lots of fixes on the CHANGELOG, will be done once when merged into master


lightsim2grid/network/from_pypowsybl/_from_pypowsybl.py
```diff
@@ -56,6 +56,37 @@ def _aux_get_bus(vl_df, bus_df, first_bus_per_vl, el_type, df, conn_key="connect
...
+            continue
+        for el_id, bus_view_id in zip(df.index, df["bus_id"].values):
+            lookup.setdefault(el_id, bus_view_id)
+    missing = [rid for rid in regulated_ids if rid not in lookup]
+    if missing:
+        raise RuntimeError("Some voltage controllers regulate an element whose bus could "
+                           "not be resolved (unknown or disconnected regulated element): "
+                           f"{missing}. This is not supported at the moment.")
+    return np.array([lookup[rid] for rid in regulated_ids], dtype=object)
+
```
BDonnot, line 101: This is resolved once at init. But afterwards, if the "regulated element" of the original grid change bus in lightsim2grid it will "desynch" both grids. This should be at least stated in the doc and associated with a comment on the CHANGELOG (with a TODO also in the code).

BDonnot (commented):
ok

PR #140, opened 2026-06-25
N1 full compute

BDonnot:


lightsim2grid/network/from_pypowsybl/_olf_params.py
```diff
@@ -0,0 +1,254 @@
...
+    for mod in (_lf, getattr(_pp, "_pypowsybl", None)):
+        if mod is not None and hasattr(mod, name):
+            return getattr(mod, name)
+    return None
+
+
+def _olf_provider_params():
+    """Return ``{name: set(possible_values) or None}`` for OpenLoadFlow, or
+    ``None`` if the installed pypowsybl does not expose the metadata."""
+    global _OLF_PROVIDER_PARAMS
```
BDonnot, line 104: Need to find an alternative if possible

src/core/batch_algorithm/BaseBatchSolverSynch.cpp
```diff
@@ -25,10 +26,36 @@ bool BaseBatchSolverSynch::compute_one_powerflow(
...
+        _algo, _algo_controler, _nb_solved, _timer_solver,
+        Ybus, V, Sbus, slack_ids, slack_weights, bus_pv, bus_pq, max_iter, tol);
+}
+
+/**
+ V is modified at each call !
+ Explicit version: operates on the passed solver / control / accumulators so it
+ can run concurrently (one solver per thread). The member Bbus_ is read-only here.
+**/
+bool BaseBatchSolverSynch::compute_one_powerflow(
```
BDonnot, line 39: This function should be const now

src/core/batch_algorithm/BaseBatchSolverSynch.cpp
```diff
@@ -52,10 +81,47 @@ bool BaseBatchSolverSynch::compute_one_powerflow(
...
     if(conv){
-        V = _algo.get_V().array();
+        V = algo.get_V().array();
+    }
+    ++nb_solved;
+    timer_solver += algo.get_computation_time();
+    return conv;
+}
+
+bool BaseBatchSolverSynch::warmup_solver(
```
BDonnot, line 90: probably const too now

src/core/batch_algorithm/ContingencyAnalysis.cpp
```diff
@@ -46,6 +50,147 @@ bool ContingencyAnalysis::check_invertible(const Eigen::SparseMatrix<cplx_type>
...
+        for(const auto & masked : _li_masked) if(is_masked(masked, bus)) ++strand;
+        const real_type weight = (bus < slack_weights_.size()) ? slack_weights_(bus) : 0.;
+        const bool better = (best_strand < 0) ||
+                            (strand < best_strand) ||
+                            (strand == best_strand && weight > best_weight) ||
+                            (strand == best_strand && weight == best_weight && bus < best_bus);
+        if(better){ best_strand = strand; best_weight = weight; best_bus = bus; }
+    }
+
+    // 4) reorder slack_ids_me_ so the chosen reference is index 0 (the NR uses
```
BDonnot, line 363: Mark (if not already) a comment in the NR and DC to tell that they MUST keep the ref slack as the first slack otherwise this breaks (and I don't even want to start debugging it)

docs/security_analysis.rst
```diff
@@ -45,11 +45,108 @@ It can be used as:
 For now this relies on grid2op, but we could imagine a version of this class that can read
 to / from other data sources.
 
-.. note:: 
-    
-    A more advanced usage is given in the `examples\\security_analysis.py` 
+.. note::
+
+    A more advanced usage is given in the `examples\\security_analysis.py`
```
BDonnot: Maybe make a script benchmarking the nb_threads, from 1 to say 8 on a security analysis for case6515rte for example. Script in the benchmarks folder and results reported here

src/core/powerflow_algorithm/BaseDCAlgo.tpp
```diff
@@ -126,11 +134,45 @@ bool BaseDCAlgo<LinearSolver>::compute_pf_dc(const Eigen::SparseMatrix<real_type
...
+        for(int col = 0; col < masked_mat.outerSize(); ++col){
+            for(typename Eigen::SparseMatrix<real_type>::InnerIterator it(masked_mat, col); it; ++it){
+                const int row = static_cast<int>(it.row());
+                if(!masked_row[row]) continue;
+                it.valueRef() = (row == col) ? 1. : 0.;
+            }
+        }
+    }
+    // the system matrix actually handed to the linear solver
+    Eigen::SparseMatrix<real_type> & sys_mat = has_mask ? masked_mat : dcYbus_noslack_;
```
BDonnot, line 168: const ref here too maybe ?

BDonnot (commented):
minor changes

lightsim2grid/network/from_pypowsybl/_from_pypowsybl.py
```diff
@@ -100,6 +250,173 @@ def _aux_regulated_bus_view_ids(net, regulated_ids):
...
+      ``ActivePowerDistribution.run`` only ever considers
+      ``LfSynchronousNetwork.getBuses()``) -- **not** to any particular country:
+      ``ActivePowerDistribution.filterParticipatingBuses`` has no country logic at
+      all, and OLF's own ``slackBusCountryFilter`` is empty by default, so foreign
+      border-equivalent generators (*eg* a cross-border interconnection's
+      equivalent generator, which can carry a very large ``max_p``/weight) do
+      participate. An earlier version of this function restricted participants to
+      the country hosting the most buses; that was a guess, not something OLF
+      actually does, and it was found to materially skew which generators absorb
+      the slack mismatch on real multi-country RTE grids -- removed;
```
BDonnot: Try to avoid any references to RTE grid and just keep "real" grids. I note it here but there are other places like this

lightsim2grid/network/from_pypowsybl/_from_pypowsybl.py
```diff
@@ -56,6 +67,145 @@ def _aux_get_bus(vl_df, bus_df, first_bus_per_vl, el_type, df, conn_key="connect
...
 
 
+def _aux_dangling_lines_fictitious(net, sort_index):
+    """One fictitious 1-bus "substation" per dangling line, at the far
+    (boundary) end of its series impedance.
+
+    Real full grids never have any dangling line -- they only appear when
+    zooming into a sub-area with
+    ``network.reduce_by_ids_and_depths(..., with_boundary_lines=True)``
+    (*eg* in ``reduce_and_compare.py`` / ``validate_olf_vs_lightsim.py``),
```
BDonnot: Do not mention these scripts directly. They are not part of lightsim2grid (yet?)

lightsim2grid/network/from_pypowsybl/_from_pypowsybl.py
```diff
@@ -56,6 +67,145 @@ def _aux_get_bus(vl_df, bus_df, first_bus_per_vl, el_type, df, conn_key="connect
...
+    entirely on the connectable (network) side; the boundary side carries
+    none (same convention already used for transformers' single ``h``, see
+    ``trafo_h`` below).
+
+    Returns ``(bus_extra, vl_extra, df_dl)``; ``bus_extra``/``vl_extra`` are
+    ``None`` and ``df_dl`` is empty if the grid has no dangling line.
+    ``df_dl`` carries two extra columns, ``boundary_bus_id`` /
+    ``boundary_vl_id``, used later to attach the equivalent branch + load.
+    """
+    df_dl = net.get_dangling_lines()
```
BDonnot: If using net, and not net_pu the unit might be wrong.
It's best to rely on pypowsybl per unit system rather than to perform the per unit here (also valid for SVC)

docs/binary_serialization.rst
```diff
@@ -0,0 +1,175 @@
...
+    exhausting memory, or silently returning garbage) on ill-formed files: garbage or truncated files,
+    corrupted internal sizes (every count in the file is checked against the real file size *before*
+    anything is allocated), files that contain an object of a different type (*eg* loading a
+    ``LoadContainer`` file with ``StorageContainer.load_binary``), and files with unexpected trailing
+    bytes are all rejected. ``save_binary`` on the other hand is **atomic by default**: it writes to a
+    temporary file that only replaces the destination once fully written, so an interrupted save never
+    destroys a previously saved file. Pass ``atomic=False`` to write the destination directly instead
+    (marginally faster -- it skips one temporary file and rename -- but without that protection). Note that the format stores raw native data: files are meant to be written
+    and read by builds sharing the same data layout (same endianness, ``real_type``, ...) -- there is
+    no checksum nor cross-platform migration, pickle remains the portable format.
```
BDonnot: Not even pickle is "portable". There should be "alert" message (stronger than warning, not sure alert is the right rst block for this) saying 1) not to trust lightsim2grid binary data from untrusted sources (better to say it) and 2) lightsim2grid binary data should ideally be generated in the same machine as the one that will consume it, with the same lightsim2grid version

docs/comparison_with_pypowsybl.rst
```diff
@@ -129,6 +121,74 @@ The parameters used to compute the powerflow in these examples are:
...
+transformers, generators, loads, shunts, HVDC lines, ... -- with the exact same API and
+DataFrame shape as a solved pypowsybl :class:`pypowsybl.network.Network`, use
+:class:`lightsim2grid.network.LightsimResultNetwork`. It wraps a converged ``LSGrid``
+(built by ``init_from_pypowsybl``) and the pypowsybl network it was built from, and
+exposes ``get_buses`` / ``get_lines`` / ``get_2_windings_transformers`` /
+``get_generators`` / ``get_loads`` / ``get_shunt_compensators`` /
+``get_static_var_compensators`` / ``get_batteries`` / ``get_hvdc_lines`` /
+``get_vsc_converter_stations`` / ``get_lcc_converter_stations``, each returning a pandas
+DataFrame indexed by the pypowsybl element id, with pypowsybl's own column names and
+sign conventions (post-solve ``p`` / ``q`` in the load convention):
```
BDonnot, line 266: A warning block after this to tell that only powerflows results data (are actually read from lightsim2grid). 

Another one might also tell that this is slower than using raw lightsim2grid accessors

lightsim2grid/network/from_matpower/_mpc_to_powermodels.py
```diff
@@ -0,0 +1,117 @@
...
+            "shift": np.deg2rad(shift_degree),
+            "transformer": transformer,
+            "rate_a": float(branch[i, RATE_A]),
+            "br_status": int(branch[i, BR_STATUS]),
+        }
+    if n_weird_shift:
+        warnings.warn(f"{n_weird_shift} branch(es) have `TAP == 0` (so are treated as a plain "
+                      "powerline) but a non-zero `SHIFT`, which is not physically meaningful "
+                      "for a plain line. The `SHIFT` will be ignored for these branches.")
+
```
BDonnot: Can you also put the lines row ids ? (usefull debug information)

lightsim2grid/network/from_pypowsybl/_from_pypowsybl.py
```diff
@@ -56,6 +67,204 @@ def _aux_get_bus(vl_df, bus_df, first_bus_per_vl, el_type, df, conn_key="connect
...
+    Returns two lists-of-lists aligned to ``trafo_index``: the phase-shift sample
+    points ``alpha`` (in **radian**, matching lightsim2grid's ``shift_``) and the
+    matching r/x correction (**in percent**), read from the pypowsybl
+    phase-tap-changer steps. Inner lists are empty for transformers without a
+    phase-tap-changer.
+
+    pypowsybl exposes the transformer r / x at the *neutral* tap and only folds the
+    tap into ``rho`` / ``alpha``; the per-step r/x deltas (percent) are dropped. They
+    matter for phase-shifting transformers whose series impedance varies with the
+    shift (RTE PSTs: tens of MW of through-flow). On those grids ``r% == x%`` for
```
BDonnot: remove any reference of RTE and say instead "real powergrids" or "real data" or something along the lines

lightsim2grid/network/from_pypowsybl/_from_pypowsybl.py
```diff
@@ -777,7 +1507,10 @@ def _aux_hvdc_station_data(df_side):
                 gen_slack_ids_int.append(gen_id)
                 model.add_gen_slackbus(gen_id, 1. / nb_conn)    
     else:
-        raise RuntimeError("You need to provide at least one slack with `gen_slack_id` or `slack_bus_id`") 
+        # nothing provided and the default distributed slack found no participating
+        # generator: fall back to a single slack on the most-connected generator bus
+        bus_id, gen_id = model.assign_slack_to_most_connected()
+        gen_slack_ids_int = [gen_id]
     
     # TODO checks
```
BDonnot: This whole file should rather be split into smaller, easier to read files, like the other importers. For example "add_substations", "add_buses", "add_generators" etc.

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -0,0 +1,820 @@
...
+Bake PowSyBl Open Load Flow (OLF) outer-loop results into an IIDM network so
+that a subsequent *outer-loop-free* power flow -- in OLF itself or in
+lightsim2grid -- reproduces the converged state exactly.
+
+Why this exists
+---------------
+lightsim2grid solves a single power-flow problem: it does not run the discrete
+"outer loops" that OLF runs (distributed slack, reactive-limit PV<->PQ
+switching, transformer/shunt voltage control, phase control, ...). Those loops
+change the *inputs* of the power flow between inner solves: a generator that
```
BDonnot: > change the *inputs* of the power flow between inner solves:

Well it depends on the definition of "power flow" I would rather, more explicitly say "change the input of the inner solves (*eg* change the input of the inner Newton Raphson)" or something like this

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -0,0 +1,820 @@
...
+discrete position, the slack mismatch is spread over participating units, etc.
+
+If you take the raw network, run OLF *with* outer loops, and then hand the same
+raw network to lightsim2grid, the two engines disagree -- not because the
+solvers differ, but because they are solving different problems.
+
+``bake_outer_loops`` rewrites the network's input setpoints to the values the
+outer loops settled on, and disables the corresponding regulation, so the
+problem becomes a plain power flow. After baking, OLF (loop-free) and
+lightsim2grid agree to solver tolerance, and they keep agreeing through
```
BDonnot: After baking, OLF (loop-free) and lightsim2grid SHOULD agree

(please add the "should" it does not agree all the time)

lightsim2grid/network/from_pypowsybl/_olf_params.py
```diff
@@ -0,0 +1,390 @@
...
+                break
+        out.append(int(num) if num else 0)
+    return tuple(out)
+
+
+_PARAM_SIG = set(inspect.signature(_lf.Parameters.__init__).parameters)
+_PYPOWSYBL_VER = _ver_tuple(_pp.__version__)
+# OLF rejects an empty ``outerLoopNames`` value before 1.4.0 (and the parameter
+# does not exist at all in 1.0.0); only use the allow-list where it is accepted.
+_OUTERLOOP_EMPTY_OK = _PYPOWSYBL_VER >= (1, 4)
```
BDonnot: safer to use the "version.parse" of "from packaging import version" so that comparisons are correct (parsing package version can be complicated with all corners cases, like alpha, release candidate, post version, pre version etc. etc.)

lightsim2grid/network/from_pypowsybl/_olf_params.py
```diff
@@ -0,0 +1,390 @@
...
+
+# Outer loops with an inline (inner-Newton-Raphson) alternative: switching the
+# mode keeps the modeled control active while removing the separate outer-loop
+# pass. Works on every pypowsybl version, so these are always applied unless
+# the caller explicitly asks to ``keep`` the outer-loop-based mode.
+_INLINE_MODE = {
+    "TransformerVoltageControl": ("transformerVoltageControlMode", "WITH_GENERATOR_VOLTAGE_CONTROL"),
+    "ShuntVoltageControl": ("shuntVoltageControlMode", "WITH_GENERATOR_VOLTAGE_CONTROL"),
+    "PhaseControl": ("phaseShifterControlMode", "CONTINUOUS_WITH_DISCRETISATION"),
+}
```
BDonnot, line 152: It's unclear here if these modes are always turned on (which SHOULD NOT be done) or if the "baking the outer loops" is not turned on for these features, if these features have certain values (which SHOULD be done)

lightsim2grid/network/from_pypowsybl/_olf_params.py
```diff
@@ -0,0 +1,390 @@
...
+    out = _lf.Parameters()
+    for name in _PARAM_SIG:
+        if name in ("self", "provider_parameters"):
+            continue
+        try:
+            value = getattr(parameters, name)
+        except AttributeError:
+            continue
+        setattr(out, name, value)
+    out.provider_parameters = dict(parameters.provider_parameters)
```
BDonnot, line 190: the provider parameters should be deep copied too here

lightsim2grid/network/from_pypowsybl/_olf_params.py
```diff
@@ -0,0 +1,390 @@
...
+            possible = valid[name]
+            if possible is not None and value not in possible:
+                return  # value not accepted in this version
+        prov[name] = value
+
+    # 1. Inline-mode loops: fold into the inner NR loop unless kept as an
+    #    outer loop. Version-independent.
+    for loop, (pname, value) in _INLINE_MODE.items():
+        if loop not in keep:
+            put_provider(pname, value)
```
BDonnot, line 269: I don't understand the behaviour here and i'm not sure it's what it is supposed to be.

If the input parameters says something (eg TransformerVoltageControl) should be done in an outer loop, this function should not change that. No outer loop will run, so the TransformerVoltageControl will not be changed (tap of transformers will not be modified).
Changing the value means changing the problem solved, this might fall into another root, diverge or whatever else... Definitely NOT what we want.

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+    """pypowsybl-``Network``-shaped view of a solved lightsim2grid ``LSGrid``.
+
+    :param ls_grid: a grid built by ``init_from_pypowsybl(net, ...)`` and
+        already solved (``ac_pf``/``dc_pf`` converged).
+    :param net: the same pypowsybl network passed to that ``init()`` call.
+
+    Every ``get_*`` method mirrors its ``pypo.network.Network`` namesake:
+    it accepts an optional ``attributes`` list and returns a DataFrame
+    indexed by the pypowsybl element id, built lazily on first call and
+    cached afterwards.
```
BDonnot, line 59: It should be clear what types of pypowsybl elements are supported here.

For example dangling_lines are not, nor are three windings transformers etc.

And I would also appreciate, for all elements supported a list of the attributes accessible from this class. For example, only "p" and "q" are accessible for loads and really fed from lightsim2grid.

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+        if self._sub_names is None:
+            self._sub_names = [s.name for s in self._grid.get_substations()]
+        if sub_id < 0 or sub_id >= len(self._sub_names):
+            return None
+        return self._sub_names[sub_id]
+
+    # ------------------------------------------------------------------ #
+    # buses
+    # ------------------------------------------------------------------ #
+    def get_buses(self, attributes: Optional[List[str]] = None) -> pd.DataFrame:
```
BDonnot, line 171: list all available attributes and make the difference between the one coming from the original pypowsybl grid and the one read from lightsim2grid

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+        if mask_orig.any() and mask_new.any():
+            offset = res.loc[mask_new, "v_angle"].median() - orig_angle.loc[mask_orig].median()
+            res["v_angle"] = res["v_angle"] - offset
+
+        return res
+
+    # ------------------------------------------------------------------ #
+    # lines / transformers
+    # ------------------------------------------------------------------ #
+    def get_lines(self, attributes: Optional[List[str]] = None) -> pd.DataFrame:
```
BDonnot, line 245: list all available attributes and make the difference between the one coming from the original pypowsybl grid and the one read from lightsim2grid

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+        self._ensure_lines_trafos()
+        return self._maybe_select(self._cache["lines"], attributes)
+
+    def get_2_windings_transformers(self, attributes: Optional[List[str]] = None) -> pd.DataFrame:
```
BDonnot, line 256: list all available attributes and make the difference between the one coming from the original pypowsybl grid and the one read from lightsim2grid

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+                df.loc[br_id, "p1"], df.loc[br_id, "q1"], df.loc[br_id, "i1"] = p1, q1, i1
+                df.loc[br_id, "p2"], df.loc[br_id, "q2"], df.loc[br_id, "i2"] = p2, q2, i2
+                df.loc[br_id, "bus1_id"], df.loc[br_id, "bus2_id"] = bus1, bus2
+                df.loc[br_id, "voltage_level1_id"], df.loc[br_id, "voltage_level2_id"] = vl1, vl2
+                df.loc[br_id, "connected1"], df.loc[br_id, "connected2"] = True, True
+
+    # ------------------------------------------------------------------ #
+    # generators / loads / shunts / svc / batteries: one-sided elements
+    # ------------------------------------------------------------------ #
+    def get_generators(self, attributes: Optional[List[str]] = None) -> pd.DataFrame:
```
BDonnot, line 461: list all available attributes and make the difference between the one coming from the original pypowsybl grid and the one read from lightsim2grid

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+        if "generators" not in self._cache:
+            self._cache["generators"] = self._build_one_sided(self._grid.get_generators(), flip_sign=True)
+        return self._maybe_select(self._cache["generators"], attributes)
+
+    def get_loads(self, attributes: Optional[List[str]] = None) -> pd.DataFrame:
```
BDonnot, line 469: list all available attributes and make the difference between the one coming from the original pypowsybl grid and the one read from lightsim2grid

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+        if "loads" not in self._cache:
+            self._cache["loads"] = self._build_one_sided(self._grid.get_loads(), flip_sign=False)
+        return self._maybe_select(self._cache["loads"], attributes)
+
+    def get_shunt_compensators(self, attributes: Optional[List[str]] = None) -> pd.DataFrame:
```
BDonnot, line 477: list all available attributes and make the difference between the one coming from the original pypowsybl grid and the one read from lightsim2grid

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+        if "shunts" not in self._cache:
+            self._cache["shunts"] = self._build_one_sided(self._grid.get_shunts(), flip_sign=False)
+        return self._maybe_select(self._cache["shunts"], attributes)
+
+    def get_static_var_compensators(self, attributes: Optional[List[str]] = None) -> pd.DataFrame:
```
BDonnot, line 485: list all available attributes and make the difference between the one coming from the original pypowsybl grid and the one read from lightsim2grid

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+        if "svcs" not in self._cache:
+            self._cache["svcs"] = self._build_one_sided(self._grid.get_svcs(), flip_sign=True)
+        return self._maybe_select(self._cache["svcs"], attributes)
+
+    def get_batteries(self, attributes: Optional[List[str]] = None) -> pd.DataFrame:
```
BDonnot, line 495: list all available attributes and make the difference between the one coming from the original pypowsybl grid and the one read from lightsim2grid

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+                "p": sign * el.res_p_mw, "q": sign * el.res_q_mvar,
+                "bus_id": self._ls_bus_to_pypo(el.bus_id), "connected": el.connected,
+                "voltage_level_id": self._ls_sub_to_vl(el.voltage_level_id),
+            })
+        return self._records_to_frame(records, self._ONE_SIDED_COLUMNS)
+
+    # ------------------------------------------------------------------ #
+    # hvdc lines / converter stations
+    # ------------------------------------------------------------------ #
+    def get_hvdc_lines(self, attributes: Optional[List[str]] = None) -> pd.DataFrame:
```
BDonnot, line 519: list all available attributes and make the difference between the one coming from the original pypowsybl grid and the one read from lightsim2grid

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+        if "hvdc_lines" not in self._cache:
+            self._build_hvdc()
+        return self._maybe_select(self._cache["hvdc_lines"], attributes)
+
+    def get_vsc_converter_stations(self, attributes: Optional[List[str]] = None) -> pd.DataFrame:
```
BDonnot, line 531: list all available attributes and make the difference between the one coming from the original pypowsybl grid and the one read from lightsim2grid

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+        if "vsc_stations" not in self._cache:
+            self._build_hvdc()
+        return self._maybe_select(self._cache["vsc_stations"], attributes)
+
+    def get_lcc_converter_stations(self, attributes: Optional[List[str]] = None) -> pd.DataFrame:
```
BDonnot, line 540: list all available attributes and make the difference between the one coming from the original pypowsybl grid and the one read from lightsim2grid

lightsim2grid/tests/test_binary_serialization.py
```diff
@@ -0,0 +1,625 @@
...
+        self._aux_test_1side(grid1, grid2, "get_generators", test_results,
+                             add_attr_int=["is_slack", "voltage_regulator_on"],
+                             add_attr_float=["slack_weight", "target_vm_pu", "min_q_mvar", "max_q_mvar"])
+        self._aux_test_1side(grid1, grid2, "get_storages", test_results)
+        self._aux_test_1side(grid1, grid2, "get_shunts", test_results)
+
+    def test_save_load(self):
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore")
+            self.env = grid2op.make("l2rpn_idf_2023", test=True, backend=LightSimBackend())
```
BDonnot, line 129: It's probably a good idea to use the ieee case 14 (eg from pypowysbl - preferred- or from pandapower) and add to this grid (or make multiple grids if it's too complicated to have a grid that converges with everything at once): 

- 1 SVC
- 1 storage unit
- 1 HVDC LCC, with droop
- 1 HVDC LCC, no drrop
- 1 HVDC VSC
- 1 transformer with phase shift

The test there is slow (init a full grid2op environment with 118 substations) and does not cover every type of elements at all.

lightsim2grid/tests/test_bus_fusion_pypowsybl.py
```diff
@@ -0,0 +1,405 @@
...
+# See AUTHORS.txt
+# This Source Code Form is subject to the terms of the Mozilla Public License, version 2.0.
+# If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+# you can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
+# This file is part of LightSim2grid, LightSim2grid implements a c++ backend targeting the Grid2Op platform.
+
+"""`fuse_zero_impedance_branches` (from_pypowsybl `init()`).
+
+A real RTE grid was found to have a line with `r_pu = x_pu = 0.0`. lightsim2grid
```
BDonnot: remove "mention of RTE" here and be more generic "real grid were found to have XXX"

lightsim2grid/tests/test_gen_reactive_curve_swap.py
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) 2026, RTE (https://www.rte-france.com)
+# See AUTHORS.txt
+# This Source Code Form is subject to the terms of the Mozilla Public License, version 2.0.
+# If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+# you can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
+# This file is part of LightSim2grid, LightSim2grid implements a c++ backend targeting the Grid2Op platform.
+
+"""Malformed reactive capability curve data (min_q/max_q effectively "swapped" at
+one of the curve points) was found on a real RTE grid snapshot: pypowsybl's
```
BDonnot: remove reference to RTE here too (and if you find others please follow the same principles)

lightsim2grid/tests/test_hvdc_main_component.py
```diff
@@ -0,0 +1,79 @@
...
+
+An HVDC line bridges two AC grids that are NOT synchronous: ``get_graph`` adds no
+edge for it, so the connectivity BFS of ``consider_only_main_component`` really
+splits the grid into *synchronous* components. When the two converters land in
+different components only one of them is in the main (solved) one.
+
+OpenLoadFlow keeps that in-main converter as a fixed boundary injection (it still
+imports / exports its scheduled HVDC power). lightsim2grid used to deactivate the
+WHOLE line as soon as one side was out of the main component, silently dropping
+that injection (hundreds of MW on real RTE grids -- see HVDC_OLF_FINDINGS.md).
```
BDonnot: Same comment: don't mention RTE

lightsim2grid/tests/test_pickleable.py
```diff
@@ -182,28 +222,37 @@ class Struct:
...
     def test_pickle_substations(self):
         self._aux_test_pickle("get_substations", _compare_substations)
 
+    def test_pickle_sgens(self):
+        self._aux_test_pickle("get_static_generators", _compare_static_generators, check_old_version=False)
+
+    def test_pickle_hvdc(self):
+        self._aux_test_pickle("get_dclines", _compare_dclines, check_old_version=False)
+
+    def test_pickle_svcs(self):
```
BDonnot, line 286: missing SVC, HVDC (LCC with droop enabled), HVDC (LCC without droop), and HVDC (VSC converters)

The same strategy than the one for test_binary_serialization could be use: make a grid (eg based on case 14 with all that, or multiple grids each having a "fancy" feature) and reuse it here too for the tests.
These grid could also be "directly converted to cpp code" and put inside the cpp tests folder to test the same things cpp side directly (but also later to test different things on grids with exotic elements)

(when I say "missing" here, I mean: these elements are not in the original grid, currently this basically tests nothing)

lightsim2grid/tests/test_pst_tap_impedance.py
```diff
@@ -0,0 +1,112 @@
...
+# you can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
+# This file is part of LightSim2grid, LightSim2grid implements a c++ backend targeting the Grid2Op platform.
+
+"""Tap-changer step r/x/g/b correction when reading a transformer from pypowsybl.
+
+pypowsybl exposes the transformer r/x/g/b at the *neutral* tap and only folds the
+tap position into ``rho`` / ``alpha``. The per-step r/x/g/b deltas (in percent)
+must be applied on top, otherwise the (phase-shifting) transformer impedance --
+and hence its through-flow -- is wrong. On real RTE grids this caused tens of MW
```
BDonnot: remove RTE name

lightsim2grid/tests/test_pst_tap_impedance.py
```diff
@@ -0,0 +1,112 @@
...
+of disagreement with PowSyBl Open Load Flow (see HVDC_OLF_FINDINGS.md).
```
BDonnot: do not reference this local files here (HVDC_OLF_FINDINGS)

lightsim2grid/tests/test_remote_control_disconnected_target.py
```diff
@@ -0,0 +1,110 @@
+# Copyright (c) 2026, RTE (https://www.rte-france.com)
+# See AUTHORS.txt
+# This Source Code Form is subject to the terms of the Mozilla Public License, version 2.0.
+# If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+# you can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
+# This file is part of LightSim2grid, LightSim2grid implements a c++ backend targeting the Grid2Op platform.
+
+"""A *connected* generator (or SVC) can remotely regulate the voltage of a
+terminal that is itself disconnected (found on a real RTE grid snapshot: two
```
BDonnot: remove reference from RTE here too

lightsim2grid/contingencyAnalysis.py
```diff
@@ -24,6 +29,45 @@
...
+
+    .. note::
+        If ``converged`` is False, ``limit_violations`` contains exactly one ``LimitViolation``
+        with ``element_type == ViolationElementType.GRID`` and ``violation_type`` either
+        ``LimitViolationType.NOT_SIMULATED`` (a pre-check skipped this contingency, eg it splits
+        the grid, without ever invoking the solver) or ``LimitViolationType.DIVERGENCE`` (the
+        solver ran but did not converge).
+    """
+    element_ids: List[int]  #: branch ids (lines then trafos) disconnected by this contingency
+    contingency_name: Optional[str]  #: user-supplied name, see `add_single_contingency`
```
BDonnot, line 59: This is not what I had in mind here. I was mentionning the name of the elements disconnected. But that is a good addition. Keep it.

But, in case there are names on the lines / trafos, can you also add "element_names" field in this class that gives the names of the element disconnected (and not only their id)

src/bindings/python/binding_batch.cpp
```diff
@@ -9,11 +9,47 @@
...
+        .def_readonly("element_id", &LimitViolation::element_id,
+                      "grid-model bus id for BUS ; local (0-based, own type) line / trafo id otherwise ; "
+                      "unused (-1) for GRID")
+        .def_readonly("side", &LimitViolation::side, "1 or 2 for LINE / TRAFO ; unused (0) for BUS / GRID")
+        .def_readonly("violation_type", &LimitViolation::violation_type)
+        .def_readonly("value", &LimitViolation::value,
+                      "value reached ; unused (NaN) for NOT_SIMULATED / DIVERGENCE")
+        .def_readonly("limit", &LimitViolation::limit,
+                      "limit that was violated ; unused (NaN) for NOT_SIMULATED / DIVERGENCE")
+        .def_readonly("name", &LimitViolation::name,
```
BDonnot: If there is a violation on a bus voltage, the constraints comes from the substation / voltage level and in this case it should be this name reported (if this exists)

lightsim2grid/network/from_pypowsybl/_result_network.py
```diff
@@ -0,0 +1,491 @@
...
+
+        # a uniform angle offset across every bus is just a difference of
+        # reference-datum convention (lightsim2grid's slack bus needs not be
+        # at the same angle-datum as whatever `net` last held), not physical:
+        # remove it by aligning medians, mirroring the "offset-removed" angle
+        # metric already used by `_olf_compare.py::compare_baked`.
+        orig_angle = self._net.get_buses()["v_angle"]
+        mask_orig = np.isfinite(orig_angle.to_numpy()) & (orig_angle.abs() > 1e-6)
+        mask_new = np.isfinite(res["v_angle"].to_numpy()) & (res["v_angle"].abs() > 1e-6)
+        if mask_orig.any() and mask_new.any():
```
BDonnot: I noticed there is a method "lsgrid.get_reference_slack_bus()" I think instead of relying on the medians computation (fragile) as done here. Substracting the angle of the get_reference_slack_bus() in lsgrid and adding the angle of the original pypowysbl grid would be more robust I believe.

BDonnot (commented):
review done for everything except cpp / core. BUt it rather huge as is

PR #141, opened 2026-07-05
Add native from_matpower initializer for LSGrid

BDonnot:
## Summary
- Adds `lightsim2grid.network.init_from_matpower`, a native MATPOWER case loader that initializes `LSGrid` directly from the raw `bus` / `gen` / `branch` matrices, without ever constructing a pandapower network.
- Accepts a `.m` file (via the optional `matpowercaseframes` package), a `.mat` file (via the optional `scipy` package), or an already-parsed mpc dict/object.
- Unlike going through pandapower's own MATPOWER converter, this keeps every `mpc.gen` row as an independent generator, including several generators co-located on the same (reference) bus, which are assigned as an equal-weight distributed slack. This sidesteps a known bug in pandapower's converter that mishandles grids with multiple generators on the same bus.
- Adds `lightsim2grid[matpower]` optional dependency extra (`scipy`, `matpowercaseframes`).

## Test plan
- [x] Added `lightsim2grid/tests/test_init_from_matpower.py` (11 tests: dict-based conversion, multi-generator-per-bus handling, shunt sign convention, deactivated generators/buses, missing-slack error handling, `.m`/`.mat` file dispatch).
- [x] Built the C++ extension from scratch (`pip install --no-build-isolation -e .`) and ran the new test suite: 11/11 passed.
- [x] Verified `init_trafo` / `init_generators_full` / `add_gen_slackbus` / etc. call signatures against `src/core/LSGrid.hpp` and the existing `from_pandapower` loader for consistency.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BCzFCW6JGoAPesqmnRUcND)_

PR #142, opened 2026-07-05
Add init_from_pf_delta loader for the PFΔ benchmark dataset

BDonnot:
## Summary
- Adds `lightsim2grid.network.init_from_pf_delta`, a loader that builds an `LSGrid` from a single row of the PFΔ benchmark dataset (arXiv:2510.22048), a PowerModels.jl/MATPOWER-format dict with a solved AC power-flow state attached. Accepts either a parsed row (dict) or a path to its `.json` file.
- Implemented as a thin translation layer (`lightsim2grid/network/from_pf_delta/_pf_delta_to_mpc.py`) on top of #141's `init_from_matpower`: the PFΔ row's PowerModels dict is converted into a raw MATPOWER-shaped `bus`/`gen`/`branch` (and, if present, `dcline`) array set and delegated to it, reusing its bus-id remapping, line/transformer splitting, generator/load/shunt/HVDC conversion and slack-bus handling rather than duplicating any of it.
- Handles the PowerModels-specific conventions that don't exist in raw MATPOWER: re-encoding the `tap == 0` "this is a plain line" sentinel (PowerModels normalizes it to `1.0`), converting `shift` from radians (PowerModels) back to degrees (raw MATPOWER), and aggregating PowerModels' separate `load`/`shunt` dicts back onto bus rows.
- Real numeric ground truth (`vm`/`va`/`pf`/`qf`/`pt`/`qt`) for validation was not available from the actual PFΔ dataset (`huggingface.co` is blocked in the dev sandbox this was written in); instead the schema/unit conventions were verified directly against the `pfdelta` and `PowerModels.jl` source repos and the official PowerModels.jl network-data docs, and the test fixture is an independent pandapower solve of `case14` (with `trafo_model="pi"`, to match MATPOWER's convention) translated into the verified PFΔ schema.

## Test plan
- [x] Added `lightsim2grid/tests/test_LSGrid_pf_delta.py` (12 tests): dict/path-based conversion, element counts, missing-`"network"`-key error, non-zero-`g_fr` rejection, radians→degrees `shift` conversion, and a `dcline` translation suite (conversion, matpower loss formula, deactivated line) since PFΔ's own pglib-derived cases never contain one.
- [x] The key correctness test: builds the `LSGrid` from the case14 fixture, runs lightsim2grid's own AC (Newton-Raphson) powerflow, and asserts `vm`/`va`/`pf`/`qf`/`pt`/`qt` match the row's independently-pandapower-solved values to 1e-5 — this is what actually catches tap-side/shift-sign/degree-vs-radian/shunt-sign convention bugs.
- [x] Built the C++ extension from scratch (`pip install --no-build-isolation -e .`) and ran the full new + existing `from_matpower` suite: 29/29 passed (1 skipped: `matpowercaseframes` not installed).

https://claude.ai/code/session_01Biv9NfcmzaTPgaC25Qip1n


lightsim2grid/network/from_pf_delta/_pf_delta_to_mpc.py
```diff
@@ -0,0 +1,157 @@
...
+    Convert a PFΔ / PowerModels.jl ``network`` dict into a raw-MATPOWER-shaped dict
+    (``{"bus", "gen", "branch"}`` numpy arrays in matpower's own column layout, plus
+    ``"baseMVA"``) consumable by :func:`lightsim2grid.network.from_matpower.init`.
+
+    Row order is ``sorted(network[<elem>], key=int)`` for bus / gen / branch: this is a
+    deterministic, documented contract (not just an implementation detail), relied upon
+    by tests that need to map a built `LSGrid`'s element positions back to the original
+    PFΔ row's bus / gen / branch keys.
+    """
+    if network.get("storage"):
```
BDonnot: Why storage is not supported ? Lightsim2grid supports it fine.

lightsim2grid/network/from_pf_delta/_pf_delta_to_mpc.py
```diff
@@ -0,0 +1,157 @@
...
+        gen[i, MBASE] = g.get("mbase", base_mva)
+        gen[i, GEN_STATUS] = g.get("gen_status", 1)
+        gen[i, PMAX] = g["pmax"]
+        gen[i, PMIN] = g["pmin"]
+
+    branch = np.zeros((len(branch_keys), MIN_BRANCH_COLS))
+    for i, k in enumerate(branch_keys):
+        br = network["branch"][k]
+        if br.get("g_fr", 0.0) != 0.0 or br.get("g_to", 0.0) != 0.0:
+            raise RuntimeError(f'branch "{k}" has a non-zero "g_fr"/"g_to" (line '
```
BDonnot: This is the second time this seems to happen, first time with the storage. 
I would reverse the logic. Instead of having this loader delegates to matpower (more simple) I would have matpower loader delegates to this one (more feature complete)

lightsim2grid/network/from_pf_delta/_pf_delta_to_mpc.py
```diff
@@ -0,0 +1,157 @@
...
+    plain line" sentinel) to `tap = 1.0`, so this loader cannot just test `tap != 0`
+    like `from_matpower.get_branch_split` does on a raw mpc branch matrix.
+    """
+    return bool(branch.get("transformer", False)) or branch.get("tap", 1.0) != 1.0 or branch.get("shift", 0.0) != 0.0
+
+
+def network_to_mpc(network: dict) -> dict:
+    """
+    Convert a PFΔ / PowerModels.jl ``network`` dict into a raw-MATPOWER-shaped dict
+    (``{"bus", "gen", "branch"}`` numpy arrays in matpower's own column layout, plus
```
BDonnot: Missing dclines on the doc I think

lightsim2grid/network/from_pf_delta/initLSGrid.py
```diff
@@ -0,0 +1,66 @@
...
+# If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+# you can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
+# This file is part of LightSim2grid, LightSim2grid implements a c++ backend targeting the Grid2Op platform.
+
+"""
+Build a LSGrid from a single PFΔ (arXiv:2510.22048) dataset row: a PowerModels.jl-format
+dict (MATPOWER-derived, per-unit) with a solved AC power-flow state attached.
+
+This loader translates the row's PowerModels ``network`` dict into a raw-MATPOWER-shaped
```
BDonnot: As said above the opposite would be better as power models format is more feature rich.

BDonnot (commented):
The main issue here is that power models is more feature rich than matpower. I thing to support all its features (which should be good) the logic should be inverted.

The grid would be initialized in this loader from the more complete data. And reading a case from MPC should reuse this (and implement only what is supported by matpower).

Also, I wonder it would be better to name the module "from_powermodels" and inside to have the data "from_pfdelta" because the data format is really power models and not pf delta.

lightsim2grid/network/from_pf_delta/_pf_delta_to_mpc.py
BDonnot: Good catch — that was only because the intermediate raw-`mpc` array format this loader was routing through has no storage table at all, not because lightsim2grid can't represent it. Fixed in fd57050 by inverting the architecture (see the review-level comment): storage is now a first-class conversion in the new `from_powermodels` engine (`_aux_add_storage.py`), calling `model.init_storages` directly. Covered by `TestPFDeltaStorage` in `test_LSGrid_pf_delta.py`.

---

lightsim2grid/network/from_pf_delta/_pf_delta_to_mpc.py
BDonnot: Agreed and done in fd57050: added `lightsim2grid/network/from_powermodels/` as the shared, feature-complete engine (bus/branch/gen/load/shunt/storage/dcline, dict-based), and rewrote `from_matpower/initLSGrid.py` to only parse the raw mpc matrices and convert them *up* into a PowerModels-style dict (`_mpc_to_powermodels.py`) before delegating to `from_powermodels.init`. `from_matpower`'s own public behavior/signature is unchanged — verified against its existing 19-test suite, which passes unmodified.

---

lightsim2grid/network/from_pf_delta/_pf_delta_to_mpc.py
BDonnot: This whole restriction is gone in fd57050 — with the architecture reversed, `from_powermodels/_aux_add_dc_line.py` reads `network["dcline"]` fields directly (no round trip through matpower's raw dcline matrix), so there's no doc note needed here anymore.

---

lightsim2grid/network/from_pf_delta/initLSGrid.py
BDonnot: Done — this file (and this whole `from_pf_delta._pf_delta_to_mpc` module) is gone. `init_from_pf_delta` now lives at `lightsim2grid/network/from_powermodels/_from_pfdelta.py` and is a ~15-line wrapper that just unwraps `row["network"]` and calls `init_from_powermodels(network, ...)` directly — no more matpower round trip.

---

BDonnot:
Addressed in fd57050 — inverted the architecture as suggested:

- New `lightsim2grid/network/from_powermodels/` package: a feature-complete, dict-based engine (`init(network) -> LSGrid`) covering bus, branch (line/transformer split — lines now keep independent per-side `g_fr`/`b_fr` and `g_to`/`b_to` directly, no longer forced through matpower's single-value `BR_B`), gen, load, shunt, **storage** (new — matpower's raw format has no storage table at all, even though lightsim2grid supports it) and dcline (HVDC), plus distributed-slack handling.
- `from_matpower/initLSGrid.py` now only parses the raw `mpc` matrices and converts them *up* into a PowerModels-style dict (`_mpc_to_powermodels.py`, the verified inverse of PowerModels.jl's own matpower parser), then delegates to `from_powermodels.init`. Its own public behavior/signature is unchanged — verified against its existing 19-test suite, which passes unmodified.
- Renamed the module from `from_pf_delta` to `from_powermodels`, per your naming suggestion; the PFΔ-row-specific bit (unwrapping `row["network"]`) now lives in `from_powermodels/_from_pfdelta.py`, with `init_from_pf_delta` kept as the public entry-point name.
- Also exposes `lightsim2grid.network.init_from_powermodels` directly for anyone with a bare PowerModels dict.

Test suite grown accordingly: storage conversion/deactivation, asymmetric line conductance, and a direct check that `shift` round-trips through the `init_trafo` degrees/radians boundary (via `TrafoInfo.shift_rad`), replacing the now-obsolete "reject non-zero `g_fr`" test. Full suite (33 tests across `test_LSGrid_pf_delta.py` + `test_init_from_matpower.py`) passes.

---

PR #143, opened 2026-07-11
Fix out-of-bounds memory access from Python bindings + RAII for linear-solver memory

BDonnot:
## Summary

Two related C++ memory-handling improvements on top of `n1_full_compute`:

1. **Fix out-of-bounds memory access reachable from the Python API.** Several Python-exposed methods passed an element id or array straight to unchecked C++ indexing (`std::vector::operator[]` / Eigen `operator()` / `.col()`, whose bounds asserts are compiled out under `NDEBUG`). A negative id wraps to a huge `size_t`; an out-of-range id reads or writes past the container. A plain call such as `grid.deactivate_load(-1)` or `grid.change_ratio_trafo(9999, 1.0)` could corrupt or leak process memory.
2. **RAII for the built-in linear solvers' memory** (the follow-up requested in review): replace manual `new[]`/`delete[]` and hand-written frees with `std::unique_ptr`.

All checks reuse the codebase's own validation idioms (`GenericContainer::_check_in_range` / `check_size` / `.at()`), the same guards the safe `change_p/q/v` and bus-id paths already use.

## Memory-safety fixes

Out-of-range ids now raise `IndexError` and mis-shaped arrays raise `RuntimeError`, instead of corrupting memory.

Out-of-bounds **writes** (highest severity):
- `update_slack_weights_by_id` — user id array indexed a `nb_gen`-sized vector unchecked.
- `set_gen_regulated_bus` — unchecked write on `gen_id`; `bus_id` was also stored without range validation.
- `change_ratio_trafo` / `change_shift_trafo` / `change_shift_trafo_deg` — unchecked write on `trafo_id`.
- `set_*_pos_topo_vect` / `set_*_to_subid` — numpy array assigned into a fixed-size Eigen view with no size check.

Out-of-bounds **reads**:
- `deactivate_* / reactivate_* / change_bus_*` (loads, gens, sgens, storages, shunts, powerlines, trafos, dclines) — the wrapper ran the unchecked `_method` hook *before* the `_generic_*` validator. Fixed once in the public `OneSideContainer` wrappers, which also covers the two-sided containers (they delegate to those `final` methods first). Mirrors the guard already present in `GeneratorContainer::_change_bus`.
- `get_status_droop_hvdc` — the setter was guarded, the getter was not.
- `update_gens_p` / `update_loads_p` / etc. — `new_values` was indexed by `has_changed`'s length; a shorter `new_values` over-read.
- `TimeSeriesCPP.compute_Vs` — the `gen_p/sgen_p/load_p/load_q` matrix column/row counts are now validated against the grid element counts before `fill_SBus_*` indexes user columns by grid element ids.

By contrast, `change_p/q/v`, `get_bus_*`, the `change_bus_*` bus-id validation, and the contingency-analysis id path were already correctly guarded and are unchanged.

## Linear-solver RAII refactor

- `CKTSOLinearSolver` and `NICSLULinearSolver`: the CSC index buffers `ai_`/`ap_` now use `std::unique_ptr<T[]>` via `std::make_unique` instead of `new[]`/`delete[]`.
- `KLULinearSolver`: the `klu_symbolic`/`klu_numeric` handles now use `std::unique_ptr` with stateful custom deleters calling `klu_free_symbolic`/`klu_free_numeric`. The deleters hold a pointer to `klu_common`, so `common_` is declared before the two handles to guarantee it outlives them at destruction (reverse-declaration-order teardown).

Lifetimes are unchanged (buffers/handles are still released on `reset()` and destruction), but ownership is RAII-managed and the manual frees are gone. This also removes a latent leak in all three: re-running `analyze()`/`factorize()` without an intervening `reset()` previously overwrote the raw pointer/handle without freeing the old one.

## Verification

- Built the extension and ran targeted tests on `case14`: 20 negative cases (negative and too-large ids across every fixed method) now raise `IndexError`; array-length/shape mismatches raise `RuntimeError`; none segfault. `compute_Vs` rejects wrong column counts and mismatched rows. No regression: AC powerflow still converges, `change_bus` still persists the change, all valid calls succeed.
- KLU solver (default build) on `case118`: converges and matches the SparseLU reference to ~2e-15 after 50 topology-change cycles (repeated `analyze`/`factorize`/`reset`).
- CKTSO/NICSLU are compiled only when those proprietary vendor libraries are present (not in CI here), so their exact `unique_ptr<T[]>` pattern was compiled standalone and run clean under AddressSanitizer.

Changes are documented in `CHANGELOG.rst` (`[0.14.0]`).

https://claude.ai/code/session_017ptbLEZy5KUt4u8qkiRjmR

---
_Generated by [Claude Code](https://claude.ai/code/session_017ptbLEZy5KUt4u8qkiRjmR)_

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -115,6 +115,22 @@ TODO: integration test with pandapower (see `pandapower/contingency/contingency.
   built its working matrix from the grid model's own (never populated, hence empty) Ybus,
   causing out-of-bounds writes. It now uses the correctly indexed internal `Ybus_` and builds
   the required inputs on demand, so it works both before and after `compute()`.
+- [FIXED] several Python-exposed methods could read or write out of bounds (potential
+  segfault / memory corruption) when given an out-of-range or negative element id or a
+  mis-shaped array, because the id / array reached unchecked ``operator[]`` /
+  Eigen ``operator()`` / ``.col()`` (whose bounds asserts are compiled out in release
+  builds) with no prior validation. Out-of-range ids now raise ``IndexError`` and
+  mis-shaped arrays raise ``RuntimeError``, instead of corrupting memory. Affected:
+  ``deactivate_* / reactivate_* / change_bus_*`` (loads, gens, sgens, storages, shunts,
+  powerlines, trafos, dclines — the bounds check ran *after* the access),
+  ``update_slack_weights_by_id`` and ``set_gen_regulated_bus`` (out-of-bounds writes, the
+  latter also stored an unvalidated bus id), ``change_ratio_trafo`` /
+  ``change_shift_trafo`` / ``change_shift_trafo_deg`` (out-of-bounds writes),
+  ``get_status_droop_hvdc``, the grid2op fast-update path
+  (``update_gens_p`` / ``update_loads_p`` / ... when ``new_values`` is shorter than
+  ``has_changed``), ``set_*_pos_topo_vect`` / ``set_*_to_subid``, and
+  ``TimeSeriesCPP.compute_Vs`` (injection-matrix column / row counts are now validated
+  against the grid element counts).
 - [FIXED] pickling (or `save_binary`/`load_binary`) an `LSGrid` whose `_ls_to_orig`
   bus-mapping was set (which `init_from_pypowsybl` and `init_from_pandapower` both do)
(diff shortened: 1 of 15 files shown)


BDonnot (commented):
Ok but missing tests

BDonnot (commented):
Ok

PR #144, opened 2026-07-11
Harden binary serialization (save_binary/load_binary) against ill-formed input

BDonnot:
## What this PR does

Robustness improvements to the fast binary serialization path (`save_binary` / `load_binary`) introduced on `n1_full_compute`, so that ill-formed input (garbage, truncation, corrupted sizes, wrong object type) always raises a clean `RuntimeError` instead of crashing, over-allocating, or silently loading wrong data.

### Cross-version compatibility via a binary format version

- Compatibility is now decided by a dedicated `BINARY_FORMAT_VERSION` (uint32, stored in the file header, defined in `src/core/BinaryArchive.hpp`) instead of requiring the exact same lightsim2grid version.
- The format number is only bumped when the serialized layout actually changes, so all lightsim2grid releases sharing the same format number read each other's files. The writer's package version is still stored in the header, but only for diagnostics.
- Loading an unsupported format raises a `RuntimeError` naming both versions and both format numbers. A `SUPPORTED_BINARY_FORMATS` table leaves room for future migration code.
- Each `StateRes` definition carries a reminder comment to bump the format number on layout changes, and `LSGrid::fixup_binary_state` (auto-detected by `load_binary_generic`) keeps the pickle-shared `set_state` version check from rejecting cross-version binary loads. Pickle behaviour is unchanged.

### Validation of untrusted input

- Every count/length field read from a file is checked against the real file size **before** any allocation (overflow-safe): a corrupted count raises a clean `RuntimeError` instead of `MemoryError` or an actual multi-gigabyte allocation.
- The header now stores a type tag (`binary_type_tag()` on `LSGrid` and the 10 containers): loading a file into the wrong class is rejected. This was a real silent failure before: `LoadContainer` and `StorageContainer` have byte-identical `StateRes` layouts.
- Trailing bytes after the end of the data are rejected as corruption.
- Magic number bumped `LSB1` → `LSB2` (the old layout was never released, no migration needed).

### Atomic saves

`save_binary` now writes to `path + ".lsb_tmp"` and renames it over the destination only after a fully successful write (Windows-safe replace): an interrupted save (crash, full disk, ...) never destroys a previously saved file.

### Layout-change safety net

A reference file saved with format 1 is committed under `lightsim2grid/tests/binary_format_fixture/`; `TestBinaryLayoutUnchanged` loads it, checks known values and runs an AC powerflow on it. Changing any `StateRes` without bumping `BINARY_FORMAT_VERSION` makes this test fail with a message pointing at the constant to bump (a `regen` helper regenerates the fixture after a deliberate bump).

### Out of scope (noted as future work in code comments)

Payload checksum (bit flips inside numeric data that keep all counts consistent still load silently) and ABI/endianness descriptor.

## Testing

- `lightsim2grid/tests/test_binary_serialization.py`: 21/21 pass — round-trips, cross-version load, wrong-format, corrupted-count, wrong-type, trailing-bytes, atomicity, golden-file layout guard.
- `lightsim2grid/tests/test_pickleable.py`: 13/13 pass (pickle untouched).
- Docs (`docs/binary_serialization.rst`) and `CHANGELOG.rst` updated.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XjPTtnmQ7Yrf6jcEyfhqSN)_


src/core/element_container/SvcContainer.hpp
```diff
@@ -65,6 +65,8 @@ class LS2G_API SvcContainer : public OneSideContainer_PQ, public IteratorAdder<S
             REACTIVE_POWER = 2
         };
```
BDonnot, line 70: If this, or any other similar constant are modified, then the magic format number should also be increased. A comment would be nice (here and if there are other enum like this somewhere else) and maybe a test python side that the enum CPP side have the correct integer values (with a clear error message that if the test fail, the magic number for serialization should be increased)

src/core/BinaryArchive.hpp
```diff
@@ -15,6 +15,27 @@
...
+//    LoadContainer file into StorageContainer.load_binary, whose StateRes
+//    layouts are otherwise identical);
+//  - every count/length field read from the file is checked against the
+//    actual file size *before* any allocation, so a corrupted count cannot
+//    trigger a multi-gigabyte allocation;
+//  - after the state is fully read, the file must be fully consumed
+//    (trailing bytes are treated as corruption);
+//  - writes go to a temporary file that is atomically renamed over the
+//    destination only on success, so a crash / full disk mid-save never
+//    destroys a previously good file.
```
BDonnot: This feature should be optional. The core idea behind this feature is speed. So maybe add a flag during the binary saving. I am not sure about a good default. True would ensure most people use it with the automatic write, which is nice but will take longer. Or maybe 2 functions, one called save_binary_fast (no atomic copy) and one like today being atomic which is a good default

BDonnot (commented):
2 changes to perform

src/core/element_container/SvcContainer.hpp
BDonnot: Done in 0cf9def: all four serialized enums (`AlgorithmType`, `SvcContainer::RegulationMode`, `HvdcLineContainer::ConvertersMode`, `ConverterStationContainer::ConverterType`) now carry a "these integer values are serialized verbatim — renumbering requires bumping BINARY_FORMAT_VERSION" comment, the three container enums are exposed to python, and a new python test (`TestSerializedEnumValues`) pins all four value sets and fails with an explicit "bump BINARY_FORMAT_VERSION in src/core/BinaryArchive.hpp" message.

---

src/core/BinaryArchive.hpp
BDonnot: Done in 0cf9def: `save_binary(path, atomic=True)` — atomic stays the default, and `atomic=False` writes the destination directly (no temporary file, no rename). Note the overhead of the atomic path is only one extra rename syscall (the data is written once either way), so the fast path mostly matters for many small files on slow filesystems. Both variants are tested (`test_atomic_save`, `test_non_atomic_save`).

---

BDonnot (commented):
Ok

PR #145, opened 2026-07-12
Add memory-hardening CI: ASan+UBSan and debug-assertions jobs (+ bugs they found)

BDonnot:
## What this PR does

Adds CI that can spot memory issues (out-of-bounds, undefined behaviour, allocator misuse) that the functional suites on CircleCI cannot see — plus fixes for the real bugs this tooling found on its very first run.

**Note:** was originally stacked on #144 (the corruption-sweep test relies on the bounds-checked binary reader introduced there); now that #144 is merged, the branch has been rebased onto `n1_full_compute` and the diff is only the CI work + UB fixes. First run of the new workflow was green: [Sanitizers run #1](https://github.com/Grid2op/lightsim2grid/actions/runs/29168080768) (ASan+UBSan: 13 min, debug-asserts: 7 min).

### New workflow `.github/workflows/sanitizers.yml` (GitHub Actions, on every push)

- **ASan + UBSan job**: builds with `__SANITIZE=1` (new CMake option, same env-var pattern as `__O3_OPTIM`: `-fsanitize=address,undefined -fno-omit-frame-pointer -g -O1` on `lightsim2grid_core` and `lightsim2grid_cpp`), then runs the binary serialization, solver control, time series and contingency/security analysis test modules (302 tests) under the sanitizers. Two runtime subtleties are handled and documented in the workflow: `LD_PRELOAD` must include **libstdc++ along with libasan** (python does not link libstdc++, and without it ASan's `__cxa_throw` interceptor aborts on the first C++ exception), and `ASAN_OPTIONS=detect_leaks=0` (CPython's allocator makes leak reports pure noise).
- **Debug-assertions job**: builds with `__DEBUG_ASSERTS=1` → `-UNDEBUG -D_GLIBCXX_ASSERTIONS`, re-enabling the **Eigen bounds assertions** that Release builds silence with `NDEBUG` (verified: Eigen's assert strings are present in this build, absent in a normal one), plus the ABI-safe libstdc++ container checks. Runs the solver-heavy modules (`test_ACPF`, `test_DCPF`, `test_solver_control`) — this catches logically-wrong-but-in-bounds Eigen indexing that ASan cannot see.
- Trigger uses `branches: '**'` — a single `'*'` does not match `/` in Actions globs, so topic branches like `foo/bar` would silently never run (this also affects `main.yml`, left untouched here).

### New corruption-sweep test (deterministic mini-fuzzer)

`TestCorruptionSweep` in `test_binary_serialization.py`: takes a valid saved grid and corrupts it at **every byte offset** (single-byte flips, 8-byte `0xFF` stamps, truncation at every length — ~14k loads, a few seconds). Each load must either succeed or raise a clean `RuntimeError`; a segfault, `MemoryError` or any other exception type fails the test. It also runs in the existing CircleCI full suite, and is most potent inside the ASan job.

### Bugs found by this tooling and fixed here

- **50 sites** of `Eigen::Map(&vec[0], vec.size())` across the element containers: undefined behaviour whenever the vector is empty (flagged by UBSan in `SubstationContainer::set_state`; also reachable through pickle). Fixed to `vec.data()`.
- `BinaryArchive` deserialized `bool` as a raw byte: a corrupted byte yields a `bool` holding 255, and merely loading such a value is UB (7 sites flagged). Bools are now normalized through `uint8_t` (same 1-byte on-disk size, no format change).
- Corrupted header strings flowed un-escaped into exception messages, so pybind11 raised `UnicodeDecodeError` instead of the intended `RuntimeError`. Messages now escape non-printable bytes and are length-capped.

## Testing

- Local run of the exact ASan job commands: 302 tests OK under ASan+UBSan, no reports (7 min 18 s).
- Local debug-asserts build: 229 solver tests OK; differential `strings` check proves the Eigen assertions are compiled in.
- After the rebase onto the merged #144: 29/29 binary-serialization tests pass locally.
- `CHANGELOG.rst` updated.

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -1,0 +1,99 @@
+# Memory-hardening CI: builds the C++ core + bindings with instrumentation
+# that the release wheels do not carry, and runs the test modules most likely
+# to surface memory issues (out-of-bounds, use-after-free, UB, silenced
+# assertions). Complements the functional suites on CircleCI, which run
+# against plain optimized builds.
+
+name: Sanitizers
+
+on:
+  push:
+    branches:
+      - '*'
+    tags:
+      - 'v*.*.*'
+
+jobs:
+  asan_ubsan:
+    # AddressSanitizer + UndefinedBehaviorSanitizer over the untrusted-input
+    # path (binary serialization) and the heaviest C++ compute paths (solver
+    # control, time series, contingency analysis).
+    name: ASan + UBSan tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install python dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -r requirements_compile_ci.txt
+          python -m pip install grid2op scipy pandapower
+
+      - name: Build lightsim2grid with ASan + UBSan
+        # __SANITIZE=1 is picked up by src/core/CMakeLists.txt and
+        # src/bindings/python/CMakeLists.txt: -fsanitize=address,undefined
+        # on both lightsim2grid_core and lightsim2grid_cpp
+        run: __SANITIZE=1 python -m pip install -v . --no-build-isolation
+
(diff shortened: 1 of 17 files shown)


BDonnot (commented):
Ok

PR #146, opened 2026-07-12
Validate powerflow inputs (ac_pf / dc_pf / every solver's compute_pf)

BDonnot:
## What this PR does

Third robustness surface after the binary files (#144) and the element-mutator bindings (#143): the powerflow entry points (`LSGrid.ac_pf` / `LSGrid.dc_pf` and every solver's `compute_pf` / `solve`) now reject ill-formed inputs with a clean python exception instead of reaching raw Eigen indexing (out-of-bounds reads/writes in Release builds, where Eigen's asserts are compiled out) or silently computing a wrong system.

### Shared validation: `BaseAlgo::check_pf_inputs` / `check_iter_tol`

One implementation, called at the top of every solver entry point — Newton-Raphson (multi and single slack), fast-decoupled, Gauss-Seidel / Gauss-Seidel synch (**which had no validation at all**), and the DC `compute_pf_dc` (same). This also implements the long-standing `TODO DEBUG MODE` checks of `BaseFDPFAlgo.tpp`, unconditionally: one O(n) pass per call, negligible next to the solve.

- **`RuntimeError`**: non-square `Ybus`; `V` / `Sbus` (or `Pbus`) / `slack_weights` not matching the size of `Ybus` (the slack weights are indexed *by bus id*, so they must be of size n); empty `slack_ids`; `max_iter <= 0`; `tol` not finite or not strictly positive (NaN / inf / 0 / negative); a bus listed in more than one of slack/pv/pq or twice in the same array (a silently-wrong system otherwise).
- **`IndexError`**: any out-of-range or negative bus id in `pv` / `pq` / `slack_ids` (same `std::out_of_range` convention as #143).

`LSGrid.ac_pf` / `dc_pf` additionally validate `max_iter` and `tol` on top of the existing `Vinit` size check (`dc_pf` ignores them numerically, but nonsensical values still indicate a bug at the call site). Wrong python argument *types* already raise `TypeError` via pybind11 — now pinned by tests.

### Segfault found by the new tests and fixed here

A standalone fast-decoupled solver (eg `FDPF_XB_SparseLU()` constructed directly from python, never attached to an `LSGrid`) dereferenced a **null pointer** in `fillBp_Bpp` on any `compute_pf` call — all four FDPF variants crashed the interpreter. Unlike NR / Gauss-Seidel, FDPF cannot work from `Ybus` alone (B'/B'' are built from the grid data), so it now raises a `RuntimeError` explaining that FDPF must be used through `LSGrid` (`change_algorithm(...)` then `ac_pf(...)`).

### Tests: `lightsim2grid/tests/test_pf_input_robustness.py`

Same pandapower-only fixture style as `test_LSGrid_out_of_bounds.py` (no grid2op needed). 16 tests / ~300 subtests:

- `TestAcDcPfInputs` (`ac_pf` and `dc_pf`): Vinit too short / too long / empty; `max_iter` 0 / negative; `tol` 0 / negative / NaN / ±inf; wrong types (string V, float max_iter, complex tol, ...); valid calls still converge.
- `TestSolverComputePfInputs`, for **every AC solver class available in the build** (NR / NRSing / FDPF XB-BX across SparseLU/KLU/NICSLU/CKTSO, GaussSeidel, GaussSeidelSynch): non-square Ybus, Sbus / V / slack_weights size mismatches, out-of-bounds and negative ids in pv / pq / slack_ids, empty slack, pv∩pq overlap, duplicate ids, slack-also-pv, bad max_iter / tol — plus a valid-input sanity check (which is what caught the FDPF segfault).

Follow-up once #145 merges: add `test_pf_input_robustness` to the ASan job's module list in `sanitizers.yml` (one line).

## Testing

- New module: 16/16 tests pass.
- Regression: `test_ACPF`, `test_DCPF`, `test_solver_control`, `test_GaussSeidelSolver`, `test_LSGrid_out_of_bounds` — 175 passed, nothing legitimate rejected.
- ASan + UBSan build (env flags): the new module runs clean under sanitizers.
- `CHANGELOG.rst` updated.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XjPTtnmQ7Yrf6jcEyfhqSN)_

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -365,14 +365,17 @@ TODO: integration test with pandapower (see `pandapower/contingency/contingency.
   comparisons against pickle (the speed up grows with grid size: up to ~17x faster to
   write and ~8x faster to read than pickle on grids with ~9000 buses).
 - [IMPROVED] robustness of the powerflow entry points against ill-formed input:
-  `LSGrid.ac_pf` / `LSGrid.dc_pf` now validate `max_iter` (> 0) and `tol`
+  `LSGrid.ac_pf` / `LSGrid.dc_pf` now validate `max_iter` (>= 0) and `tol`
   (finite and > 0) in addition to the size of `Vinit`, and the python-facing
   `Solver.compute_pf` / `Solver.solve` (Newton-Raphson, single-slack,
   fast-decoupled, Gauss-Seidel -- shared `BaseAlgo::check_pf_inputs`, bound
   via the new `BaseAlgo::compute_pf_with_input_validation`) now validate
   their inputs before touching them: non-square `Ybus`, `V` / `Sbus` /
-  `slack_weights` not matching the size of `Ybus`, empty `slack_ids` and
-  non-positive / non-finite `max_iter` / `tol` raise `RuntimeError`;
+  `slack_weights` not matching the size of `Ybus`, empty `slack_ids`,
+  a negative `max_iter` and a non-positive / non-finite `tol` raise
+  `RuntimeError` (`max_iter=0` is accepted: every solver builds its initial
+  state / first Jacobian before iterating, so it is a well-defined "return
+  the pre-iteration state" call, used by internal tests);
   out-of-range or negative bus ids in `pv` / `pq` / `slack_ids` raise
   `IndexError` (previously they reached raw Eigen indexing: out-of-bounds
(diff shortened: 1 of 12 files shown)


BDonnot (commented):
Ok

PR #148, opened 2026-07-14
Eigen unnecessary-copy audit: hot-path fixes, compute_pf/compute_pf_dc Eigen::Ref, init consistency, clang-tidy CI

BDonnot:
## Summary

Audit and fix pass for unnecessary Eigen copies (`const T&`/`T&` vs `Eigen::Ref`), plus two real bugs found along the way and a new report-only clang-tidy CI check.

- **Hot path** (every NR/Gauss-Seidel/FDPF iteration): `BaseFDPFAlgo` solves the P-iteration in place instead of copying; `NRSystem::mismatch()` caches its zero `dx` vector instead of reallocating every call; Gauss-Seidel's `one_iter()` and `BaseAlgo::isin()` take `Eigen::Ref` instead of concrete/by-value types; element containers' `compute_results`/`_get_amps` chain now takes `bus_vn_kv`/`p`/`q`/`v` as `Eigen::Ref<const RealVect>`, removing 9+6 full-vector copies per solve.
- **Batch/time-series layer**: `TimeSeries::fill_SBus_real/imag` take `Eigen::Ref<const RealMat>`; two ternaries (`ContingencyAnalysis`, `BaseBatchSolverSynch`'s DC path) that unconditionally copied `slack_weights_`/`Pbus_` even in the common no-op branch are now split into branches.
- **`BaseAlgo::compute_pf`/`compute_pf_dc`** (`[BREAKING]` for C++ solver plugins): `Sbus`/`Pbus`/`slack_weights` now flow as `Eigen::Ref` through the entire solver interface (`BaseAlgo`, `NRAlgo`, `BaseDCAlgo`, `BaseFDPFAlgo`, `GaussSeidelAlgo`, `AlgorithmSelector`, the example plugin, and `BaseBatchSolverSynch::compute_one_powerflow`), closing the batch layer's per-timestep `_Sbuses.row(i)` copy. Along the way, fixed a latent dangling-pointer hazard: `NRSystem::update_state` used to cache `Sbus_ptr_ = &Sbus`; with `Sbus` now an `Eigen::Ref` (a function-local wrapper), that would dangle the instant `update_state()` returns. Replaced with a data-pointer + size pair (`Eigen::Map` reconstructed on demand), matching the existing `Ybus_ptr_` contract. Verified clean under a full AddressSanitizer + UBSan run.
- **Two real bugs fixed**, not just inefficiencies:
  - `LineContainer::dc_x_tau_shift()` returned an `Eigen::Ref` bound to a destroyed temporary (dangling, currently harmless by luck).
  - `get_ptdf`/`get_ptdf_solver`/`get_lodf`/`get_Bf`/`get_Bf_solver` were bound in Python with `return_value_policy::reference` despite returning **by value** — wraps a numpy array around freed stack memory. Switched to the default (copying) policy.
- **`init`/`init_full`/`init_osc_pq`/`init_bus`** across every element container and the `LSGrid`/`SubstationContainer` wrappers now take `Eigen::Ref` too — not hot path (construction-time only), done for consistency with the rest of the codebase and for zero-copy numpy binding at grid-build time.
- **Documented, not fixed**: `BaseBatchSolverSynch::compute_amps_flows`/`compute_active_power_flows`'s per-element `CplxVect(_voltages.col(...))` construction isn't a simple reference-type fix (`_voltages` is RowMajor, so a column isn't contiguous; the surrounding ternary also needs a common type with `Zero()`) — recorded as a `CHANGELOG.rst [TODO]` and in-code comment instead of silently left.
- **New**: `.github/workflows/clang-tidy.yml` + `.clang-tidy`, running `performance-*`/`readability-const-return-type` over `src/core` on `workflow_dispatch`, push to `master`, and version tags. Report-only for now (posts to job summary + artifact, doesn't fail the job) — validated locally, including a documented false-positive class (`performance-unnecessary-value-param` on by-value `Eigen::Ref` params, which is the established idiom here).

## Test plan

- [x] Catch2 C++ unit test suite: 43/43 passing after every commit
- [x] Full clean rebuild from scratch: 43/43 passing
- [x] AddressSanitizer + UBSan run (Debug build) after the `compute_pf`/`compute_pf_dc` and `init`/`init_full` changes: 43/43 passing, no use-after-free/UB detected
- [x] Every `src/bindings/python/binding_*.cpp` syntax-checked clean against pybind11 (full wheel build not attempted in this sandbox — no network-installed python solver deps)
- [x] `clang-tidy.yml` validated locally: config auto-discovery, header filter scoping, and a representative run against `src/core` all confirmed working

---
_Generated by [Claude Code](https://claude.ai/code/session_01QVRtwmWk8KSm3HH3msoKqr)_


BDonnot (commented):
Ok

PR #149, opened 2026-07-14
Compiler-friendliness pass: final/override hygiene, const-correctness, [[nodiscard]]

BDonnot:
## Summary

Continues the "help the compiler generate better code" work from `n1_full_compute` (the Eigen-copy-elimination pass, the report-only clang-tidy job) with a final/override/const-correctness sweep, plus a couple of related, smaller wins found along the way.

- **`final` on leaf classes / `override` everywhere it was missing** in `element_container/` and `powerflow_algorithm/`+`batch_algorithm/`. Verified signature-by-signature (not just grepped) — this caught real gaps a first pass missed (declarations overriding a base virtual without even the `virtual` keyword) and one false lead (a method with no base declaration at all, where the fix was to drop `virtual`, not add `override`).
- **Dropped 4 vestigial `virtual` keywords** on methods that declare nothing a base class declares and have no overrider, including `NRSystem<Base,Rest...>::mismatch/apply_step/mismatch_sq_norm_at` — called only through a concrete `NRSystem` member (never a base pointer) in the hottest part of the Newton-Raphson loop, so the vtable indirection they were paying for was pure waste.
- **Const-correctness**: const-qualified read-only getters, switched several `Eigen::Ref<...>` setter parameters to `Eigen::Ref<const ...>`. Also caught a case that looked const-safe but wasn't (`get_slack_weights_solver` mutates `bus_slack_weight_` as a side effect consumed later by `set_p_slack` — left non-const).
- **Dropped `virtual` wherever `override`/`final` already implies it** (codebase-wide), which let `modernize-use-override` actually be enabled in the report-only clang-tidy job with low noise (was ~80% noise against the pre-existing style before this cleanup). Also surfaced 5 more genuine override gaps invisible to a `virtual`-keyword-based manual search.
- **CI/build**: `.clang-tidy` gains `modernize-use-override` and `misc-const-correctness` (report-only). `src/core/CMakeLists.txt` gains `-Wsuggest-override -Wnon-virtual-dtor` as real (build-breaking) warnings on `lightsim2grid_core` — both zero-warning clean as of this PR, so a future regression fails the build instead of waiting to be read in the advisory report.
- **`[[nodiscard]]`** on `LSGrid`'s public getters (154 of them). Scoped to `LSGrid.hpp` for this pass since its getters are non-virtual; the virtual-heavy hierarchies would need per-override-chain review, left as a follow-up. `noexcept` was evaluated and intentionally skipped for the same reason (a `noexcept` base virtual constrains every override).

## Test plan

- [x] Standalone C++ build (`cmake -S src/tests -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=ON && cmake --build build`) after every commit: zero warnings, zero errors.
- [x] Full Catch2 suite (`ctest`): 43/43 passing after every commit.
- [x] `modernize-use-override` / `misc-const-correctness` / `-Wsuggest-override` / `-Wnon-virtual-dtor` all spot-checked directly with clang-tidy / GCC to confirm noise levels and zero regressions before enabling.
- [ ] Python extension build / pytest — not run in this environment (pybind11 not installed, no `setup.py` at the standalone-test layer); the `Eigen::Ref<const ...>` signature changes are ABI-visible to the bindings but follow an idiom (`update_topo`) already proven to bind cleanly with pybind11 in this codebase.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -3,7 +3,28 @@
 # family targets exactly the class of bug found in the Eigen unnecessary-copy
 # audit (by-value Eigen/STL params, const-ref-to-value copies, ...);
 # readability-const-return-type catches `const T foo()` return types that
-# block move semantics on the caller side.
+# block move semantics on the caller side. misc-const-correctness covers the
+# same ground as the manual const-correctness sweep (see the element_container
+# / LSGrid commit it landed alongside): local variables that are never
+# reassigned but aren't declared const. Runs in ~5 minutes over src/core as of
+# this writing (own measurement, GCC 13 / clang-tidy 18) -- noticeably slower
+# than the other checks here, but still well within the report-only job's
+# budget.
+#
+# modernize-use-override: this codebase used to keep the `virtual` keyword on
+# overrides alongside `override`/`final`, which this check flags as
+# redundant on every single one of them -- ~80% pure noise against that
+# style in an initial run. Rather than suppress the check, the redundant
+# `virtual` was dropped codebase-wide (virtual is implied by `override`, and
+# by `final` on an actual override) as part of the same final/override sweep,
+# bringing this down to ~40 residual warnings, all in two explainable
+# buckets: (1) destructors not annotated `override`/`final` -- valid C++ to
+# add, but this codebase doesn't follow that particular style and it wasn't
+# worth changing here; (2) `override final` used together for defensive
+# compile-time checking (marks BOTH "this really overrides something" and
+# "no further overrides") -- the check prefers `final` alone since it
+# implies `override`, but the pair is intentional in a few places and left
+# as-is.
 #
 # Known false-positive class: performance-unnecessary-value-param flags every
 # `Eigen::Ref<...>` parameter passed by value (the idiom used throughout this
@@ -14,7 +35,9 @@
 Checks: >
   -*,
   performance-*,
-  readability-const-return-type
+  readability-const-return-type,
+  misc-const-correctness,
+  modernize-use-override
 WarningsAsErrors: ''
 # Only report on our own headers (src/core, src/bindings) -- clang-tidy's
 # default (empty) header filter would otherwise only ever look at the
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -17,14 +17,16 @@
 # style in an initial run. Rather than suppress the check, the redundant
 # `virtual` was dropped codebase-wide (virtual is implied by `override`, and
 # by `final` on an actual override) as part of the same final/override sweep,
-# bringing this down to ~40 residual warnings, all in two explainable
-# buckets: (1) destructors not annotated `override`/`final` -- valid C++ to
-# add, but this codebase doesn't follow that particular style and it wasn't
-# worth changing here; (2) `override final` used together for defensive
+# and every destructor overriding a base's virtual destructor was annotated
+# `override` too, bringing this down to 2 residual warnings, both the same
+# explainable case: `override final` used together for defensive
 # compile-time checking (marks BOTH "this really overrides something" and
-# "no further overrides") -- the check prefers `final` alone since it
-# implies `override`, but the pair is intentional in a few places and left
-# as-is.
+# "no further overrides") on a method in a class that ISN'T itself `final`
+# (OneSideContainer::disconnect_if_not_in_main_component,
+# TwoSidesContainer::nb_line_end) -- the check prefers `final` alone since it
+# implies `override`, but the pair is intentional there and left as-is.
+# (Classes that are themselves `final` don't repeat `final` on their
+# methods -- redundant, since the class already can't be subclassed at all.)
 #
 # Known false-positive class: performance-unnecessary-value-param flags every
 # `Eigen::Ref<...>` parameter passed by value (the idiom used throughout this
--- a/src/core/LSGrid.hpp
+++ b/src/core/LSGrid.hpp
@@ -990,7 +990,7 @@ class LS2G_API LSGrid final
          */
         void set_status_droop_hvdc(int dcline_id, int status) {hvdc_lines_.set_status_droop(dcline_id, status, algo_controler_); }
         [[nodiscard]] int get_status_droop_hvdc(int dcline_id) const {return hvdc_lines_.get_status_droop(dcline_id);}
-        [[nodiscard]] std::vector<int> get_status_droop_hvdc_vect() const {return hvdc_lines_.get_status_droop_vect();}
+        [[nodiscard]] Eigen::Ref<const IntVect> get_status_droop_hvdc_vect() const {return hvdc_lines_.get_status_droop_vect();}
         /**
          * Per-solve data of the connected angle-droop (AC emulation) hvdc
          * lines, in solver bus labelling and per-unit. Consumed by the Hvdc
--- a/src/core/LSGrid.hpp
+++ b/src/core/LSGrid.hpp
@@ -197,7 +197,7 @@ class LS2G_API LSGrid final
             generators_.turnedoff_pv(algo_controler_);
         }
         [[nodiscard]] bool get_turnedoff_gen_pv() const {return generators_.get_turnedoff_gen_pv();}
-        void update_slack_weights(Eigen::Ref<const Eigen::Array<bool, Eigen::Dynamic, Eigen::RowMajor> > could_be_slack){
+        void update_slack_weights(const Eigen::Ref<const Eigen::Array<bool, Eigen::Dynamic, Eigen::RowMajor> > & could_be_slack){
             generators_.update_slack_weights(could_be_slack, algo_controler_);
         }
         void update_slack_weights_by_id(Eigen::Ref<const IntVect> slack_ids){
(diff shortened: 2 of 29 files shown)


src/core/LSGrid.hpp
```diff
@@ -989,8 +989,8 @@ class LS2G_API LSGrid final
          * pmax_2to1) is meant to be run between two solves, in Python.
          */
         void set_status_droop_hvdc(int dcline_id, int status) {hvdc_lines_.set_status_droop(dcline_id, status, algo_controler_); }
-        int get_status_droop_hvdc(int dcline_id) const {return hvdc_lines_.get_status_droop(dcline_id);}
-        std::vector<int> get_status_droop_hvdc_vect() const {return hvdc_lines_.get_status_droop_vect();}
+        [[nodiscard]] int get_status_droop_hvdc(int dcline_id) const {return hvdc_lines_.get_status_droop(dcline_id);}
+        [[nodiscard]] std::vector<int> get_status_droop_hvdc_vect() const {return hvdc_lines_.get_status_droop_vect();}
```
BDonnot: There should be a reference here std:: vector<int> & I think
And I suppose same for get_status_droop_vect from HVDC container

src/core/LSGrid.hpp
```diff
@@ -1071,69 +1071,69 @@ class LS2G_API LSGrid final
...
+        [[nodiscard]] const std::vector<bool>& get_trafo_status_side2() const { return trafos_.get_status_side_2();}
+        [[nodiscard]] tuple3d get_storages_res() const {return storages_.get_res();}
+        [[nodiscard]] const std::vector<bool>& get_storages_status() const { return storages_.get_status();}
+        [[nodiscard]] tuple3d get_sgens_res() const {return sgens_.get_res();}
+        [[nodiscard]] const std::vector<bool>& get_sgens_status() const { return sgens_.get_status();}
+        [[nodiscard]] tuple3d get_dcline_res1() const {return hvdc_lines_.get_res_side_1();}
+        [[nodiscard]] tuple3d get_dcline_res2() const {return hvdc_lines_.get_res_side_2();}
+        [[nodiscard]] const std::vector<bool>& get_dclines_status() const { return hvdc_lines_.get_status_global();}
+
+        [[nodiscard]] Eigen::Ref<const RealVect> get_gen_theta() const  {return generators_.get_theta();}
```
BDonnot, line 1105: Elsewhere in the code the idiom Eigen:Ref<const T> & is used rather more simple Eigen::Ref<const T> alone. If this makes sense, for consistency across the codebase I think it's preferable to use one.
And the cpp ref to the Eigen Ref seems more natural (and recommended, according to you, so please check, from Eigen docs)

src/core/LSGrid.hpp
```diff
@@ -1536,53 +1536,53 @@ class LS2G_API LSGrid final
...
         // VoltageControl (remote gen + SVC) converged reactive injection per
         // controller (pu), + its (kind: 0=GEN,1=SVC; element id) identity, in
         // controller registration order (empty when the extension is inactive).
         // Ground truth for external solvers deriving their own controller_q.
-        RealVect get_controller_q_solver()       const { return _algo.get_controller_q(); }
-        IntVect  get_controller_kind_solver()    const { return _algo.get_controller_kind(); }
-        IntVect  get_controller_elem_id_solver() const { return _algo.get_controller_elem_id(); }
+        [[nodiscard]] RealVect get_controller_q_solver()       const { return _algo.get_controller_q(); }
+        [[nodiscard]] IntVect  get_controller_kind_solver()    const { return _algo.get_controller_kind(); }
+        [[nodiscard]] IntVect  get_controller_elem_id_solver() const { return _algo.get_controller_elem_id(); }
```
BDonnot, line 1559: Eigen::Ref<const T> & here too maybe ?

src/core/element_container/HvdcLineContainer.hpp
```diff
@@ -345,17 +345,17 @@ class LS2G_API HvdcLineContainer final : public TwoSidesContainer<ConverterStati
...
+        void fillpv(std::vector<int>& bus_pv,
                             std::vector<bool> & has_bus_been_added,
                             const SolverBusIdVect & slack_bus_id_solver,
                             const SolverBusIdVect & id_grid_to_solver) const override {
             side_1_.fillpv(bus_pv, has_bus_been_added, slack_bus_id_solver, id_grid_to_solver);
             side_2_.fillpv(bus_pv, has_bus_been_added, slack_bus_id_solver, id_grid_to_solver);
         }
 
-        virtual void fillBp_Bpp(std::vector<Eigen::Triplet<real_type> > & /*Bp*/,
+        void fillBp_Bpp(std::vector<Eigen::Triplet<real_type> > & /*Bp*/,
```
BDonnot, line 358: I think this function is part of base class twosidedcontainer ? Or maybe HVDC is no longer a two sides container. I don't remember

src/core/powerflow_algorithm/BaseDCAlgo.hpp
```diff
@@ -81,22 +81,21 @@ class BaseDCAlgo final: public BaseAlgo
...
-
-        virtual RealMat get_ptdf() final;
-        virtual RealMat get_lodf(const IntVect & from_bus,
-                                 const IntVect & to_bus) final;
-        virtual Eigen::SparseMatrix<real_type> get_bsdf() final;  // TODO BSDF
-        
-        virtual void update_internal_Ybus(const Coeff & coeff, bool add) final{
+                           ) override final;
+
+        RealMat get_ptdf() override final;
```
BDonnot: A question here, the class is final. Is it necessary to write final to all the methods ?

.clang-tidy
```diff
@@ -3,7 +3,28 @@
...
+# budget.
+#
+# modernize-use-override: this codebase used to keep the `virtual` keyword on
+# overrides alongside `override`/`final`, which this check flags as
+# redundant on every single one of them -- ~80% pure noise against that
+# style in an initial run. Rather than suppress the check, the redundant
+# `virtual` was dropped codebase-wide (virtual is implied by `override`, and
+# by `final` on an actual override) as part of the same final/override sweep,
+# bringing this down to ~40 residual warnings, all in two explainable
+# buckets: (1) destructors not annotated `override`/`final` -- valid C++ to
```
BDonnot: You can these keywords if needed. If it removes some false positive here that is already a win

src/core/powerflow_algorithm/NRSystem.hpp
```diff
@@ -877,9 +877,9 @@ class NRSystem<Base, Rest...>
 
     // ----- NR iteration primitives -----------------------------------------------
 
-    virtual RealVect   mismatch()                           const;
-    virtual void       apply_step(const RealVect& dx);
-    virtual real_type  mismatch_sq_norm_at(const RealVect& dx) const;
+    RealVect   mismatch()                           const;
```
BDonnot, line 880: It was out of scope for this, but maybe this could be optimized somehow, returning a ref (modifiable) to an element initialized once per NR for example

BDonnot (commented):
Some modifications

src/core/LSGrid.hpp
BDonnot: Checked this — it can't safely be a reference, unfortunately. `status_droop_` is stored as `IntVect` (`Eigen::VectorXi`), not `std::vector<int>`:

```cpp
std::vector<int> get_status_droop_vect() const {
    return std::vector<int>(status_droop_.begin(), status_droop_.end());
}
```

Since the element types differ, this has to build a brand-new `std::vector<int>` each call — there's no `std::vector<int>` member anywhere to bind a `const std::vector<int>&` to. Returning a reference here would bind to a temporary that's destroyed at the end of the `return` statement, i.e. a dangling reference bug. Avoiding the copy would need either storing `status_droop_` as `std::vector<int>` instead of `IntVect` (bigger change, and it's used elsewhere as an Eigen vector) or caching a converted copy as a member (adds staleness-tracking complexity for what looks like an introspection-only getter, not a hot-path one). Left as-is; happy to revisit if this turns out to be hot after all.

---

src/core/LSGrid.hpp
BDonnot: Good catch on the inconsistency, though this exact line (`get_gen_theta`) is a getter *returning* `Eigen::Ref<const RealVect>` by value, not a parameter — for a return type that has to stay by-value (Eigen's own docs only recommend `const Ref<const T>&` for *parameters*; returning a reference here would bind to the temporary `Ref` the callee just constructed and dangle immediately after the return). `get_bus_vn_kv` a bit above it does the same thing and is correct as-is.

I think what you're pointing at is real, just landed a few lines further down than this comment: `update_gens_p`/`update_sgens_p`/`update_gens_v`/`update_loads_p`/`update_loads_q`/`update_storages_p` (and `update_continuous_values`, `update_slack_weights`) took their `Eigen::Ref<const T>` *parameters* by value instead of `const Eigen::Ref<const T> &`, inconsistent with e.g. `GeneratorContainer::init`. Fixed in 3f35ae4 for all of those. `update_topo` right next to them still uses by-value — it's pre-existing, not part of this PR, so left it alone for now rather than fold in an unrelated change, but happy to align it too if you'd like.

---

src/core/LSGrid.hpp
BDonnot: Same as my reply on the `get_gen_theta` thread above — this exact line (`get_controller_elem_id_solver`) is a by-value return, which needs to stay by-value. The actual parameter-style fix (bare `Eigen::Ref<const T>` → `const Eigen::Ref<const T> &`) landed in 3f35ae4 on `update_gens_p` and its siblings.

---

src/core/element_container/HvdcLineContainer.hpp
BDonnot: Checked — HVDC is still a `TwoSidesContainer` (`HvdcLineContainer : public TwoSidesContainer<ConverterStationContainer>`), but `TwoSidesContainer` itself never declares its own `fillBp_Bpp` — only `GenericContainer` does (the common root all containers derive from). So `HvdcLineContainer::fillBp_Bpp` overrides `GenericContainer::fillBp_Bpp` directly, skipping over `TwoSidesContainer` in between. `override` here is correct as-is.

---

src/core/powerflow_algorithm/BaseDCAlgo.hpp
BDonnot: Good point, no — dropped in 3f35ae4. `final` on individual methods only adds anything when the *class* can still be subclassed (so you want to lock down just that one method); when the class itself is already `final` like here, nothing can inherit from it at all, so the per-method `final` was pure noise. Kept `override` (still useful — self-checks the signature actually matches a base virtual) but removed the redundant `final` throughout `BaseDCAlgo` and did the same for the `ScalingPolicy` derived classes, which had the identical pattern.

---

.clang-tidy
BDonnot: Done in 3f35ae4 — every destructor that overrides a base's virtual destructor now says `override` explicitly (two of them, `ContingencyAnalysis`/`TimeSeries`, had neither `virtual` nor `override` before). That was the whole first bucket mentioned in this comment; `modernize-use-override` is down from ~40 residual warnings to 2 now (updated the comment above to match — the remaining 2 are the other bucket, `override final` kept together intentionally on two non-final classes).

---

src/core/powerflow_algorithm/NRSystem.hpp
BDonnot: Agreed that's a real opportunity, but seconding that it's out of scope here — this PR is deliberately just the final/override/const-correctness hygiene pass, not a hot-path allocation pass. Returning a mutable reference to a member `RealVect` (sized/allocated once at `init_topology` and reused every `compute_pf` iteration instead of a fresh vector per `mismatch()`/`apply_step()` call) sounds like a solid follow-up — happy to pick it up separately if useful.

---

src/core/LSGrid.hpp
BDonnot: Update: switched to `Eigen::Ref<const IntVect>` in d5bd5dd, per your follow-up. `status_droop_` is a real member (not a temporary), so returning a `Ref` view over it directly is safe and avoids the copy — matches `get_bus_vn_kv` and friends. Worth flagging: since `get_status_droop_hvdc_vect` is bound straight to Python, this flips its return type from `list[int]` to a numpy array (int32); confirmed that's wanted.

---

BDonnot (commented):
ok

PR #151, opened 2026-07-17
Fix core/bindings -march=native ABI mismatch, add runtime ABI-drift guard

BDonnot:
## Summary

- Fixes a real, previously-hit bug: `lightsim2grid_core` and `lightsim2grid_cpp` (the pybind11 bindings) are separate shared libraries, and `-march=native`/`-O3` were only ever applied to `lightsim2grid_core`. Since `-march=native` changes which SIMD ISA Eigen sees enabled, it changes `EIGEN_MAX_ALIGN_BYTES` and thus how Eigen aligns/allocates/frees its dynamic-size matrices — an object allocated under one alignment assumption and freed under another silently corrupted the heap (`free(): invalid size` / `double free or corruption`), reliably crashing on the very first call whose return value crossed the module boundary. Fixed by mirroring the flags onto `lightsim2grid_cpp` in `src/bindings/python/CMakeLists.txt`, the same way `__SANITIZE`/`__DEBUG_ASSERTS` already were.
- Adds `examples/cmake/MatchLightsim2gridBuildFlags.cmake`, which the above fix's example-plugin `CMakeLists.txt`s reference via `include()` but never committed — both examples failed to configure without it.
- Adds a runtime "ABI tag" guard (`src/core/Ls2gAbiTag.hpp`) as defense-in-depth on top of the CMake-level flag matching: `EIGEN_MAX_ALIGN_BYTES`, the resolved `EIGEN_VECTORIZE_*` flags, and the Eigen version are compared between independently-compiled translation units at two points — `lightsim2grid_cpp`'s module init vs. `lightsim2grid_core` (the exact boundary that broke), and `AlgorithmRegistry::register_solver` vs. any third-party solver plugin loaded via `load_algorithm_plugin()`. A mismatch now raises a clear error at load time instead of silently corrupting the heap later. A `static_assert` can't do this: no single translation unit has visibility into another, separately-compiled one's compiler flags — only an actual cross-library function call comparing two independently-evaluated tags can.

## Test plan

- [x] Built `lightsim2grid_core` standalone once plain and once with `__COMPILE_MARCHNATIVE=1`, then linked a separately-compiled (no `-march=native`) test program against the AVX-enabled core: the ABI-tag mismatch was detected exactly (`align=16 bytes` vs `align=64 bytes`), and a matching-flags build produced no false positive.
- [x] Forced a mismatched tag into `AlgorithmRegistry::register_solver`: confirmed it throws before the plugin's factory ever runs, and a correctly-tagged registration still succeeds.
- [x] Full `pip install .` (real wheel build) succeeds; module imports cleanly; `available_solver_names()` works.
- [x] Full Catch2 suite passes (1200 assertions, 49 cases), including the new `src/tests/test_ls2g_abi_tag.cpp`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Prjn9yN3C5r1doKzEMu2av)_


src/core/Ls2gAbiTag.hpp
```diff
@@ -0,0 +1,127 @@
...
+// visibility into what flags another, separately-compiled TU used.
+struct Ls2gAbiTag {
+    // Bump only if this struct's meaning changes (e.g. a field is added or
+    // reinterpreted), not for ordinary Eigen/compiler upgrades.
+    unsigned tag_version = 1;
+
+    unsigned eigen_max_align_bytes = EIGEN_MAX_ALIGN_BYTES;
+    unsigned eigen_world_version = EIGEN_WORLD_VERSION;
+    unsigned eigen_major_version = EIGEN_MAJOR_VERSION;
+    unsigned eigen_minor_version = EIGEN_MINOR_VERSION;
```
BDonnot, line 45: https://libeigen.gitlab.io/news/eigen_5.0.0_released/ 

Eigen will now uses a semver (or semver-like) structure :

> From now on, version numbers will follow the MAJOR.MINOR.PATCH format, indicating breaking changes, new features, and bug fixes,

Wonder if it's worth having also the patch version here too.

src/core/Ls2gAbiTag.hpp
```diff
@@ -0,0 +1,127 @@
...
+#else
+    bool vectorize_neon = false;
+#endif
+
+    bool operator==(const Ls2gAbiTag& o) const {
+        return tag_version == o.tag_version
+            && eigen_max_align_bytes == o.eigen_max_align_bytes
+            && eigen_world_version == o.eigen_world_version
+            && eigen_major_version == o.eigen_major_version
+            && eigen_minor_version == o.eigen_minor_version
```
BDonnot, line 103: I wonder (now that eigen uses semver) if it's really mandatory to have same minor and patch version.
They are tagged as new feature and bug fix.
Or maybe defensively, we put a safeguard and say everything should match

BDonnot (commented):
2 small questions rather than modifications

src/core/Ls2gAbiTag.hpp
BDonnot: Good catch — added `eigen_patch_version` (guarded with `#ifdef`, since pre-5.0 Eigen never defined `EIGEN_PATCH_VERSION`). Also fixed `describe()`, which was printing `world.major.minor` (misleadingly looked like "3.5.0") instead of Eigen's own `major.minor.patch` convention — now shows "5.0.1" matching `EIGEN_VERSION_STRING`. See e5d4b01.

---

src/core/Ls2gAbiTag.hpp
BDonnot: Went with the defensive option: require everything to match exactly (added patch too). Reasoning: Eigen is header-only, so its semver promise covers API compatibility for code *using* Eigen, not "the internal aligned-malloc offset arithmetic stays byte-identical across a patch bump" — that's not something Eigen tests or promises, since there's no shared-object ABI to keep stable in the first place. In normal use, core/bindings/any in-repo plugin all build against the one Eigen commit vendored in `eigen/`, so this only ever fires for a plugin deliberately pointed at a different Eigen copy — being strict there is the safer default, and the error message says exactly what mismatched. Documented this reasoning + the recommendation to build against the provided Eigen headers in docs/solver_plugin.rst. See e5d4b01.

---

BDonnot (commented):
Ok

BDonnot (commented):
Ok

PR #152, opened 2026-07-18
Add Levenberg-Marquardt Newton-Raphson as an out-of-tree example plugin

BDonnot:
## Summary

Some real grids have ill-conditioned clusters (high resistance, heavily loaded, little local generation) where plain Newton-Raphson diverges, and the existing step-scaling policies (`Iwamoto`, `MaxVoltageChange`) only rescale the same bad Newton direction rather than fixing it — they can't help when the underlying Jacobian sub-block is itself near-singular. Levenberg-Marquardt damping addresses this directly by solving `(J + λD)dx = -F` instead of `Jdx = -F`, with λ adapted via an accept/reject loop on each trial step's actual (nonlinear) mismatch reduction.

- **Core** (`src/core/powerflow_algorithm/NRSystem.hpp`/`.tpp`): one small, generic addition — `NRSystem::update_trailing_feature_values(count, deltas)`. It directly updates the last `count` already-declared Jacobian feature entries in `J_.valuePtr()`, bypassing `fill_internal_variables()`/`fill_J()` entirely. This is intentionally not diagonal- or LM-specific: any extension that reserves trailing feature slots and needs a fast partial refresh can use it.
- **`examples/lm_algorithm/`** (new, out-of-tree plugin — not wired into the main build/CI, following the same pattern as `examples/dist_slack_algorithm/` and `examples/external_algorithm/`):
  - `DiagonalDamping.hpp` — `DiagonalEntry`, a minimal `NRSystem` extension that reserves one feature slot per existing unknown (for the diagonal load) and claims no rows/cols of its own, plus `with_diagonal_entry<NRSystem>` to inject it as the last extension of an existing `SingleSlackNRSystem`/`MultiSlackNRSystem`.
  - `LMNRAlgo.hpp`/`.tpp` — the algorithm itself: a `BaseAlgo` sibling to `NRAlgo` (not built on top of it, since the inner accept/reject loop — potentially several factorize+solve+mismatch-eval cycles per outer NR iteration — doesn't fit `NRAlgo`'s single solve-per-iteration control flow, its `RefactorPolicy` (assumes stable-across-iterations J values), or its `ScalingPolicy` (only rescales an already-solved step post-hoc, can't touch the matrix being solved)). Trial evaluation uses the existing `NRSystem::mismatch_sq_norm_at(dx)` to check the true nonlinear residual without mutating internal voltage state, so a rejected trial never needs a rollback.
  - `LMNRAlgo_plugin.cpp` + `CMakeLists.txt` — registers `"NR_LM_SparseLU"` and `"NR_LM_KLU"` via the standard `AlgorithmRegistry` self-registration mechanism.
  - `test_plugin.py` — registration/load smoke test, same scope as the existing `external_algorithm` one.

## Test plan

- [x] `examples/lm_algorithm` builds cleanly (including with `-Wall -Wextra`), with only the expected core symbols left undefined for the dynamic linker (`Base`/`Hvdc`/`VoltageControl::update_state`, `SparseLULinearSolver::*`, `AlgorithmRegistry::*`) — confirming the new code only depends on symbols already compiled into `lightsim2grid_core`.
- [x] The existing `examples/dist_slack_algorithm` plugin still builds unchanged against the modified `NRSystem`, i.e. no regression from the core addition.
- [ ] End-to-end power flow run (`test_plugin.py` + a real grid via `LightSimBackend`) — not run in this environment (no compiled `lightsim2grid_cpp` extension available here).
- [ ] Validation on a real ill-conditioned case where plain NR/Iwamoto diverge and `MaxVoltageChange` lands on an unrealistic branch, compared against a PowSyBl/OpenLoadFlow reference — needs to be run against local grid data outside this repo.

---
_Generated by [Claude Code](https://claude.ai/code/session_018PXF6if9WFcunG1GrsFLm5)_

### DIFF
--- a/examples/lm_algorithm/CMakeLists.txt
+++ b/examples/lm_algorithm/CMakeLists.txt
@@ -1,0 +1,99 @@
+cmake_minimum_required(VERSION 3.15)
+project(lm_algorithm CXX)
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+# -----------------------------------------------------------------------
+# Strategy 1: find via installed CMake config
+#   pip install lightsim2grid  →  lightsim2grid.get_cmake_dir()
+#   cmake --install             →  _install/lib/cmake/lightsim2grid_core/
+# Pass -DLIGHTSIM2GRID_CMAKE_DIR=<path> to hint the location, or rely on
+# CMAKE_PREFIX_PATH.
+# -----------------------------------------------------------------------
+find_package(lightsim2grid_core CONFIG QUIET
+    HINTS "${LIGHTSIM2GRID_CMAKE_DIR}")
+
+# -----------------------------------------------------------------------
+# Strategy 2: fall back to source tree (in-repo development builds).
+# -----------------------------------------------------------------------
+if(NOT lightsim2grid_core_FOUND)
+    if(NOT DEFINED LIGHTSIM2GRID_SRC)
+        set(LIGHTSIM2GRID_SRC "${CMAKE_CURRENT_SOURCE_DIR}/../../src/core")
+    endif()
+    if(NOT DEFINED Eigen3_INCLUDE)
+        set(Eigen3_INCLUDE "${CMAKE_CURRENT_SOURCE_DIR}/../../eigen")
+    endif()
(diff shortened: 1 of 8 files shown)


BDonnot (commented):
Ok

PR #153, opened 2026-07-21
core: const-correctness and copy-elimination low-hanging fruit

BDonnot:
Follow-up to the const / `const Eigen::Ref` sweep (#151), picking up member functions and return types that pass missed under `src/core`. Everything here is either a read-only marker or a copy-enabling change — **no behavioural change**.

### What changed

**`const` on read-only member functions** (callable through a `const` object / reference, and documents intent):
- `LSGrid`: `get_bus1_powerline`, `get_bus2_powerline`, `get_bus1_trafo`, `get_bus2_trafo`, `get_Ybus_solver`, `get_dcYbus_solver`
- `PandaPowerConverter`: `get_line_param`, `get_line_param_legacy`, `get_trafo_param_pp2`, `get_trafo_param_pp3`, `_check_init` — the whole class is read-only apart from its two setters
- `BaseAlgo::is_linear_solver_valid` (reads `err_` only)

**Drop top-level `const` on by-value return types** (`readability-const-return-type`): returning `const T` **by value** blocks move-construction / move-assignment at the call site (e.g. the pybind copy-to-Python path) and buys nothing. Applied to the freshly-computed getters in `LSGrid`: `get_Ybus`, `get_dcYbus`, `get_Sbus`, `get_dcSbus`, `get_pv[_numpy]`, `get_pq[_numpy]`, `get_slack_ids[_numpy]`, `get_slack_ids_dc[_numpy]`, `get_slack_weights`, `get_V`, `get_Va`, `get_Vm`.

**Misc**: `std::endl` → `'\n'` in `GeneratorContainer::cout_v` (drops the per-line stream flush; `performance-avoid-endl`).

### How it was found
The report-only clang-tidy config already in the tree (`performance-*`, `readability-const-return-type`, `misc-const-correctness`) run over every `src/core` translation unit, plus a manual member-function pass. Deliberately **not** touched: the `get_ptdf` / `get_lodf` chain (genuinely mutates `timer_ptdf_` and the linear-solver factorization — correctly non-const), the `performance-enum-size` findings (several enums cross the pybind / `BinaryArchive` boundary, so narrowing the width risks ABI/format breakage), and the `misc-const-correctness` const-local findings (cosmetic, no codegen effect).

### Verification
`src/core` builds and the C++ unit test suite passes (1201 assertions in 49 test cases).

---
_Generated by [Claude Code](https://claude.ai/code/session_01LY1xhzf8cUjqFoNNj7y2Gc)_

### DIFF
--- a/src/core/DataConverter.cpp
+++ b/src/core/DataConverter.cpp
@@ -13,7 +13,7 @@
 namespace ls2g {
 
 
-void PandaPowerConverter::_check_init(){
+void PandaPowerConverter::_check_init() const {
     if(sn_mva_ <= 0.){
         throw std::runtime_error("PandaPowerConverter::_check_init: sn_mva has not been initialized");
     }
@@ -35,7 +35,7 @@ std::tuple<RealVect,
                                                     const Eigen::Ref<const RealVect> & trafo_vkr_percent,
                                                     const Eigen::Ref<const RealVect> & trafo_sn_trafo_mva,
                                                     const Eigen::Ref<const RealVect> & trafo_pfe_kw,
-                                                    const Eigen::Ref<const RealVect> & trafo_i0_pct)
+                                                    const Eigen::Ref<const RealVect> & trafo_i0_pct) const
 {
     //TODO consistency: move this class outside of here
     _check_init();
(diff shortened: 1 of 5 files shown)


BDonnot (commented):
Ok

PR #154, opened 2026-07-23
Make solver-plugin registration exception-safe at load time

BDonnot:
## Motivation

Solver plugins previously self-registered from a static `AlgorithmRegistrar` constructor that fired during `dlopen()`. That constructor can throw — on an ABI-tag mismatch, or on a duplicate solver name — and an exception escaping a static constructor while the dynamic loader runs its init routines cannot be caught: it unwinds into C loader frames with no handler and calls `std::terminate()`, aborting the whole Python interpreter (and `ctypes` couldn't translate it to a Python exception either). So the very ABI check meant to turn silent heap corruption into an actionable error instead became a hard, uncatchable crash. The docs even promised an `OSError` on ABI mismatch that never actually surfaced.

## What this changes

Registration no longer happens inside a static constructor during `dlopen`. Instead:

- **New `LS2G_PLUGIN_ENTRY(fn)` macro** generates an exported `ls2g_register_plugin` entry point from a plain `void(ls2g::PluginRegistrar&)` function the plugin author writes. `LS2G_PLUGIN_EXPORT` (new, in `ls2g_api.hpp`) handles `dllexport`/visibility for a symbol the plugin itself exports. The old static `AlgorithmRegistrar` is gone.
- **`load_algorithm_plugin()` is now a pybind11 C++ function** (`_load_algorithm_plugin`) that `dlopen`/`dlsym`s the plugin and calls the entry point inside `try/catch`. Every failure — ABI mismatch, duplicate name, missing entry point, missing file — surfaces as a catchable Python `RuntimeError` instead of terminating the process. A rejected plugin is `dlclose`'d; a successful one stays mapped.
- **Atomic registration:** `AlgorithmRegistry::register_all()` ABI-checks once, verifies no name collides, then commits via copy-into-merged-map + `noexcept` swap (strong exception guarantee). A plugin exposing several solvers registers all of them or none.
- **Buffer safety:** the `noexcept` entry helper copies its diagnostic into the caller's buffer with an explicit bounded write (respects `errbuf_len`, always NUL-terminates, null/zero-length safe). Covered by a valgrind test.

## Behaviour changes (noted in CHANGELOG)

- `load_algorithm_plugin()` now raises `RuntimeError` (not `OSError`) on failure.
- Loading a plugin whose solver name is already registered is rejected — **including loading the same plugin twice** (previously a silent no-op, since static init ran once per file).
- Plugin authors: register via `LS2G_PLUGIN_ENTRY` + a `PluginRegistrar` function instead of a static `AlgorithmRegistrar`.

## Tests / CI

- **`src/tests/test_plugin_registration.cpp`** (Catch2, runs under valgrind in the existing `cpp_unit_tests` workflow): atomic rollback, ABI rejection, entry-point return-codes-not-exceptions, and short/null error-buffer handling.
- **`lightsim2grid/tests/test_plugin_registration.py`**: loads a real compiled plugin and asserts every error path raises **without crashing the interpreter**.
- **`lightsim2grid/tests/test_solver_registry.py`**: fixed `TestPluginLoading`, which was silently skipping (it pointed at a non-existent `examples/external_solver/` and imported a non-existent `load_solver_plugin`). It now genuinely loads the plugin and passes.
- Both python modules are now run in `main.yml`'s existing `test_plugin_against_installed` job (ubuntu/windows/macos), which already builds the example plugin after installing lightsim2grid — so `TestPluginLoading` actually runs and the new Windows `LoadLibrary` path gets cross-OS coverage. The temporary standalone workflow was folded into that job.

## Docs & examples

- `docs/solver_plugin.rst` reworked to the entry-point mechanism (lookup-flow diagram, C++ API, worked example, `:raises RuntimeError:`), plus a few stale-reference fixes.
- The three example plugins (`external_algorithm`, `lm_algorithm`, `dist_slack_algorithm`) converted to `LS2G_PLUGIN_ENTRY`; the example smoke scripts updated to `available_algorithm_names()`.

## Local verification

- C++ suite green (58 cases, valgrind-clean on the `[plugin]` tests).
- Extension builds; all three example plugins build against the installed package; the Python plugin-registration and solver-registry tests pass, including the previously-skipped `TestPluginLoading` (Linux locally — Windows/macOS legs run in this PR's CI).

---
_Generated by [Claude Code](https://claude.ai/code/session_01J97bpvJk9BS1qXnGPGYnGp)_


src/bindings/python/binding_module.cpp
```diff
@@ -48,6 +56,66 @@
...
+// file) surfaces as a normal Python exception. This is the whole safety
+// improvement over dlopen'ing the plugin directly from ctypes: there, an
+// exception thrown by a static-constructor registration would unwind into the
+// C loader frames and std::terminate() the interpreter. A rejected plugin is
+// unloaded again so it leaves no trace; a successfully-registered one is
+// intentionally kept mapped for the life of the process, since the registry now
+// holds factories whose code lives in it.
+static void load_algorithm_plugin_impl(const std::string& path)
+{
+    char errbuf[512];
```
BDonnot: I want to be certain that whatever happens, nothing will overflow the buffer. I don't see lots of check regarding this in this function.

And also the 512 here should be declared a a (static) const size_t max_buffer_size = 512 or something.

src/bindings/python/binding_module.cpp
```diff
@@ -48,6 +56,66 @@
...
+                                 "' (LoadLibrary failed, error code " + std::to_string(::GetLastError()) + ").");
+    }
+    auto entry = reinterpret_cast<ls2g::ls2g_plugin_entry_fn>(
+        ::GetProcAddress(handle, "ls2g_register_plugin"));
+    if (entry == nullptr) {
+        ::FreeLibrary(handle);
+        throw std::runtime_error("lightsim2grid.load_algorithm_plugin: '" + path +
+                                 "' is not a lightsim2grid plugin (no 'ls2g_register_plugin' entry point).");
+    }
+    int rc = entry(&ls2g::AlgorithmRegistry::instance(), errbuf, sizeof(errbuf));
```
BDonnot: There is no check that the user defined "entry" function cannot overflow the errbuf size. 
Maybe a check with a plugin with a name >= 512 char would be good. Just to be sure.

BDonnot (commented):
2 small things to check

src/bindings/python/binding_module.cpp
BDonnot: Fixed in 044fcea.

- The `512` is now `static const std::size_t max_errbuf_size = 512;`.
- To make the host over-read impossible *by design* (regardless of what the plugin's entry point does), I force a terminator into the last slot right after the entry call, before the buffer is ever read as a C string: `errbuf[max_errbuf_size - 1] = '\0';`. So even a plugin that ignored the size contract cannot make the `... + errbuf` read run past the buffer.
- I also refactored the function so the entry call + buffer handling happen exactly once (via an `unload` closure) instead of being duplicated in the Windows/POSIX branches — that duplication was what made the buffer logic easy to get wrong.

The write side is bounded by construction for plugins built with `LS2G_PLUGIN_ENTRY` (see the reply on the other thread).

---

src/bindings/python/binding_module.cpp
BDonnot: The write side is bounded by construction for the supported path. For a plugin built with `LS2G_PLUGIN_ENTRY`, the entry point runs through `ls2g::detail::ls2g_run_plugin_entry`, whose `write_err` never writes past index `errbuf_len - 1` and always NUL-terminates:

```cpp
if (errbuf == nullptr || errbuf_len == 0) return;
std::size_t i = 0;
const std::size_t last = errbuf_len - 1;
if (msg != nullptr)
    for (; i < last && msg[i] != '\0'; ++i) errbuf[i] = msg[i];
errbuf[i] = '\0';   // i <= last
```

So the plugin author's error message (e.g. a huge solver name) is truncated, not overflowed — the author can't cause an overflow through the normal API. (A plugin that hand-rolls its own `ls2g_register_plugin` and ignores `errbuf_len` is arbitrary native code we can't constrain, but the host-side read is still protected by the forced terminator from the other thread.)

Added the exact check you suggested — a solver name of 1000 chars (≫ 512) pushed through the real entry helper with the same 512-byte buffer the host uses, asserting truncation + in-bounds termination, in `src/tests/test_plugin_registration.cpp` ("a solver name longer than the buffer cannot overflow it"). It runs under valgrind in the `cpp_unit_tests` workflow, so an out-of-bounds write would fail CI.

---

PR #155, opened 2026-07-24
Harden grid loading: whole-grid validation, solver-output checks, trust-boundary docs

BDonnot:
## Why

Release wheels are built `-O3 -DNDEBUG`, so Eigen's and the standard library's bounds checks are compiled out. The grid carries several fields that are used to index internal vectors on the powerflow / topology-update hot paths — element `bus_id`, `subid`, `pos_topo_vect`, generator slack and `regulated_bus_id`. Until now `set_state()` (used by **both** pickle and the fast binary format) validated only array *lengths*, not *values*. So a well-formed-but-poisoned state — e.g. an out-of-range bus id in a crafted binary file, the one genuinely-untrusted, non-pickle input channel — loaded cleanly and then caused an out-of-bounds read/write on the next powerflow.

This PR closes that gap at the source boundary, adds a cheap sanity check at the solver→grid seam, and documents the trust model.

## What

- **`LSGrid::check_grid()` (exposed as `GridModel.check_grid()`)** — a whole-grid validator: every index field is range-checked (`bus_id`, `subid`, `pos_topo_vect`, slack weights, `regulated_bus_id`), `pos_topo_vect` must be a permutation of `[0, dim_topo)`, and the electrical input arrays (`r`/`x`/half-line shunts, thermal limits, `p`/`q`/`vm` set-points) must be finite. Raises `IndexError` (out-of-range index) / `RuntimeError` (structural or non-finite), or returns `None`. It runs in O(number of elements), off the solver hot path.
- **Automatic validation on load** — `check_grid()` is called at the end of `set_state()`, so loading a pickle or a binary file that is well-formed but inconsistent now raises a clean exception instead of causing a later OOB. Every grid loader (`init_from_pandapower` / `init_from_pypowsybl` / `init_from_matpower` / `init_from_powermodels`) also calls it before returning.
- **Solver-output sanity check** at the algorithm→grid seam — a (possibly plugin / custom) solver that reports convergence but returns a wrong-sized `V`/`Va`/`Vm` now raises `RuntimeError` (instead of an OOB read), and one returning non-finite voltages is reported as non-converged (no `NaN`/`Inf` leaks into the results).
- **Tests** — `test_check_grid.cpp` (Catch2, run under `ctest` + valgrind) and `test_check_grid.py`, plus the existing `test_LSGrid_out_of_bounds.py`, are wired into the ASan/UBSan job. The binary corruption sweep (`TestCorruptionSweep`) now accepts `IndexError` as an equally-clean rejection alongside `RuntimeError`.
- **Docs** — a new `docs/security.rst` trust-boundary page (pickle and solver plugins are trusted-input-only by design; the binary format is hardened against untrusted input and validated with `check_grid` on load), plus CHANGELOG and binary-serialization doc updates.

## Design notes

- Validation is deliberately **always-on** (not gated behind a debug flag): it runs once at load, off the hot path, and is what turns a malformed input into a clean exception rather than an OOB access.
- The source boundary (`set_state` / `init_*`) is the highest-leverage place to validate: everything downstream — including the already-well-guarded `update_topo` path, which trusts `pos_topo_vect_` — is protected once nothing poisonous can get past the source.
- `check_grid` keeps `IndexError` for an out-of-range index (clearer than a generic `RuntimeError`); the binary corruption-sweep contract was widened to accept it rather than weakening the exception type.

## Verification

- Full C++ suite (**71 tests**) passes; **valgrind clean (0 errors, 0 leaks)** on the poisoned-input tests — the proof there is no OOB.
- Python extension builds; `check_grid()` returns `None` on a valid grid, maps `std::out_of_range → IndexError` / `std::runtime_error → RuntimeError`, and `ac_pf` still converges normally.
- A local single-byte-flip corruption sweep over a saved grid: 1335 `RuntimeError`, 331 clean loads, 32 `IndexError`, **0 disallowed outcomes**.

https://claude.ai/code/session_01YReCvKFeTQiLQeacZwP32G

---
_Generated by [Claude Code](https://claude.ai/code/session_01YReCvKFeTQiLQeacZwP32G)_


docs/security.rst
```diff
@@ -0,0 +1,107 @@
...
+       an untrusted source.
+   * - Loading a solver plugin (:func:`lightsim2grid.load_algorithm_plugin`)
+     - **Trusted code only.** A plugin is a native shared library that runs
+       in-process; once loaded it can do anything the host process can. Only
+       load plugins you built or otherwise trust.
+   * - Loading a grid from the fast binary format (``save_binary`` /
+       ``load_binary``)
+     - **Hardened against untrusted input.** See below.
+   * - Building a grid from a source file (pandapower, pypowsybl, matpower,
+       powermodels)
```
BDonnot: Also add (pandapower, pypowsybl, matpower, power models, including grid2op environment)

docs/security.rst
```diff
@@ -0,0 +1,107 @@
...
+     - **Validated on load.** See below.
+
+Pickle and plugins are trusted-input-only by design
+---------------------------------------------------
+
+Unpickling arbitrary data and loading a native plugin are both, fundamentally,
```
BDonnot: Unpickling arbitrary data, loading native plugin or using a grid2op environment are all fundamentally etc

docs/security.rst
```diff
@@ -0,0 +1,107 @@
...
+ways of running code chosen by whoever produced the input. No amount of
+validation inside lightsim2grid changes that, so the only safe rule is: **do not
+load a pickle or a plugin from a source you do not trust.** This is the same
```
BDonnot: Add "load a pickle or plugin or call grid2op.make"

docs/security.rst
```diff
@@ -0,0 +1,107 @@
...
+posture as ``numpy``, ``pandas`` (``read_pickle``), ``torch.load`` and any other
+library that supports pickling.
+
+The version string embedded in a pickled grid is a *compatibility* check (it
+refuses to load a grid saved by a different lightsim2grid version); it is **not**
+a security control.
+
+The binary format is the channel to use for untrusted data
```
BDonnot: Prefer a terminology like "least dangerous", I don't want people to think it's a good idea to load untrusted data (regardless of the protection made here)

docs/security.rst
```diff
@@ -0,0 +1,107 @@
...
+* **Semantic level.** After the bytes are read back into a grid, ``check_grid()``
+  runs automatically (see below): a file that is byte-wise well-formed but
+  *inconsistent* — for example one with an out-of-range bus id — raises a clean
+  exception instead of leaving the grid in a state that would cause an
+  out-of-bounds memory access on the next powerflow.
+
+``check_grid()``: whole-grid consistency validation
+---------------------------------------------------
+
+:py:meth:`LSGrid.check_grid` (also available as ``GridModel.check_grid``) verifies that a grid
```
BDonnot: Gridmodel is now called LSGrid. It is deprecated

docs/security.rst
```diff
@@ -0,0 +1,107 @@
...
+
+You normally do not need to call it yourself:
+
+* it runs automatically when a grid is loaded from a pickle or the binary format
+  (via ``set_state``);
+* every grid loader (``init_from_pandapower``, ``init_from_pypowsybl``,
+  ``init_from_matpower``, ``init_from_powermodels``) calls it before returning.
+
+It is exposed so you can validate a grid you built or modified by hand. It runs
+in time proportional to the number of elements in the grid, so it is cheap
```
BDonnot: So it is relatively cheap

lightsim2grid/tests/test_check_grid.py
```diff
@@ -0,0 +1,85 @@
...
+
+import pandapower.networks as pn
+from lightsim2grid.network import init_from_pandapower
+
+
+# an id far above any element count, but still inside the C `int` range
+BIG = 10 ** 6
+
+
+class TestCheckGrid(unittest.TestCase):
```
BDonnot, line 40: Can you do same type of tests for substation id (***_to_subid)

src/core/LSGrid.cpp
```diff
@@ -1164,6 +1228,37 @@ void LSGrid::process_results(bool conv,
     }
 }
 
+bool LSGrid::_check_solver_output(bool ac)
```
BDonnot, line 1291: Is this possible to run this check only for external plugins ?

BDonnot (commented):
Also the direct powerflow from python (ac_pf or run_pf or something function that is imported from pandapower)

src/core/LSGrid.cpp
BDonnot: Yes, and it turned out to be free: `AlgorithmType::Custom` already means exactly "not one of the built-in solvers" — `name_to_algo_type()` returns it for any name outside the built-in set, which is precisely the plugin case. So `process_results` now gates the check on it:

```cpp
const bool is_external_algo =
    (ac ? _algo.get_type() : _dc_algo.get_type()) == AlgorithmType::Custom;
if (is_external_algo) conv = _check_solver_output(ac);
```

Built-in solvers now skip it entirely and pay nothing per solve. Done in 0bafd6d.

The two seam tests still cover the behaviour (their solvers register under non-built-in names, so they are `Custom`); I added an explicit `REQUIRE(grid.get_algo_type() == AlgorithmType::Custom)` to each so the coupling is visible and cannot silently break if that classification ever changes.

---

BDonnot (commented):
Ok

PR #156, opened 2026-07-26
Fix memory-safety holes on the untrusted-input paths

BDonnot:
Audit of every entry point consuming data from outside the C++ core — the python bindings, pickle, the binary format, and the `newtonpf` / solver entry point — in the spirit of the `refactor_every_n` SIGFPE fix.

Every issue below was confirmed with a working proof of concept **before** the fix (`SIGSEGV`, or `malloc(): unsorted double linked list corrupted`) and re-run after. Release wheels are `-O3 -DNDEBUG`, so neither Eigen's nor the standard library's own bounds checks are there to catch any of them.

## The serious one: `SubstationContainer::set_state` did no validation at all

Its body was `// check sizes` / `// TODO dev switches` followed by nothing — and it restores the **root of the grid's index space**: `nb_bus()` is the bound `check_grid()` validates every element bus id against, `bus_status_` is the vector those same ids then index, and `n_sub_` is both the substation-id bound and a modulo divisor. Nothing re-derives them from one another, so `check_grid()` was validating ids against one array while the code indexed a different, unrelated one.

Patching a saved `.lsb` to shrink `bus_status` from 4000 entries to 64:

```
LSGrid.load_binary('bad.lsb')  ->  LOADED WITHOUT ERROR. total_bus = 4000
free(): invalid pointer                    # heap corruption
```

`disconnect_all_buses()` writes `nb_bus()` entries into a 64-entry vector. This contradicted what `docs/security.rst` promised about the binary format, so that page is corrected too.

The same file also gave `n_sub_ == 0` → `bus_id % n_sub_` in `sub_id_of_bus()` — the same SIGFPE class as `refactor_every_n == 0`, in a different place. `n_sub_ * nmax_busbar_per_sub_` could also overflow the `int` holding it.

## The one needing no attacker at all

`SubstationInfo` read `sub_names_[id]` bounded only against the *substation count*, but names are optional and the pandapower / matpower / powermodels loaders never set them:

```python
for sub in grid.get_substations():   # Segmentation fault
```

A live crash in ordinary API use, unrelated to untrusted input.

## The per-substation voltage invariant was never enforced

With `n` = `n_sub` and `k` = `nmax_busbar_per_sub`, the buses of substation `I` are `I, I+n, …, I+(k-1)n`, so the invariant is

```
for all 0 <= I < n, for all 0 <= j < k:  bus_vn_kv[I] == bus_vn_kv[I + j*n]
```

`init_bus` has a check for exactly this, but it went through 1-based `LocalBusId` and looped local ids `[1, k)` — so it opened by comparing bus `I` with **itself** and stopped one busbar short. At the usual `k == 2` it compared bus `I` to bus `I` and checked nothing whatsoever; a substation with busbars at 100 kV and 999 kV was accepted. It is now written as the plain index relation above, and lifted into `check_valid()` so it is re-checked on load — `set_state` bypasses `init_bus` entirely.

## The rest

| Issue | Reachable from |
|---|---|
| `set_ls_to_orig_internal` sizes the reverse map from `lpNorm<Infinity>()` (max **absolute** value) then indexes with the values — `-5` sizes from 5, writes at index `-5` | `_ls_to_orig` property, pickle, binary |
| `_init_kwargs` keys/values lengths stored independently; `set_state` walked keys indexing values — over-read of `vector<string>` | pickle, binary |
| `update_topo` never checked its arrays are `dim_topo` long, though it indexes them by topology position | python |
| `update_slack_weights` never checked `could_be_slack` has one entry per generator | python |
| out-of-range policy enum in an `AlgoConfig` silently hit a `default:` branch and round-tripped back out | pickle, binary |
| `init_sub()` wrote `n_sub` doubles into an unsized `sub_vn_kv_` — latent (uncalled, unbound) but wrong for whoever wires it up | — |

Also fixed `set_orig_to_ls`, which did not build the inverse it claimed (wrong iteration range *and* wrong stored value), and `check_grid()` now covers the substation container and the bus-id mapping vectors (`_ls_to_orig` / `_orig_to_ls` / `_bus_fusion_rep`), which it previously ignored entirely.

## `sub_vn_kv` is dead state — documented, not silently left

Neither of its writers (`init_sub`, the two-argument constructor) is called anywhere and nothing reads it back, so it is empty on every grid every loader produces and in every binary file saved so far — including this repo's own format-4 fixture — yet it is serialized into all of them. It is also redundant: `sub_vn_kv_[I] == bus_vn_kv_[I]` for `I < n_sub`, which `check_valid()` now enforces when it is present. **It therefore has to stay optional**: requiring it would reject every existing grid and every already-saved file. A `TODO` on the member and an entry in the CHANGELOG's `[TODO]` section record the two ways out — drop it from `StateRes` (needs a format bump), or have `init_bus` fill it and require it (changes what newly-saved files contain, and can only become mandatory once no already-saved file needs to load). That call is left to you.

## Verification

- Every PoC now raises a clean `IndexError` / `RuntimeError` instead of crashing, across all three channels (python property, `pickle.loads`, `LSGrid.load_binary`).
- **No behaviour change on real grids**: case14 / case118 / case300 (pandapower), ieee14 (pypowsybl) and the shipped format-4 binary fixture all still pass `check_grid()`.
- C++ suite: **108 cases green** (~35 added here). `[check_grid]` under **valgrind: 0 errors from 0 contexts**.
- Python: the same **123 pre-existing** failures as the unmodified tree, **zero new** — verified by building the pre-fix tree and diffing the failure lists, which come back identical. Those 123 are environmental in this container (missing grid2op data, missing `old_pickle/` fixtures, no KLU in this build).

## Worth knowing before merging

The per-substation voltage invariant has been unenforced for as long as that loop has existed, so if any grid in the wild violates it, this turns a silent acceptance into a hard failure on load. Every grid I could test is clean, but only pandapower and pypowsybl were available in this container — **matpower and powermodels went untested**.

https://claude.ai/code/session_015CfJVFAd2ifp4qyZxp5Ec5

---
_Generated by [Claude Code](https://claude.ai/code/session_015CfJVFAd2ifp4qyZxp5Ec5)_

### DIFF
--- a/src/core/LSGrid.hpp
+++ b/src/core/LSGrid.hpp
@@ -1824,7 +1824,13 @@ class LS2G_API LSGrid final
         }
 
     protected:
-        void set_ls_to_orig_internal(const Eigen::Ref<const IntVect> & ls_to_orig) noexcept;  // set both _ls_to_orig and _orig_to_ls
+        // set both _ls_to_orig and _orig_to_ls. `noexcept`, and it indexes
+        // _orig_to_ls with the values of `ls_to_orig`: only ever call it on a vector
+        // that check_ls_to_orig_values() has already accepted (set_ls_to_orig does).
+        void set_ls_to_orig_internal(const Eigen::Ref<const IntVect> & ls_to_orig) noexcept;
+        // throws std::out_of_range unless every entry is -1 or a sane, non-negative
+        // original-grid bus id -- see the definition in LSGrid.cpp for why.
+        static void check_ls_to_orig_values(const Eigen::Ref<const IntVect> & ls_to_orig);
 
         // init the Ybus matrix (its size, it is filled up elsewhere) and also the 
         // converter from "my bus id" to the "solver bus id" (id_me_to_solver and id_solver_to_me)
--- a/src/core/LSGrid.hpp
+++ b/src/core/LSGrid.hpp
@@ -1824,10 +1824,19 @@ class LS2G_API LSGrid final
         }
 
     protected:
-        // set both _ls_to_orig and _orig_to_ls. `noexcept`, and it indexes
-        // _orig_to_ls with the values of `ls_to_orig`: only ever call it on a vector
-        // that check_ls_to_orig_values() has already accepted (set_ls_to_orig does).
-        void set_ls_to_orig_internal(const Eigen::Ref<const IntVect> & ls_to_orig) noexcept;
+        // Set both _ls_to_orig and _orig_to_ls.
+        //
+        // NOT noexcept (it used to be declared so, wrongly): it assigns one Eigen
+        // vector and allocates another, either of which throws std::bad_alloc on
+        // failure -- and a throw out of a noexcept function is not an error the
+        // caller can handle, it is an immediate std::terminate(). Nothing needs the
+        // guarantee (the only callers are the copy constructor and set_ls_to_orig,
+        // neither of them noexcept), so the honest declaration is the plain one.
+        //
+        // It indexes _orig_to_ls with the *values* of `ls_to_orig`: only ever call
+        // it on a vector check_ls_to_orig_values() has already accepted, as
+        // set_ls_to_orig does.
+        void set_ls_to_orig_internal(const Eigen::Ref<const IntVect> & ls_to_orig);
         // throws std::out_of_range unless every entry is -1 or a sane, non-negative
         // original-grid bus id -- see the definition in LSGrid.cpp for why.
         static void check_ls_to_orig_values(const Eigen::Ref<const IntVect> & ls_to_orig);
--- a/src/core/SubstationContainer.cpp
+++ b/src/core/SubstationContainer.cpp
@@ -9,7 +9,9 @@
 #include "SubstationContainer.hpp"
 #include "BinaryArchive.hpp"
 
+#include <cstdint>
 #include <iostream>
+#include <limits>
 #include <sstream>
 
 namespace ls2g {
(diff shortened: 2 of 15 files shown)


src/core/LSGrid.hpp
```diff
@@ -1824,7 +1824,13 @@ class LS2G_API LSGrid final
         }
 
     protected:
-        void set_ls_to_orig_internal(const Eigen::Ref<const IntVect> & ls_to_orig) noexcept;  // set both _ls_to_orig and _orig_to_ls
+        // set both _ls_to_orig and _orig_to_ls. `noexcept`, and it indexes
+        // _orig_to_ls with the values of `ls_to_orig`: only ever call it on a vector
+        // that check_ls_to_orig_values() has already accepted (set_ls_to_orig does).
+        void set_ls_to_orig_internal(const Eigen::Ref<const IntVect> & ls_to_orig) noexcept;
+        // throws std::out_of_range unless every entry is -1 or a sane, non-negative
```
BDonnot: A noexcept function that throws ? Isn't it forbidden?

src/core/SubstationContainer.cpp
```diff
@@ -34,30 +36,180 @@ SubstationContainer::StateRes SubstationContainer::get_state() const
...
+            std::ostringstream exc_;
+            exc_ << "SubstationContainer::set_state: n_sub (" << n_sub << ") * nmax_busbar_per_sub ("
+                 << nmax_busbar_per_sub << ") overflows the maximum number of buses representable.";
+            throw std::runtime_error(exc_.str());
+        }
+    }
+    const std::int64_t n_bus_max = static_cast<std::int64_t>(n_sub) * static_cast<std::int64_t>(nmax_busbar_per_sub);
 
-    // check sizes
-    // TODO dev switches
```
BDonnot, line 47: Keep the Todo dev switches please (keep the comment)

src/core/SubstationContainer.cpp
```diff
@@ -34,30 +36,180 @@ SubstationContainer::StateRes SubstationContainer::get_state() const
...
+        exc_ << "LSGrid::check_grid: the per-bus max voltage vector has " << bus_vmax_kv_.size()
+             << " entries for " << bus_vn_kv_.size() << " buses (it must be either empty or complete).";
+        throw std::runtime_error(exc_.str());
+    }
+
+    // All busbars of a substation share its nominal voltage. init_bus() is supposed
+    // to enforce this at build time, but set_state() bypasses init_bus() entirely,
+    // so a pickle / binary file could always carry a grid violating it. Re-check
+    // here, which is the point of check_grid().
+    check_bus_vn_kv_uniform_per_sub();
```
BDonnot, line 180: We should also check that vn_kv are > 0

src/core/SubstationContainer.hpp
```diff
@@ -92,13 +113,31 @@ class LS2G_API SubstationContainer final : public IteratorAdder<SubstationContai
...
             if(sub_vn_kv.size() != n_sub_){
                 throw std::range_error("SubstationContainer::init_sub: sub_vn_kv should have the size of the number of substations on the grid.");
             }
+            // Both destinations are written with an unchecked operator[] below, and
+            // neither is guaranteed to be sized yet: only the two-argument
+            // constructor sizes sub_vn_kv_, and the live construction path is the
+            // default constructor followed by init_bus(), which leaves sub_vn_kv_
+            // EMPTY. Writing n_sub_ doubles into it would be an out-of-bounds heap
+            // write. Size both here instead of assuming.
+            const Eigen::Index nb_bus_total =
```
BDonnot: How about the "it will not overflow" check

src/core/SubstationContainer.hpp
```diff
@@ -183,20 +222,46 @@ class LS2G_API SubstationContainer final : public IteratorAdder<SubstationContai
...
+         * `sub_id + (local_bus_id - 1) * n`), and the loop ran local ids `[1, k)`, so
+         * it opened by comparing bus `I` with itself and stopped one busbar short.
+         * At the usual k == 2 that compared bus `I` to bus `I` and checked nothing at
+         * all -- the invariant was silently unenforced on every path.
+         */
+        void check_bus_vn_kv_uniform_per_sub() const
+        {
+            const int n = n_sub_;
+            const int k = nmax_busbar_per_sub_;
+            for(int I = 0; I < n; ++I){
```
BDonnot: I lowercase would be better

BDonnot (commented):
Also, the enum about algorithm is used to distinguish between lightsim2grid "internal" solvers and the "plug-in" solvers which is wrong for some of them (at least the "try with factor if refactor fails" which is not match to a enum). 
This means the consistency check about the size of V and the absence of NaN in the returned V is performed where it should not

src/core/LSGrid.hpp
BDonnot: You're right, and it's worse than a doc inconsistency — the `noexcept` was a lie.

`set_ls_to_orig_internal` assigns one Eigen vector and allocates another:

```cpp
_ls_to_orig = ls_to_orig;                        // allocates
_orig_to_ls = IntVect::Constant(size + 1, -1);   // allocates
```

Both throw `std::bad_alloc` on failure. It isn't forbidden by the compiler — it compiles fine — but a throw escaping a `noexcept` function calls `std::terminate()` directly, so instead of an exception the caller could handle, it kills the process. Not a good trade for a function with no need of the guarantee.

Dropped it (the two callers, the copy constructor and `set_ls_to_orig`, are not `noexcept` themselves, so nothing required it). The comment now says why it is deliberately absent, so it doesn't get re-added.

The neighbouring `check_ls_to_orig_values` is the one documented as throwing — it is not, and must not be, `noexcept`.

---

src/core/SubstationContainer.cpp
BDonnot: Restored. I put it back on the size checks it actually refers to, rather than at the top of the function where the validation now starts with the positivity / overflow guard:

```cpp
    // check sizes
    // TODO dev switches
    // bus_vn_kv_ defines nb_bus(), the bound every other index is checked against;
    // bus_status_ is indexed with those same ids and MUST match it exactly.
    const auto check_len = [](std::size_t actual, std::int64_t expected, const char * name){
```

Say the word if you'd rather have it verbatim at the original spot.

---

src/core/SubstationContainer.cpp
BDonnot: Added, as `check_vn_kv_positive()` — finite **and** strictly positive — called from both `init_bus()` and `check_valid()`.

One deliberate restriction: it covers `bus_vn_kv_` and `sub_vn_kv_` only, **never** `bus_vmin_kv_` / `bus_vmax_kv_`. Those use NaN as the documented "no limit set for this bus" sentinel, so checking them would re-introduce exactly the false positives that d324994 had to back out (non-finite thermal limits and reactive bounds are legitimate on real grids). A *nominal* voltage has no such convention — 0 or negative is simply meaningless, and it silently corrupts the per-unit conversion since `v_kv_from_vpu` multiplies by it. The reasoning is in the comment so nobody widens it later.

Verified rejected: `0`, `-138`, `NaN`. Verified still accepted: case14 / case118 / case300 (pandapower), ieee14 (pypowsybl), and the shipped format-4 binary fixture.

---

src/core/SubstationContainer.hpp
BDonnot: Good catch — I'd only put the overflow check in `set_state`, so `init_bus` and `init_sub` were still computing `n_sub * nmax_busbar_per_sub` in plain `int`. `init_bus` in particular did

```cpp
if(bus_vn_kv.size() != n_sub * nmax_busbar_per_sub){   // int overflow, UB
```

Rather than repeat it a third time, all three now share one helper:

```cpp
static int checked_nb_bus(int n_sub, int nmax_busbar_per_sub, const char * caller);
```

It does the positivity checks (`n_sub > 0` — it is also the `sub_id_of_bus` modulo divisor — and `nmax_busbar_per_sub > 0`), multiplies in `int64_t`, and refuses a product that does not fit the `int` the count is stored in and indexed with. `init_bus`, `init_sub` and `set_state` all call it, so a fourth entry point cannot forget.

Verified: `init_bus(100000, 100000, ...)` is now rejected with a clear message instead of overflowing to a negative bus count.

---

src/core/SubstationContainer.hpp
BDonnot: Done — the loop variable is `i` now, and I renamed it in the surrounding doc comment too so the formula and the code use the same letter:

```
for all 0 <= i < n, for all 0 <= j < k:  bus_vn_kv[i] == bus_vn_kv[i + j*n]
```

---

BDonnot:
## On the enum being the wrong way to tell built-in from plugin

You're right, and I confirmed it mechanically — diffing the names `BuiltinSolversRegistration.cpp` registers against the names `name_to_algo_type()` maps:

```
built-in solvers with NO AlgorithmType member (=> reported as Custom, ie "plugin"):
  NRRefactorRetry_KLU
  NRRefactorRetry_NICSLU
  NRRefactorRetry_CKTSO
```

Exactly the family you named. So `LSGrid::process_results` gating on

```cpp
const bool is_external_algo = (ac ? _algo.get_type() : _dc_algo.get_type()) == AlgorithmType::Custom;
```

ran the voltage size + finiteness check on three **in-tree** solvers, on every single solve — the opposite of the "built-in solvers pay nothing for it" claim in `docs/security.rst`.

### The fix

The root problem is the one you put your finger on: `AlgorithmType` is a fixed enum of *serialized solver identities*, so it can only answer "is this solver ours?" for solvers someone remembered to add a member for. Adding three members would have fixed today's symptom and left the trap armed for the next built-in.

So origin is now recorded where it is actually known — at registration:

```cpp
enum class SolverOrigin { Builtin, External };
```

- `AlgorithmRegistry` stores one per registered name, exposing `origin_of()` / `is_builtin()`.
- The default is **`External`**, deliberately: External is the cautious answer (it is what gets a solver's output validated), so anything registered without saying otherwise stays checked. Only `BuiltinSolversRegistration.cpp` opts out.
- All 25 built-in registrations go through a single tagging helper in that file, so a built-in added later cannot be misclassified by forgetting the argument.
- `AlgorithmSelector` caches the flag in `change_algorithm()` and exposes `is_builtin_algo()`, so the hot path stays a bool read rather than a per-solve string hash.

`get_type() == Custom` is no longer used for this anywhere.

### Tests

Two new C++ cases. The important one asserts the decoupling directly: for whichever `NRRefactorRetry_*` this build registers, `name_to_algo_type(name) == AlgorithmType::Custom` **and** `registry.is_builtin(name)` — i.e. the enum still says "plugin" while the registry correctly says "ours". It also checks a plugin-style registration stays External, an unregistered name reports External, and the selector's cached flag follows `change_algorithm` both ways.

Those three solvers need KLU / NICSLU / CKTSO, none of which this container has, so that assertion is skipped locally and will first really run in CI.

### Everything else from the review

| | |
|---|---|
| `noexcept` that throws | It genuinely could — it allocates, and `bad_alloc` out of `noexcept` is `std::terminate()`. Dropped. |
| Keep `TODO dev switches` | Restored, on the size checks it refers to. |
| `vn_kv > 0` | Added (finite **and** > 0), for `bus_vn_kv` / `sub_vn_kv` only — `bus_vmin_kv` / `bus_vmax_kv` use NaN as "no limit", the same distinction that made d324994 back out the element-level finiteness checks. |
| overflow check | Was only in `set_state`; `init_bus` / `init_sub` now share one `checked_nb_bus` helper with it. |
| lowercase `I` | Done, in the doc formula too. |

Verified: NaN / 0 / negative `vn_kv` and the overflowing product are rejected; case14 / case118 / case300 (pandapower), ieee14 (pypowsybl) and the format-4 fixture all still pass `check_grid()`. C++ **110 cases** green, `[check_grid]` clean under **valgrind (0 errors)**, and the python run shows the same **123 pre-existing** environment failures as the unmodified tree — none new.

---

src/core/SubstationContainer.cpp
```diff
@@ -34,30 +36,166 @@ SubstationContainer::StateRes SubstationContainer::get_state() const
...
+    // `nb_bus()` (= bus_vn_kv_.size()) is the upper bound LSGrid::check_grid()
+    // validates every element's bus id against, while `bus_status_` is what
+    // `is_bus_connected()` / `disconnect_all_buses()` actually index with those same
+    // ids, and `n_sub_` is both the substation-id bound and a modulo divisor
+    // (sub_id_of_bus). Nothing downstream re-derives these from each other, so if
+    // they are allowed to disagree here, a *validated* bus id still reads and writes
+    // past the end of bus_status_ -- release wheels are built -O3 -DNDEBUG, so
+    // neither Eigen nor the standard library catches it. Validate them all now,
+    // before anything is assigned.
+    const int n_sub = std::get<0>(my_state);
```
BDonnot, line 50: Instead of std::get<0>(...) can you declare (constexpr or static const, members or "global" variables, something like constexpr int N_SUB_ID = 0; and then use std::get<N_SUB_ID>(...) like it is done in the other containers

BDonnot (commented):
Ok

PR #157, opened 2026-07-28
Second security audit: six more holes on the untrusted-input paths

BDonnot:
Follow-up to the audit merged in #155 / #156, over the channels `docs/security.rst` covers: pickle, the fast binary format, and the python bindings including the grid loaders.

## Memory-safety holes (commit `ac5399f`)

Five issues, each **reproduced as a segfault or as heap corruption** before the fix:

| Where | Reachable from | Effect |
|---|---|---|
| `TrafoContainer::set_state` — the two per-transformer `alpha -> r/x correction` tables copied with no check, then `_update_model_coeffs()` indexes them per transformer with an unchecked `operator[]` | pickle, `.lsb` | builds a `std::vector` out of heap memory past the end of the array and dereferences its pointers → **segfault**. Runs *inside* `set_state`, before `check_grid()` sees the grid |
| `TwoSidesContainer::set_tsc_state` never checked `status_global_`'s length (`nb()` is `side_1_.nb()`) | pickle, `.lsb` | that vector is indexed with element ids bounded by `nb()` throughout the class and the batch solvers → **segfault**. Affects powerlines, transformers and hvdc lines |
| `check_grid()` did not range-check `SvcContainer::regulated_bus_id_` | pickle, `.lsb`, **and `init_svcs` / `init_from_pypowsybl`** | it is a gridmodel bus id used directly as an index — `id_grid_to_solver[...]` in `SvcContainer::set_vm`, followed by a **write into `V`** → **segfault on `ac_pf`**. The generator field of the same name was already validated |
| `check_grid()` collected `pos_topo_vect` from every container, so the `dim_topo` it proved the permutation against could exceed the one `update_topo()` sizes its caller arrays with (which excludes shunts, sgens, svcs, hvdc) | pickle, `.lsb` | a *validated* load position could still be past the end of the array `update_topo()` indexes with it → **segfault** |
| `PandaPowerConverter` never checked that its inputs have the same length (there was a `TODO check all vectors have the same size` where the check belonged) | **plain python + `init_from_pandapower`** | element count comes from the first argument, the rest are read with unchecked accessors and combined in Eigen expressions sized from one of them → writes past the end of the shorter arrays, reproducible as `free(): invalid next size (fast)` |

The last one is reachable from `init_from_pandapower` on a net with a **duplicated bus label**: the loader resolves branch end voltages with `pp_net.bus.loc[pp_net.trafo["hv_bus"]]`, which then returns more rows than there are branches.

## Silent wrong answers (commit `4f339bd`)

A second pass over the Ybus / Sbus construction, looking for divisions by zero and degenerate physical parameters.

**No memory-safety problem there.** Every division by a physical parameter (`1/(r+jx)`, `1/x`, `1/tau`, `/sn_mva`) is floating point, so it yields Inf/NaN rather than a signal, and none of those values is used as an index. 30+ degenerate cases (`r`/`x`/`ratio`/`vn_kv` = 0, negative, NaN, Inf) produced no crash and no hang.

What it did find is a *silent* failure mode, in the one place the solvers' guards cannot see. `sn_mva` is the base power of the whole per-unit system, and it was validated **nowhere** — not by `set_sn_mva`, not by `check_grid()`, and the loaders take it verbatim from their source file (`init_from_powermodels` / `init_from_matpower` do `set_sn_mva(float(network["baseMVA"]))`):

* `sn_mva = NaN` or `Inf` → the **DC powerflow reports CONVERGENCE and returns NaN branch flows**. The built-in solvers' finiteness guards cannot catch it: they inspect `Va`, which is perfectly finite (`Bbus` does not involve `sn_mva`), and the NaN only appears afterwards when results are scaled back to MW. The size/finiteness check at the solver→grid seam is deliberately reserved for external solvers.
* `sn_mva < 0` → a plausible-looking but sign-inverted per-unit system.

`sn_mva` and `init_vm_pu` must now be finite and strictly positive, checked by their setters and re-checked by `check_grid()` (which is what covers pickle, the binary format and every loader, since `set_state` bypasses the setters). `PandaPowerConverter::_check_init` had the same shape of bug for its own `sn_mva` / `f_hz`: it tested `<= 0.`, and every comparison with NaN is false, so a NaN sailed through both — and they are divisors in every method of that class.

## Deliberately *not* changed

`check_grid()` still does not validate the *values* of `r` / `x` / `h` / tap ratio. #156 reverted exactly such a check over false positives, and that reasoning still holds — pypowsybl encodes "no thermal limit" as a non-finite `limit_a_ka`, unbounded reactive limits are ±Inf. Two further invariants were tried, implemented, and **reverted after the test suite disproved them**:

* **`r, x >= 0`** — no. A negative reactance is a series capacitor, and negative r/x also fall out of 3-winding-transformer star equivalents. pandapower's own `case300` has 1 line with negative x; `case9241pegase` has 14 with r < 0 and 16 with x < 0.
* **`r == x == 0` rejected on a connected branch** — a real degenerate case (DC reports convergence with non-finite flows), but not rejectable: `test_bus_fusion_pypowsybl` asserts that an ideal 225/90 kV transformer *must stay connected*, and that with `fuse_zero_impedance_branches=False` the grid **loads** and the powerflow then fails. A NaN check on `h` was also wrong — a zero-impedance line's per-unit conversion legitimately yields `g = NaN` in pypowsybl data.

The zero-impedance case is documented as a known limitation in `docs/security.rst` instead, with the guidance to check that returned flows are finite.

## How it was found

Reading, plus four mechanical sweeps run per-case in a subprocess so a crash is a signal rather than a lost run:

* **pickle-state structural sweep** — every leaf of the nested `StateRes` truncated / extended / poisoned (640 cases).
* **paired sweep** — every boolean flag flipped *and* a sibling array poisoned (415 cases), since feature flags gate whole code paths.
* **`.lsb` byte-level sweep** — 14 443 single-byte pokes, truncations and random corruption. **0 crashes**: the byte-level reader is solid.
* **degenerate-physics probe** — the r/x/ratio/vn_kv/sn_mva matrix above.

All sweeps come back clean after the fixes.

## Tests

`src/tests/test_state_poisoning.cpp` (17 cases, in the suite that also runs under valgrind/ASan in `cpp_unit_tests`) and `lightsim2grid/tests/test_state_poisoning.py` (28 cases).

## Verification

* **C++ suite: 127 cases / 1433 assertions green.**
* **Python suite: 1226 passed, 0 failed.**

Excluded from that python run are `test_LightSimBackend.py`, `test_Runner.py` and `test_RedispatchEnv.py` — all 99 of their failures in this sandbox are environmental (grid2op's `data_test/` fixtures are absent from the pip wheel, and `grid2op.make` cannot reach `api.github.com`), and they fail during env construction before a lightsim2grid backend is built. Worth confirming on real CI.

https://claude.ai/code/session_019K1jRmqbpB4JUsuHHy4t9A

---
_Generated by [Claude Code](https://claude.ai/code/session_019K1jRmqbpB4JUsuHHy4t9A)_

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -134,6 +134,25 @@ TODO: Levenberg-Marquardt damping (a.k.a. Tikhonov-regularized Newton) : adding 
   / ``init_from_powermodels``). A well-formed but inconsistent state (e.g. an
   out-of-range bus id in a crafted binary file) now raises a clean exception instead
   of causing an out-of-bounds access during the next powerflow.
+- [FIXED] ``sn_mva`` (the base power of the whole per-unit system) and ``init_vm_pu``
+  (the flat-start voltage magnitude) were validated **nowhere**: not by their python
+  setters, not by ``check_grid()``, and the loaders take them verbatim from their source
+  file (``init_from_powermodels`` / ``init_from_matpower`` do
+  ``set_sn_mva(float(network["baseMVA"]))``). A degenerate value does not make the
+  powerflow fail, it makes it *quietly wrong*: with ``sn_mva = NaN`` the **DC powerflow
+  reported convergence and returned NaN branch flows**, and with a negative one it
+  returned a plausible-looking but sign-inverted per-unit system. The built-in solvers'
+  own finiteness guards cannot catch this -- they check ``Va``, which is finite, since
+  ``Bbus`` does not involve ``sn_mva``; the NaN only appears afterwards, when the results
+  are scaled back to MW / MVar -- and the size/finiteness check on the solver output is
+  deliberately reserved for external solvers. Both scalars must now be finite and
+  strictly positive, checked by the setters and re-checked by ``check_grid()`` (which is
+  what covers pickle, the binary format and every loader, since ``set_state`` bypasses
+  the setters).
+- [FIXED] ``PandaPowerConverter::_check_init`` tested ``sn_mva_ <= 0.`` / ``f_hz_ <= 0.``.
+  Every comparison with NaN is false, so a NaN sailed through both -- and they are
+  divisors in every method of that class, so the whole converted grid came back NaN with
+  no error raised anywhere. Both are now checked as finite and strictly positive.
 - [FIXED] ``PandaPowerConverter`` (``get_trafo_param_pp2`` / ``get_trafo_param_pp3`` /
   ``get_line_param`` / ``get_line_param_legacy``, all exposed to python) never checked
   that the arrays it is given have the same length -- there was a ``TODO check all
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -134,6 +134,55 @@ TODO: Levenberg-Marquardt damping (a.k.a. Tikhonov-regularized Newton) : adding 
   / ``init_from_powermodels``). A well-formed but inconsistent state (e.g. an
   out-of-range bus id in a crafted binary file) now raises a clean exception instead
   of causing an out-of-bounds access during the next powerflow.
+- [FIXED] ``PandaPowerConverter`` (``get_trafo_param_pp2`` / ``get_trafo_param_pp3`` /
+  ``get_line_param`` / ``get_line_param_legacy``, all exposed to python) never checked
+  that the arrays it is given have the same length -- there was a ``TODO check all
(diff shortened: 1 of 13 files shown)


BDonnot (commented):
Ok

PR #158, opened 2026-07-30
Documentation audit fixes for the 1.0 release (n1_full_compute)

BDonnot:
## Summary

Full pass over `docs/*.rst`, `README.md`, `DISCLAIMER.md`, `examples/`, and `benchmarks/` ahead of the 1.0 release, driven by an initial audit organized into sections A–K. Each section below was fixed, verified, committed, and pushed incrementally. Section **F** (benchmark tables contradicting their own prose) is intentionally **not** included — the user is handling that in a dedicated session where the benchmark scripts will be re-run to produce the correct numbers directly.

- **A. Wrong API** (A1–A24): code samples that would fail if copied (wrong imports, wrong kwargs, swapped `get_Va`/`get_Vm`, stale `compute_pf` signature, wrong solver counts, dead file links, a copy-pasted CKTSO/NICSLU env var bug, etc).
- **B. Version identity**: unified the release version to `1.0.0` everywhere (docs, `__init__.py`, changelog).
- **C. Rendering failures**: fixed real Sphinx build bugs (leaked markdown links, missing blank lines before directives, misplaced backticks, title underlines, dead `conf.py` options).
- **D. Dangling cross-references**: fixed via a temporary `nitpicky=True` sweep; also exported a genuinely-missing `TimerJac`.
- **E. Deprecated names**: swept `SecurityAnalysis`/`gridmodel`/`GridModel`/`get_solver_type` out of prose, example output, and benchmark scripts; removed dead `except ImportError` fallbacks; fixed a real `AttributeError` bug in `benchmark_grid_size.py`.
- **H. Stale statements**: removed claims that are no longer true (an obsolete 0.5.5 bug warning, wrong solver-family counts, a `reset()`-reverts-your-change bug in a doc example, stale env names, README's O2/O3 and `setup.py` claims, ~70 lines of pre-1.0 SuiteSparse install legacy, `docs/disclaimer.rst` drift from `DISCLAIMER.md`).
- **I. Example/benchmark structural issues**: found and fixed a real bug where `init_from_n_powerflow` was set *after* the computation it configures (verified ~3x timing difference); a real CMake bug where a plugin's default `-O3` didn't match `lightsim2grid_core`'s; added a working "advanced usage" demo to `contingency_analysis.py`.
- **J. Packaging metadata**: fixed `requires-python`, `build.verbose`; kept `Development Status :: 4 - Beta` per explicit instruction (not stable yet even at 1.0).
- **K. Typos**: ~24 typos fixed, including two real `NameError` bugs duplicated across three files.
- **G. Missing documentation** (largest section): the big find — `ContingencyAnalysis` and `TimeSerie` were defined as `class __ContingencyAnalysis` / `class ___TimeSerie` and aliased to their public names; Sphinx autodoc treats that as a plain alias and silently drops all method docs, so **the two main user-facing classes of the package had no API reference anywhere**. Renamed both directly, which surfaces `nb_thread`, `handle_disconnected_grid`, `compute_limit_violations`, `run_ac`/`run_dc`, etc. for the first time. Also fixed a fully-broken `PhysicalLawChecker` (wrong grid2op import paths + missing `load_grid_public`/`assert_grid_correct` setup), documented `ScalingPolicyType`/`RefactorPolicyType`/`FDPFMethod`/`AlgoConfig`, `SubstationContainer`/`SubstationInfo`, PTDF/LODF, the third solver-plugin example, `pandapower_compat`, corrected `LightSimBackend` kwargs docs, added a "Key features" section to README, and explained the OLF outer-loop vs. lightsim2grid single-solve architecture difference.
- Also fixed two CircleCI failures on `compile_gcc_earliest` unrelated to docs (Python 3.7 default, `uv` install failure, and an LTO-related OOM) and a `-Wsuggest-override` regression on three element containers.

## Verification

Every change was verified before committing: `sphinx-build -b html -W -E` (zero warnings) after each doc change, `py_compile`/`pyflakes` on every touched Python file, and targeted `unittest` runs for every module touched (100+ tests across `ContingencyAnalysis`, `TimeSerie`, `Checker`, SVC, HVDC, scaling/refactor policies — all green). Several fixes were verified by actually running the corrected code (e.g. the `init_from_n_powerflow` ordering bug, the `PhysicalLawChecker` fix reporting a ~1e-12 KCL mismatch on a real converged powerflow, the CMake `-O3` default fix via a standalone CMake project).

## Test plan

- [x] `sphinx-build -b html -W -E docs <out>` — zero warnings
- [x] `py_compile` / `pyflakes` on every touched Python file
- [x] Targeted `unittest` runs for every touched module (100+ tests, all passing)
- [x] Manual verification of corrected runtime behavior where a real bug was found

---
_Generated by [Claude Code](https://claude.ai/code/session_01FSFvKnbpy6u4oSKJc5JaXf)_

### DIFF
--- a/docs/algorithm_names.rst
+++ b/docs/algorithm_names.rst
@@ -191,7 +191,7 @@ Usage example
 Migration from old names
 ------------------------
 
-Before version 0.14, the enum values and class names used shorter names that
+Before version 1.0.0.rc2, the enum values and class names used shorter names that
 did not make the algorithm component explicit (``KLU``, ``SparseLU``, ``DC``,
 etc.).  These old names are kept as **deprecated aliases** in the Python enum
 so that existing code continues to work, but they will be removed in a future
--- a/docs/algorithm_names.rst
+++ b/docs/algorithm_names.rst
@@ -175,11 +175,15 @@ String-only built-in solvers: ``NRRefactorRetry_*``
 ``NRRefactorRetry_KLU``, ``NRRefactorRetry_NICSLU`` and ``NRRefactorRetry_CKTSO`` are
 built-in solvers that never got their own ``AlgorithmType`` enum value, so unlike
 everything in the table above they can only be selected by their string name, not by an
-``AlgorithmType.NRRefactorRetry_*`` attribute (which does not exist):
+``AlgorithmType.NRRefactorRetry_*`` attribute (which does not exist). Since lightsim2grid
+1.0, :func:`~lightsim2grid.lightSimBackend.LightSimBackend.set_algo_type` accepts such a
+string directly, and -- unlike changing the algorithm through the private
+``env.backend._grid.change_algorithm(...)`` -- remembers the choice so it is re-applied
+after every ``env.reset()`` and preserved by ``backend.copy()``:
 
 .. code-block:: python
 
-    env.backend._grid.change_algorithm("NRRefactorRetry_KLU")
+    env.backend.set_algo_type("NRRefactorRetry_KLU")
 
 Each is the same algorithm as its ``NR_*`` counterpart (Newton-Raphson, distributed
 slack), except that when a Jacobian ``refactorize()`` fails it falls back to a full
@@ -191,20 +195,10 @@ same value reported for an actual plugin solver.
 
 .. warning::
     ``env.backend._grid.change_algorithm(...)`` acts directly on the underlying
-    C++ grid model and does **not** survive ``env.reset()``: a reset always
-    re-applies whichever :class:`~lightsim2grid.algorithm.AlgorithmType` was last
-    set through :func:`~lightsim2grid.lightSimBackend.LightSimBackend.set_algo_type`
-    (:class:`~lightsim2grid.algorithm.AlgorithmType.NRSing_KLU` by default), silently
-    discarding the string-only solver choice.
-
-    Unlike the enum-based solvers, ``set_algo_type`` cannot be used as a
-    persistent alternative here: it requires an actual ``AlgorithmType`` value
-    and raises a ``BackendError`` if given anything else, so it has no way to
-    represent ``NRRefactorRetry_*`` (which have no enum value at all, see the
-    ``Custom`` entry above). At the moment there is no supported way to make a
-    string-only solver choice survive a reset: you need to call
-    ``env.backend._grid.change_algorithm(...)`` again after every
-    ``env.reset()`` if you want to keep using it.
+    C++ grid model and does **not** survive ``env.reset()`` / ``backend.copy()``: use
+    the persistent :func:`~lightsim2grid.lightSimBackend.LightSimBackend.set_algo_type`
+    shown above instead (it works the same for a plain :class:`~lightsim2grid.algorithm.AlgorithmType`
+    or a string-only / plugin name).
 
 Usage example
 -------------
--- a/docs/algorithm_names.rst
+++ b/docs/algorithm_names.rst
@@ -161,12 +161,40 @@ Complete table
    * - ``GaussSeidelSynch``
      - :class:`~lightsim2grid.algorithm.GaussSeidelSynchAlgo`
      - Synchronous Gauss-Seidel iterative
+   * - ``Custom``
+     - --
+     - Sentinel value reported by :func:`~lightsim2grid.lightSimBackend.LightSimBackend.get_algo_types`
+       / ``get_algo_type()`` for any solver loaded through the plugin mechanism (see
+       :ref:`solver_plugin`). It is also, incidentally, what the three built-in
+       ``NRRefactorRetry_*`` solvers below report, since they were never given their own
+       enum value either.
+
+String-only built-in solvers: ``NRRefactorRetry_*``
+----------------------------------------------------
+
+``NRRefactorRetry_KLU``, ``NRRefactorRetry_NICSLU`` and ``NRRefactorRetry_CKTSO`` are
+built-in solvers that never got their own ``AlgorithmType`` enum value, so unlike
+everything in the table above they can only be selected by their string name, not by an
+``AlgorithmType.NRRefactorRetry_*`` attribute (which does not exist):
+
+.. code-block:: python
+
+    env.backend._grid.change_algorithm("NRRefactorRetry_KLU")
+
+Each is the same algorithm as its ``NR_*`` counterpart (Newton-Raphson, distributed
+slack), except that when a Jacobian ``refactorize()`` fails it falls back to a full
+``factorize()`` (reusing the existing symbolic factorization) before giving up, instead
+of reporting an error immediately. Use ``get_linear_solver_stats()`` on the solver to
+see how often the fallback fires. Because they have no dedicated enum value,
+``get_algo_type()`` reports ``AlgorithmType.Custom`` while one of them is active -- the
+same value reported for an actual plugin solver.
 
 Usage example
 -------------
 
 .. code-block:: python
 
+    import grid2op
     import lightsim2grid
     from lightsim2grid.algorithm import AlgorithmType
 
@@ -177,7 +205,7 @@ Usage example
 
     # inspect what algorithm is currently active
     print(env.backend.get_algo_types())
-    # >>> (<AlgorithmType.NRSing_KLU: 7>, <AlgorithmType.DC_KLU: 9>)
+    # >>> (<AlgorithmType.NR_KLU: 1>, <AlgorithmType.DC_KLU: 9>)
 
     # list all algorithms available in this build
     env.backend._grid.available_algorithm_names()
--- a/docs/algorithm_names.rst
+++ b/docs/algorithm_names.rst
@@ -235,7 +235,7 @@ The same renaming affects two other places that predate this convention:
   ``AlgorithmType``. A ``SolverType`` value is still accepted wherever an
   ``AlgorithmType`` is expected (it is converted automatically, with a
   ``DeprecationWarning``);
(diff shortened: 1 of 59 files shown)


docs/algorithm_names.rst
```diff
@@ -161,12 +161,40 @@ Complete table
...
+----------------------------------------------------
+
+``NRRefactorRetry_KLU``, ``NRRefactorRetry_NICSLU`` and ``NRRefactorRetry_CKTSO`` are
+built-in solvers that never got their own ``AlgorithmType`` enum value, so unlike
+everything in the table above they can only be selected by their string name, not by an
+``AlgorithmType.NRRefactorRetry_*`` attribute (which does not exist):
+
+.. code-block:: python
+
+    env.backend._grid.change_algorithm("NRRefactorRetry_KLU")
```
BDonnot: This formulation is misleading (a reset will change the algorithm) prefer showing (here but if that is the case somewhere else) the "recommended" way, in this case : env.backend.change_algorithm(...) [or env.backend.change_algo(...) I don't remember).
And then of course env.reset() (otherwise it does not have any effect)

docs/physical_law_checker.rst
```diff
@@ -5,8 +5,6 @@ This module allows to check if a given vector checks the KCL (Kirchhoff Current
 
 It is a wrapper around the :func:`lightsim2grid.network.LSGrid.check_solution` function.
 
-TODO DOC in progress !
-
 .. warning::
     The grid2op environment is read from a `grid.json`` file. Make sure to use an environment that can be
```
BDonnot: This is no longer true, lightsim2grid can be init from lots of different format now

docs/solvers.rst
```diff
@@ -230,6 +232,78 @@ You can also change the algorithm after creation (not recommended, but supported
    :ref:`algorithm_names`.
 
 
+Fine-tuning the Newton-Raphson iteration
+------------------------------------------
```
BDonnot, line 236: Reading that now makes me wonder if these are properly kept when resetting or copying the backend? I'm speaking of all these customization

docs/use_solver.rst
```diff
@@ -89,13 +98,13 @@ All algorithms can be accessed with the same API (if you want to use the raw pyt
...
-  solver.get_Vm()  # voltage angle
+  solver.get_Va()  # voltage angle
+  solver.get_Vm()  # voltage magnitude
   solver.get_V()  # complex voltage
   # for compatible solvers
-  solver.get_J()  # see documentation of the `newton_pf` function for more information about the shape of J.
+  solver.get_J()  # see documentation of the `newtonpf` function for more information about the shape of J.
 
-  # since lightsim2grid 0.14, the Jacobian rows / columns no longer follow the
+  # since lightsim2grid 1.0.0, the Jacobian rows / columns no longer follow the
```
BDonnot, line 107: We can also mention here which rows of the Jacobian holds which state variables (p or q of which buses) and mentioned that these are given with the solver labelling. And what could be good would be also to show how to get the corresponding Gridmodel / global bus labelling (with the proper lsgrid.ac_solver_to_me() call)

docs/use_solver.rst
BDonnot: I wonder if this still true now. I think most of the above is now checked. And this whole paragraph should be rephrased more like "lightsim2grid expect this that and this, otherwise it might raise an error" or something

docs/use_solver.rst
```diff
@@ -139,27 +148,32 @@ python virtual machine at execution time. So we encourage you to check that:
...
-distributed slack bus (but can be a bit slower) as the other one does not allow for such (in case of multiple slack bus, only 
+There are 8 solvers in this categorie. They can in turn, be split into two main sub categories. The first one allows for a
+distributed slack bus (but can be a bit slower) as the other one does not allow for such (in case of multiple slack bus, only
 the first one is used as a real slack bus, the other ones are converted silently to PV buses)
 
 The list is:
 
-- `NR_KLU` \*: implementation of the Newton Raphson algorithm supporting the distributed slack bus, where the 
+- `NR_KLU` \*: implementation of the Newton Raphson algorithm supporting the distributed slack bus, where the
   fast `KLU` implementation is used to iteratively update the jacobian matrix `J`.
```
BDonnot: It's not to update the Jacobian matrix it's to solve the linear system J.dx=mismatch, same for all items of this list

docs/use_solver.rst
```diff
@@ -139,27 +148,32 @@ python virtual machine at execution time. So we encourage you to check that:
...
-  all elements of `ref` into `pv` except the first one], where the 
-  Eigen default implementation is used to iteratively update the jacobian matrix `J` (instead of the faster `KLU` or `NICSLU`)
+- `NRSing_CKTSO` \*: implementation of the Newton Raphson algorithm only supporting single slack bus [ignores `slack_weight`, assign
+  all elements of `ref` into `pv` except the first one], where the
+  fast `CKTSO` implementation is used to iteratively update the jacobian matrix `J`.
+- `NRSing_SparseLU`: implementation of the Newton Raphson algorithm only supporting single slack bus [ignores `slack_weight`, assign
+  all elements of `ref` into `pv` except the first one], where the
+  Eigen default implementation is used to iteratively update the jacobian matrix `J` (instead of the faster `KLU`, `NICSLU` or `CKTSO`)
 
 You can use them as:
```
BDonnot, line 222: Before this you might check the actual behavior of the non distributed slack algorithm. I think each generator tagged as slack on the grid is considered as a slack and thus has theta = 0 and vm fixed. But I'm not sure. 
It's worth mentioning the actual behavior here so as to avoid any issues.

docs/use_solver.rst
```diff
@@ -217,6 +232,36 @@ There are 3 solvers of this type that are different in the way they solve `Ybus
...
   # process the results as above
 
+Fast Decoupled solvers
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+There are 8 solvers of this type. The "Fast Decoupled Power Flow" (FDPF) method exploits the (approximate)
+decoupling between active power / voltage angle on one side, and reactive power / voltage magnitude on the
+other, to avoid rebuilding the full Newton-Raphson jacobian at each iteration. Two variants are available,
+`XB` and `BX`, that differ in which simplifying assumption is made on the B' / B'' matrices; neither supports
+distributed slack bus [ignores `slack_weight`, assign all elements of `ref` into `pv` except the first one].
```
BDonnot, line 297: We might also need to check this statement 

And in any case advise people to remove all generators from the slack except one (if possible on a big voltage level, well connected and rather "central" to the powerflow graph).
This warning should also be made for single slack algorithm

docs/use_solver.rst
```diff
@@ -217,6 +232,36 @@ There are 3 solvers of this type that are different in the way they solve `Ybus
...
+
+  from lightsim2grid.algorithm import FDPF_XB_SparseLU  # or any of the names above
+
+  # retrieve some Ybus, V0, etc. as above
+  solver = FDPF_XB_SparseLU()
+  converged = solver.solve(Ybus, V0, Sbus, ref, slack_weights, pv, pq, max_it, tol)
+  # process the results as above
+
+.. note::
+  \* these 6 solvers might not be available on all platforms (`KLU` is available if installed from pypi, but not
```
BDonnot: It's 8 right ?

lightsim2grid/contingencyAnalysis.py
```diff
@@ -93,13 +93,13 @@ class __ContingencyAnalysis(object):
...
         import grid2op
-        from lightsim2grid import SecurityAnalysis
+        from lightsim2grid import ContingencyAnalysis
         from lightsim2grid import LightSimBackend
         env_name = ...
         env = grid2op.make(env_name, backend=LightSimBackend())
 
         0) you create
-        security_analysis = SecurityAnalysis(env)
+        security_analysis = ContingencyAnalysis(env)
```
BDonnot: Renmz the security_analysis variable

lightsim2grid/contingencyAnalysis.py
```diff
@@ -93,13 +93,13 @@ class __ContingencyAnalysis(object):
...
         
         1) you add some contingencies to simulate
         security_analysis.add_multiple_contingencies(...) # or security_analysis.add_single_contingency(...)
```
BDonnot: Rename security_analysis here too

lightsim2grid/contingencyAnalysis.py
```diff
@@ -304,20 +319,20 @@ def add_single_contingency(self, *args, name=None):
...
 
             import grid2op
-            from lightsim2grid import SecurityAnalysis
+            from lightsim2grid import ContingencyAnalysis
             from lightsim2grid import LightSimBackend
             env_name = ...
             env = grid2op.make(env_name, backend=LightSimBackend())
 
-            security_anlysis = SecurityAnalysis(env)
+            security_analysis = ContingencyAnalysis(env)
```
BDonnot: Rename security_analysis variable everywhere (I will lot tag them all)

pyproject.toml
```diff
@@ -53,8 +56,8 @@ Repository = "https://github.com/Grid2op/lightsim2grid"
 cmake.source-dir = "."
 # La destination d'installation doit correspondre au nom du package importable
 wheel.packages = ["lightsim2grid"] 
-# Pour un build verbose lors du debug
-build.verbose = true
+# mettre à true temporairement pour un build verbose lors du debug
```
BDonnot: Why this "franglish" here

BDonnot (commented):
Some issues to fix

PR #160, opened 2026-08-01
Documentation audit: centralize and complete C++ core / bindings docstrings

BDonnot:
## Summary

Full audit of `src/core/help_fun_msg.hpp`/`.cpp` (the shared docstring library) and `src/bindings/python/*.cpp` (the pybind11 bindings that consume it), ahead of the 1.0.0 release. Goal, per the request: (1) every public, actually-used method should have real documentation instead of `"TODO"` / "internal, do not use", (2) documentation should live centrally in `help_fun_msg.hpp/.cpp` rather than being duplicated/hand-written per binding, (3) flag/fix documentation whose content no longer matches the current code.

- **Section 2 (centralization)**: moved `TimerJac`/`LinearSolverStats`/NR-policy docs, `AlgoConfig`, and the `ContingencyAnalysis` `LimitViolation` family out of inline literals into `help_fun_msg`.
- **Section 3 (18 confirmed accuracy bugs)**: docstrings that contradicted the current code — stale module names, wrong solver-family cross-references, an outdated DC distributed-slack claim, a sign-convention contradiction in the HVDC docs, wrong min/max p/q usage descriptions, and more.
- **Section 4 (missing documentation)**: `PandaPowerConverter`, `AlgoControl`, admittance fields, `get_bus_id`, and the `SubstationContainer`/`SubstationInfo` classes (which had zero Sphinx coverage at all — added the missing `autoclass` directives).
- **19 remaining `LSGrid` `"TODO"` docstrings** found while sweeping the bindings during section 4 (`copy`, `timer_last_ac_pf`/`_dc_pf`, slack-weight management, half-open HVDC handling, the `set_*_names` family, etc).
- **SVC family** (`SvcContainer`/`SvcInfo`): had zero centralized docs at all. Added the full family plus the bordered-formulation voltage-regulation/droop equations (rendered as MathJax) on request.
- **Dangling-reference cleanup**: a full (untruncated) Sphinx warning diff — rather than the first-token-only diff used earlier — surfaced several numpydoc colon-parsing misfires and one real `Raises`-section bug (`LSGrid.check_grid`) that an earlier, coarser verification pass had missed.
- **146 of 155 `LSGrid` `_internal_do_not_use`-tagged methods**: traced each to its actual C++ implementation and gave it a real docstring (per-element bus getters/setters, activation status, setpoint setters, slack designation, bulk vector getters/setters, structural setup, container constructors). The remaining 9 (bus-level primitives with no per-bus Python class, plus pure solver-debug/bookkeeping calls) keep the internal marker — itself rewritten to be accurate rather than claiming a single caller that wasn't true for all its uses.
- **Reciprocal container ↔ `LSGrid` cross-references**: every container attribute (`LoadInfo.bus_id`, `GenInfo.target_p_mw`, etc.) that now has a corresponding `LSGrid` getter/setter documented above links back to it, and vice versa; read-only result fields (`res_p_mw` and friends) note there's no setter. Also clarifies throughout that `bus_id` fields are the gridmodel/global id, not the solver's internal numbering.

## Verification

Every pass was built, tested, and verified before committing: a real `pip install --no-build-isolation --no-deps -e .` build, a real-import round-trip on every touched attribute (non-empty, non-`"TODO"` docstrings), and a nitpicky Sphinx rebuild (`sphinx-build -b html -n`) diffing the full (untruncated) set of unresolved-reference targets before/after each change — confirming zero new dangling references were introduced by ~350 new/edited docstrings and cross-references.

## Test plan

- [x] Full C++ extension build after every change
- [x] Real-import round-trip of every touched docstring (non-empty, non-`"TODO"`, no leftover `"internal, do not use"` where real docs were added)
- [x] Nitpicky Sphinx rebuild, full unresolved-reference target diff before/after each pass — no new dangling references


BDonnot (commented):
Ok

PR #161, opened 2026-08-03
Close four more memory-safety holes on the untrusted-input / public-API paths

BDonnot:
Follow-up security audit of the `n1_full_compute` branch, in the spirit of #155 / #156 / #157, over the channels `docs/security.rst` covers (python bindings, pickle, the binary format) **plus** the new `ContingencyAnalysis` (n-1) code path that the earlier audits predate.

Four issues, each reproduced as a **segfault / heap corruption before** the fix and as a **clean exception (or safe behaviour) after**. Release wheels are `-O3 -DNDEBUG`, so neither Eigen's nor the standard library's own bounds checks catch any of these.

## Findings

| # | Issue | Trigger | Before the fix |
|---|-------|---------|----------------|
| 1 | `ContingencyAnalysis` stale results → out-of-bounds **write** | `ca.add_n1(x)` after `compute()`, then `compute_flows()` | `free(): invalid pointer`, heap corruption |
| 2 | `OneSideContainer::update_topo` → out-of-bounds **read** | `set_load_pos_topo_vect([2**28, …])` then `update_topo()` | SIGSEGV |
| 3 | `check_grid` accepted vmin/vmax with mismatched presence → OOB **read** | pickle / binary with `bus_vmin_kv` set, `bus_vmax_kv` empty | SIGSEGV (state passed `check_grid()`) |
| 4 | LSGrid sub-object getters → **use-after-free** | `make_grid().get_lines()` then drop the grid | disclosed freed heap (a 53M-element "bus id" array), SIGSEGV |

### Details

1. **`ContingencyAnalysis` stale results → OOB write.** `compute()` sizes the result matrices (`_voltages` and the flow matrices derived from it) with one row per registered contingency. Adding / removing contingencies afterwards, then calling `compute_flows()` / `compute_power_flows()` without recomputing, left `clean_flows()` (and the per-contingency violation loop) walking the new, longer `_li_defaults` while indexing the old, shorter matrices. `compute_flows()` / `compute_power_flows()` now refuse the stale state with a clear message.

2. **`OneSideContainer::update_topo` → OOB read.** `has_changed` / `new_values` are indexed by the stored `pos_topo_vect` with an unchecked Eigen `operator()`. `check_grid()` proves those positions form a permutation of `[0, dim_topo)`, but the `set_*_pos_topo_vect()` setters only length-check their argument, not its values — so a position written straight through a setter (bypassing `check_grid`) read past the caller arrays. Validate `el_pos` against `has_changed.rows()` before indexing, same idiom as the entry-path checks in #143.

3. **vmin/vmax mismatched presence → OOB read.** `bus_vmin_kv_` / `bus_vmax_kv_` are each individually optional (empty-or-complete), but `ContingencyAnalysis::check_bus_voltage_violations` consumes them together: it loops to `vmin.size()` and then reads `vmax(grid_id)`. A state with one set and the other empty passed every per-field check yet over-read the empty one (reachable from a pickle / a binary file). `check_grid` now requires them to have the same presence.

4. **LSGrid getters → use-after-free.** `get_lines` / `get_generators` / `get_substations` / … and the Eigen-view getters `get_bus_vn_kv` / `get_V_solver` / … were bound with `return_value_policy::reference`, which does **not** keep the parent grid alive. Dropping the grid while holding the returned container / numpy view left it dangling. Switch the getters that return a reference / view into grid-owned memory to `reference_internal` so their lifetime is tied to the grid (same fix for `AlgorithmSelector::get_fdpf_{xb,bx}_lu`).

## Docs

Adds a subsection to `docs/security.rst` stating the invariant these fixes uphold explicitly: an `LSGrid` must **raise, never corrupt memory** — whatever it is handed (malformed pickle/binary, a grid from a source file, or an ordinary API call in the wrong order or with an out-of-range id), the worst it may do is raise a clean Python exception. A genuine crash instead of an exception is a bug to report.

## Tests

- 13 new python cases in `test_state_poisoning.py` (four new test classes, one per finding).
- One new `test_check_grid.cpp` case for the vmin/vmax presence check.
- Full C++ suite (**128** tests) green; the python poisoning suite passes.

## Not covered / blind spots

This audit concentrated on the untrusted-input and public-API boundary. It did **not** deeply review the numeric solver internals (`powerflow_algorithm/`), the external sparse-solver wrappers (`linear_solvers/`), or each concrete element container's `fillYbus` / `compute_results` indexing. Running the repo's ASan/UBSan CI over the full suite is the recommended next step to reach those.

https://claude.ai/code/session_01Vxxj4GukTFz1TKh52GZxLP

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vxxj4GukTFz1TKh52GZxLP)_


src/core/batch_algorithm/ContingencyAnalysis.hpp
```diff
@@ -240,6 +242,26 @@ class LS2G_API ContingencyAnalysis final: public BaseBatchSolverSynch
...
+        // raises if the registered contingency set no longer matches the last
+        // computed results. `compute()` sizes `_voltages` (and therefore the flow
+        // matrices derived from it) with one row per contingency in `_li_defaults`
+        // at the time it ran. Adding / removing contingencies afterwards without
+        // recomputing leaves `_voltages` shorter than `_li_defaults`, and both
+        // `clean_flows()` and the per-contingency violation loop walk `_li_defaults`
+        // while indexing the (shorter) result matrices -- an out-of-bounds write
+        // (heap corruption) on release wheels (-O3 -DNDEBUG, no bounds checks). Fail
+        // loudly instead, before any result matrix is indexed.
+        void _check_results_match_defaults(const std::string & fun_name) const {
```
BDonnot, line 271: For this bug, this also means that removing a contingency after compute will be wrong (though no OOB read). This fox is good but not enough. The better fix is that any function modifying the contingency list (remove or add) should clear the results of compute (and a new call compute should be made)

src/core/element_container/OneSideContainer.hpp
BDonnot, line 279: Given the findings in "update_topo" I would also enforce here that the values of pos_topo_vect are checked.
And this for all "set_pos_topo_vect" methods (in Twosidedcontainer too if needed)

BDonnot, line 298: Same comments as for set_pos_topo_vect : the sub_id values here should (if possible) be checked and not just it's length. I suppose that right now this is set by LightsimBackend so after the check_grid has been run on the full grid. So in practice no enforced.
Maybe the real fix would be to set the sub_id here in the init_* functions directly and not in LightsimBackend?

BDonnot (commented):
Some changes needed

src/core/batch_algorithm/ContingencyAnalysis.hpp
BDonnot: Good catch — confirmed and fixed in a5b8a5b. A size-only check can't tell `remove_n1(x)` + `add_n1(y)` (y != x) apart from a clean state, so I reproduced it: same array shape, silently different (wrong) values, no exception.

Went with your suggested fix: every mutating method (`add_all_n1` / `add_n1` / `add_multiple_n1` / `add_nk` / `remove_n1` / `remove_multiple_n1` / `remove_nk`) now calls `clear_results_only()` eagerly. Validation still runs to completion *before* any mutation/clearing, so a rejected id leaves both the contingency set and any existing results untouched — the batch methods now validate every id first, then mutate, rather than interleaving the two (previously a batch call could partially mutate `_li_defaults` before throwing on a later invalid id).

Kept the size-based check in `compute_flows()`/`compute_power_flows()` too, as a cheap backstop.

---

src/core/element_container/OneSideContainer.hpp
BDonnot: Fixed in a5b8a5b. `set_pos_topo_vect()` now rejects a negative position immediately — that's the only bound it can check locally: the upper bound (`dim_topo`) is the sum of every topology-participating container's element count, not just this one's, so it isn't known here (only `update_topo()`, via the `has_changed`/`new_values` arrays it's given, has that context — which is where the previous commit's range check lives).

Applied to `set_pos_topo_vect` and (comment below) `set_subid`. Checked `TwoSidesContainer::set_pos_topo_vect_side_1/2` — they delegate straight through to `side_1_.set_pos_topo_vect(...)` / `side_2_.set_pos_topo_vect(...)`, i.e. this exact method, so lines/trafos inherit the fix with no separate change needed there.

---

BDonnot: Fixed in a5b8a5b, and this one turned out to be a live bug, not just hardening. Traced the only consumer of `subid_`: `update_topo()` feeds it to `SubstationContainer::local_to_gridmodel(sub_id, busbar)`, which computes `sub_id + (busbar - 1) * n_sub` and bound-checks the *output* (the resolved bus id) before it's stored — but never checks `sub_id` itself. An out-of-range `sub_id` can land by coincidence on another substation's legitimate bus id and pass that check, so the element gets silently reconnected to the WRONG bus instead of raising. Reproduced it: a load with a poisoned `subid` got aliased onto an unrelated substation's bus, no exception.

`set_subid()` now rejects a negative value immediately (same reasoning as `set_pos_topo_vect` above — `n_sub` is owned by `SubstationContainer`, not available here). `update_topo()` validates the full `[0, n_sub)` range right before calling `local_to_gridmodel`, since it already has `substations` in scope. Also added a guard for `subid_` never having been set at all (empty vector) — that was an unchecked read on a size-0 Eigen vector, same class of bug.

`TwoSidesContainer::set_subid_side_1/2` delegate to this method too, so lines/trafos are covered.

On your architectural suggestion (setting `sub_id` in the `init_*` functions instead of via `LightsimBackend`): agreed that's the more thorough fix — it would let validation happen with full context in one place instead of deferring to `update_topo()`/`check_grid()`. That's a bigger change to the loading pipeline and the Python/C++ contract though, so I left it out of this defensive-programming PR; happy to open a follow-up if you want to pursue it.

---

BDonnot (commented):
Ok

PR #162, opened 2026-08-03
Simplify and fix benchmarks/ so the docs stay in sync with what's measured

BDonnot:
## Summary

The docs pages built from `benchmarks/` (`docs/benchmarks.rst`, `docs/benchmarks_dc.rst`,
`docs/benchmarks_grid_sizes.rst`, `docs/comparison_with_pypowsybl.rst`,
`docs/security_analysis.rst`, `docs/time_series.rst`, `docs/binary_serialization.rst`) each carry
prose commenting on the tables above them (speed ups, comparisons between solvers/backends,
percentages). That prose was hand-derived and hand-typed after every benchmark run, and had
drifted out of sync with its own tables in several places — sometimes badly (`docs/benchmarks_dc.rst`'s
TL;DR no longer matched its tables at all, and referenced a pypowsybl row that table doesn't even have).

This PR makes the benchmark scripts generate that text themselves, from the numbers they just
measured, so updating the docs after a run on your own machine (still a manual step, as requested)
is a copy/paste instead of manual float formatting and re-derivation:

- **`benchmarks/benchmark_solvers.py`** (and its DC counterpart `benchmark_dc_solvers.py`) now print
  a `generate_narrative()`-produced paragraph after their tables, covering speed-ups vs
  pandapower/pypowsybl, linear-solver comparisons, the "time external to the solver" breakdown, and
  max differences vs pandapower — computed from the same run, not retyped. Along the way this fixed a
  real unit bug in the old hand-written text (values labelled "ns" that were actually µs).
- **`benchmarks/compare_lightsim2grid_pypowsybl.py`** can now run a batch of cases (default: all 6
  IEEE cases) in one process and print the 5 tables + narrative used in
  `docs/comparison_with_pypowsybl.rst` directly, instead of requiring one invocation per case and
  manually placing each number in the right table cell. `benchmark_pypowysbl.sh` updated to match.
- **`benchmarks/security_analysis.py` → `benchmarks/contingency_analysis.py`** (renamed — its
  `examples/` counterpart was already renamed away from the old `SecurityAnalysis` name in a previous
  pass, this one was missed) and **`time_serie.py`** now print their closing "N times faster than raw
  grid2op" sentence themselves, computed from the timings printed just above it, instead of leaving it
  to drift as a separate hand-rounded restatement.
- **`benchmarks/benchmark_grid_size.py`** gained the same treatment (recycling vs no-recycling,
  `TimeSerie`/`ContingencyAnalysis` vs a plain grid2op step, as computed ranges across the grid-size
  scan) and now fills in `docs/benchmarks_grid_sizes.rst`'s previously-empty "Comments" section.
- **`benchmarks/benchmark_binary_serialization.py`**'s single-grid table gained a "speedup vs pickle"
  column (mirroring the grid-size-scan table below it), replacing a hand-written sentence that
  independently restated the same ratio.
- **Orchestration fixes**: `run_all_benchmarks.sh` referenced a nonexistent `./benchmarks_grid_size.sh`
  (real file is `benchmark_grid_size.sh`, singular) and never ran `benchmark_binary_serialization.py` at
  all; `benchmark_ts_ca.sh` never ran `benchmark_ca_nb_threads.py` (the script `docs/security_analysis.rst`'s
  nb_thread-scaling section is built from) and called the old `security_analysis.py` name. All fixed.

Several doc pages were also corrected in place using the real numbers already published in their own
tables (no benchmark re-run needed for these): `docs/benchmarks_dc.rst`'s TL;DR and previously-`TODO`
"Comments" section, `docs/comparison_with_pypowsybl.rst`'s "between 30 and 5x" claim (real range is
~7-37x), and the stale "22x"/"15x" summary sentences in `docs/security_analysis.rst` /
`docs/time_series.rst`.

## Test plan

- [x] `python3 -m py_compile` on every touched script in `benchmarks/`
- [x] `bash -n` on every touched shell script
- [x] Validated the new `generate_narrative()` / report-generation functions against both synthetic
      data and the real numbers already published in the docs (stubbing out grid2op/pypowsybl/
      lightsim2grid, none of which are installed in this environment)
- [x] Verified every new/edited RST table parses cleanly with `docutils` (no "text in column margin" /
      malformed-table errors; only pre-existing, unrelated Sphinx-role warnings remain)
- [ ] Actually running the benchmarks on real hardware and re-pasting fresh numbers into the docs is
      still a manual step, as requested — not part of this PR

https://claude.ai/code/session_015bgqPwrM2fV1XpFDtgu16S

---
_Generated by [Claude Code](https://claude.ai/code/session_015bgqPwrM2fV1XpFDtgu16S)_


BDonnot (commented):
Ok

PR #164, opened 2026-08-04
ci: fix publish.yml version guard to read lightsim2grid/__init__.py

BDonnot:
## Summary
- `publish.yml`'s GUARD 1 ("version in setup.py must match the release tag") still read `setup.py`, which no longer exists on `dev_0.13.2` — the project moved to `pyproject.toml` + scikit-build-core, with `__version__` now living in `lightsim2grid/__init__.py`. That file is already the declared source of truth in `pyproject.toml`'s `[tool.scikit-build.metadata.version]` regex.
- Left unfixed, the check would crash with `FileNotFoundError: setup.py` the moment any release is published, regardless of tag — this was caught while checking whether tagging `v1.0.0.rc2` would successfully trigger the PyPI publish flow.
- Points the check at `lightsim2grid/__init__.py` instead, using the same regex `pyproject.toml` already relies on, and renames the shell variable (`SETUP_VERSION` → `PKG_VERSION`) and log lines to match.

## Test plan
- [ ] Regex verified locally against the current `lightsim2grid/__init__.py` (`__version__ = "1.0.0.rc2"` → correctly extracts `1.0.0.rc2`, expects tag `v1.0.0.rc2`)
- [ ] Exercise the actual publish workflow via a real tag + published release (can't be verified from a PR alone, since `publish.yml` only fires on `release: published`)

---

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -24,36 +24,40 @@ jobs:
       actions: read     # required to download artifacts from main.yml
 
     steps:
-      # Checkout is needed to read setup.py for the version consistency check.
+      # Checkout is needed to read lightsim2grid/__init__.py for the version
+      # consistency check.
       - name: Checkout sources
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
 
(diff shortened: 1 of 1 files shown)


BDonnot (commented):
Ok

PR #163, opened 2026-08-04
ci: pin all GitHub Actions to full commit SHAs

BDonnot:
## Summary
- Pin every `uses:` action reference across all workflow files (`main.yml`, `publish.yml`, `clang-tidy.yml`, `cpp-standards.yml`, `cpp_unit_tests.yml`, `sanitizers.yml`) to a full commit SHA with a trailing version comment, e.g. `actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1`, instead of a floating major-version tag (`@v4`) — hardens CI against supply-chain attacks via tag mutation.
- Upgraded every pinned action to its current latest release (`actions/checkout` v4→v7.0.1, `actions/setup-python` v5→v7.0.0, `actions/upload-artifact` v4→v7.0.1, `actions/download-artifact` v4→v8.0.1, `docker/setup-qemu-action` v3.7.0→v4.2.0, `pypa/cibuildwheel` v3.3.0→v4.1.1, `lewagon/wait-on-check-action`→v1.9.0, `dawidd6/action-download-artifact`→v21, `pypa/gh-action-pypi-publish`→v1.14.2, `bus1/cabuild/action/msdevshell`→latest `v1` branch commit, no tagged releases upstream).
- Fixed a broken action reference already on this branch: `pypa/gh-action-pypi-publish@release/cef221092ed1bacb1cc03d23a2d87d1d172e277b` was not a valid git ref (invalid branch name) and would have failed the publish job; replaced with the correct pinned SHA for v1.14.2.

## Notes / things to watch when testing CI
- `pypa/cibuildwheel` jumped a major version (v3.3.0 → v4.1.1) — worth watching the `exotic_build` job on the first CI run for behavior changes.
- `pypa/gh-action-pypi-publish` is normally left on the floating `@release/v1` branch per PyPA's own guidance (it's an intentionally-tracked branch tied to their OIDC trusted-publishing security model), but was SHA-pinned here for consistency with the rest of this PR. Happy to revert that one specifically if preferred.

## Test plan
- [ ] YAML validated locally (`yaml.safe_load` on all workflow files — passes)
- [ ] Trigger CI on this branch/PR and confirm all jobs (manylinux/windows/macos/exotic builds, cmake standalone, plugin, KLU strategies, march-native, sanitizers, clang-tidy, cpp-standards) still pass with the upgraded action versions
- [ ] Verify the release-candidate publish flow once tested (per branch owner)

---
[...]

PR #165, opened 2026-08-04
fix: copy lightsim2grid_core.dll next to the legacy root .pyd on Windows

BDonnot:
## Summary
- `INSTALL_LEGACY_LIGHTSIM2GRID_CPP` (ON by default) installs a second copy of `lightsim2grid_cpp` at the site-packages root for deprecated top-level `import lightsim2grid_cpp`. On APPLE/UNIX, that copy is made to find `lightsim2grid_core` via an added rpath entry pointing back into `lightsim2grid/` (`install_name_tool`/`patchelf`). Windows has no rpath equivalent — the loader only searches the directory of the loading module, a few OS paths, and `PATH` — so the root-level `lightsim2grid_cpp.pyd` could never resolve `lightsim2grid_core.dll`, which lives one directory away in `lightsim2grid/`.
- This was silently broken at runtime already (top-level `import lightsim2grid_cpp` would fail with a DLL-not-found error on Windows), and turned into a hard CI failure once `delvewheel`'s Windows wheel-repair step got stricter about resolving PE import-table dependencies — surfaced by the `pypa/cibuildwheel` version bump in #163 (`delvewheel` couldn't find `lightsim2grid_core.dll` while analyzing the root-level `.pyd` and aborted the repair with `FileNotFoundError`, breaking the `exotic_build` Windows ARM64 job).
- Fix: on Windows, copy `lightsim2grid_core.dll` alongside the root-level `.pyd` (`install(FILES "$<TARGET_FILE:lightsim2grid_core>" DESTINATION .)`), since same-directory lookup is the one DLL search rule Windows always honors.

## Trade-off
This adds a second physical copy of `lightsim2grid_core.dll` to Windows wheels (macOS/Linux don't pay this cost — they redirect via rpath instead of duplicating). If that turns out to matter for wheel size, dropping the legacy install for Windows entirely is a straightforward follow-up, since it wasn't functional there before this fix anyway.

## Test plan
- [ ] Verified the `install(FILES "$<TARGET_FILE:...>")` mechanism works as intended with an isolated CMake project locally (Linux `.so`, confirms the generator-expression + copy semantics; the actual `.dll` path can only be exercised on the Windows CI runner)
- [ ] Confirm the `exotic_build` Windows ARM64 `cibuildwheel` job (currently failing at `delvewheel repair`) goes green with this change

---

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -610,6 +610,18 @@ jobs:
           CIBW_ARCHS_MACOS: arm64 
           CIBW_ARCHS_WINDOWS: ARM64 
           CIBW_ENVIRONMENT: __O3_OPTIM=1
+          # lightsim2grid_cpp's legacy root-level copy (see
+          # INSTALL_LEGACY_LIGHTSIM2GRID_CPP in CMakeLists.txt) depends on
+          # lightsim2grid_core.dll, which lives one directory away in
+          # lightsim2grid/. delvewheel's default repair command only
+          # resolves DLL dependencies via directories already inside the
+          # wheel when --ignore-existing is passed, or PATH -- it never
+          # searches the same directory as the module being analyzed. Add
+          # --ignore-existing (cibuildwheel's default command otherwise,
+          # see delvewheel's own --ignore-existing/--ignore-in-wheel help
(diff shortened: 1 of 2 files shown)


BDonnot (commented):
Ok

PR #174, opened 2026-08-08
Harden the Newton-Raphson hot path against OOB in its extensions, and cache their solver data

BDonnot:
## What this is

An audit of `src/core/powerflow_algorithm/` for out-of-bounds / use-after-free in the extensions (HVDC angle droop, remote voltage control / SVC, distributed slack), plus the fixes and tests that came out of it.

The design went through several rounds of review — the early commits put checks in places that broke the caching contract or did unnecessary per-solve work. **Read the last three commits for the design that actually landed**; the first three are the route there and are kept for the history.

## The gap

`BaseAlgo::check_pf_inputs` already covers everything that reaches a solver through `compute_pf`'s arguments. The extensions don't get their data that way: they pull it from the grid through the raw `lsgrid_ptr_` inside `update_state`, so none of it was range-checked. Every index in it is then used raw — `ledger.p_row(bus)` is a `std::vector::operator[]`, `Va(bus)` an Eigen `operator()` — and release wheels are `-O3 -DNDEBUG`.

Underneath that sat a structural hazard: `update_state` re-pulls that data on **every** solve, while `register_in` re-derives the row/column arrays indexing it only on a **topology rebuild**. If an element count ever changed without one being signalled, the two drift apart. Confirmed with ASan rather than argued:

```
ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 4 ... in ls2g::VoltageControl::fill_feature_values
0x502000003f94 is located 0 bytes after 4-byte region
```

That garbage handle then goes to `FeatureWriter::add` → `J_values_[positions_[h]] +=`, an out-of-bounds **write** into the Jacobian.

No public API reaches that state today — the element containers are careful. The point is that the invariant was upheld by convention across ~20 files rather than stated anywhere.

## What landed

**Validation at the producer, not the consumer.** `fill_voltage_control_solver_data` now checks each controller's solver ids where they are first used as indices (`is_pq[ctrl_solver]` reads an `nb_bus_solver`-sized vector a few lines later, so an out-of-range id was an OOB read inside that function), plus two postconditions of its own construction: the groups tile the controller list exactly, and no group is empty — `VoltageControl` sizes its sharing vectors from `grp_count - 1` as a `std::size_t`, which underflows for an empty group.

**A dedicated `AlgoControl` flag** — `tell_hvdc_droop_changed()` / `has_hvdc_droop_changed()` — for the set `{droop_enabled AND connected}`. Raised in `disable_droop` (which previously took no solver control at all and mutated `droop_enabled_` silently) and in HVDC connect/disconnect for droop-enabled lines. Deliberately **not** raised for `status_droop`: that is a regime value the extension absorbs without touching the J sparsity pattern, which is what lets KLU reuse its symbolic factorization across saturation flips.

**Caching of the HVDC / voltage-control solver data.** Both extensions rebuilt it from the grid on every solve. The grid is `const` during a solve, so it can only change between solves and `AlgoControl` already records that. Also: `fill_voltage_control_solver_data` did its expensive setup *before* discovering there was nothing to do — two `nb_bus_solver`-sized `vector<bool>` and a `std::set` built per powerflow to produce an empty result on any grid without controllers. A cheap pre-scan now returns first.

**On the per-solve path there is only** `NRSystem::validate_registration()`: size comparisons between quantities the components already hold. No allocation, no per-element sweep, nothing proportional to bus count. `NRLedger`'s `assert(n_rows_ == n_cols_)` also became a real check (`-DNDEBUG` removed it, and a non-square ledger means J is resized to `(n_rows, n_rows)` then written past it).

## Cost

Benchmarked with `unset_changes()` between solves, so the symbolic factorization is reused — the path that matters. Against a build with the call removed:

| grid | median with | median without |
|---|---|---|
| case1354pegase | 13.98 ms | 14.06 ms |
| case2869pegase | 33.45 ms | 33.73 ms |
| case9241pegase | 179.88 ms | 179.63 ms |

No difference. (`case1354pegase` / `case9241pegase` carry 0 droop HVDC, 0 SVC, 0 remote-regulating generators, so every extension loop is empty there.)

## Tests

New `src/tests/test_powerflow_algorithm.cpp` (18 cases): each extension alone and all engaged in one Jacobian; the layout invariants; `status_droop` flips preserving the sparsity pattern for KLU symbolic reuse; a droop end on a slack bus; a sloped SVC; a two-generator control group (the `cnt - 1` sharing path); bus masking; `clear_jacobian` reuse; a null grid pointer; a 40-step mutate-and-resolve loop; and direct drives of the 3-phase protocol.

Two CI jobs added running the C++ suite under ASan+UBSan and under re-enabled Eigen/libstdc++ assertions — it previously ran only under valgrind, and the sanitizer jobs covered only Python.

Verified: 150 C++ cases across plain / ASan+UBSan / Eigen assertions; 308 Python tests across the HVDC, voltage-control, solver-control and batch suites; all three example plugins build.

## Worth flagging

- **A commit in this branch broke CI and is fixed by the last one.** `6c28a37` added an HVDC bus-id check whose bound came from `id_ac_solver_to_me_`, the *reverse* mapping — and only the forward one is synced back onto the grid when a batch algorithm drives the solve, so it read empty and 14 tests failed. `6963b1c` removes that check rather than papering over it: the bound that matters for those ids is the solver's own bus count, which `fill_hvdc_droop_solver_data` does not know and should not be taught, and nothing inside it indexes anything with them.
- **`compute_pf` vs `compute_pf_with_input_validation` is unaddressed.** Nothing prevents the raw virtual being called from outside LSGrid / the batch algorithms. A passkey parameter would enforce it at compile time with no runtime cost, at the price of changing every override signature including the documented plugin API. Not attempted here.
- **Remote voltage control appears inactive in the batch path.** `fill_voltage_control_solver_data` early-returns when `id_ac_solver_to_me_` is empty, which is the case when ContingencyAnalysis / TimeSeries drive the solve. Pre-existing, not touched, but noticed while chasing the bound bug above and probably worth a look.
- The extensions' `fill_*_solver_data` grouping pass is O(n_controllers × n_groups). The caching makes it run less often; it is still quadratic when it does run.

https://claude.ai/code/session_011aAdXxU5KzaZhfVQxZCiDo

---
_Generated by [Claude Code](https://claude.ai/code/session_011aAdXxU5KzaZhfVQxZCiDo)_

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -123,6 +123,52 @@ TODO: Levenberg-Marquardt damping (a.k.a. Tikhonov-regularized Newton) : adding 
 
 [1.0.0] 2026-xx-yy
 --------------------
+- [ADDED] memory-safety hardening of the Newton-Raphson hot path
+  (``src/core/powerflow_algorithm/``), which until now had no validation of its own: the previous
+  input-validation work covers what reaches a solver through ``compute_pf``'s arguments (``Ybus``,
+  ``Sbus``, ``V``, ``slack_ids`` / ``slack_weights`` / ``pv`` / ``pq`` -- see ``BaseAlgo::
+  check_pf_inputs`` and ``docs/security.rst``), but the *extensions* (``Hvdc`` angle droop,
+  ``VoltageControl`` for remote-regulating generators and SVCs, ``MultiSlack``) pull their per-solve
+  data from the grid inside ``NRSystem::update_state`` instead, so none of it was ever range-checked.
+  Every index in that data is then used raw -- ``ledger.p_row(bus)`` is a ``std::vector::operator[]``,
+  ``Va(bus)`` an Eigen ``operator()``, and release wheels are ``-O3 -DNDEBUG``. Two hooks were added to
+  the component protocol: ``validate_data(n_bus)``, run just before ``register_in`` (its first
+  consumer) and checking that the data is self-consistent and every bus id is in ``[0, n_bus)``; and
+  ``validate_registration(dim_J)``, run once per solve by ``NRAlgo::compute_pf``, checking that the row
+  / column arrays a component cached in ``register_in`` still match the data ``update_state`` just
+  pulled and all index a ``dim_J``-square Jacobian. The second is the load-bearing one: ``update_state``
+  runs on every solve while ``register_in`` only runs on a topology rebuild, so an element-count change
+  that failed to signal one (a missing ``tell_pv_changed()`` in any element container) made
+  ``VoltageControl`` / ``Hvdc`` read their handle arrays past the end and, through ``FeatureWriter``,
+  **write out of bounds into** ``J.valuePtr()``. No public API reaches that state today -- this makes
+  the invariant enforced rather than merely upheld. Also turned ``NRLedger``'s debug-only
+  ``assert(n_rows_ == n_cols_)`` into a real check (``-DNDEBUG`` removed it, and a non-square ledger
+  means J is resized to ``(n_rows, n_rows)`` and then written at a column beyond it), and guarded the
+  ``cnt - 1`` sharing-row arithmetic in ``VoltageControl`` against an empty control group (it is a
+  ``std::size_t``, so it underflowed). Like ``check_grid()``, none of this is compiled out by
+  ``-DNDEBUG``: it is what turns the inconsistency into an exception rather than heap corruption.
+  Nothing runs per Newton iteration, and nothing on the per-solve path scales with grid size: Base's
+  ``pv`` / ``pq`` range sweep (the only O(n_bus) part) runs *only* on a topology rebuild, since that is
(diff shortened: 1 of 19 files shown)


CHANGELOG.rst
```diff
@@ -123,6 +123,52 @@ TODO: Levenberg-Marquardt damping (a.k.a. Tikhonov-regularized Newton) : adding
 
 [1.0.0] 2026-xx-yy
 --------------------
+- [ADDED] memory-safety hardening of the Newton-Raphson hot path
+  (``src/core/powerflow_algorithm/``), which until now had no validation of its own: the previous
+  input-validation work covers what reaches a solver through ``compute_pf``'s arguments (``Ybus``,
```
BDonnot, line 128: This is not true you just missed the validation performed entirely

BDonnot (commented):
Will not merge

PR #176, opened 2026-08-08
Fix remote voltage control being silently dropped by the batch algorithms

BDonnot:
## The bug

Remote voltage control (remote-regulating generators, voltage-mode SVCs) **and** the free `Vm` unknown of a distributed-slack participant were silently inactive in every batch algorithm — `TimeSeriesCPP`, `ContingencyAnalysisCPP`, `SecurityAnalysis`, i.e. everything going through `BaseBatchSolverSynch`. The powerflow still converged and still returned plausible voltages; they were just the voltages of a grid with no such regulation. A silent wrong answer, not an error. `LSGrid::ac_pf` was unaffected.

On a 4-bus feeder with a generator remotely regulating the load bus at 1.04 pu:

| path | \|V(reg bus)\| |
|---|---|
| `LSGrid::ac_pf` | **1.040000** |
| `TimeSeries`, 1 step, same injections | 0.961689 |
| `ContingencyAnalysis`, empty contingency | 0.961689 |

## Root cause

The NR extensions do not receive their data as solver arguments. They reach back into the `LSGrid` through the algorithm's `lsgrid_ptr` and read its **member** solver-side labelling:

| extension | entry point | member read |
|---|---|---|
| `Base` | `get_free_vm_slack_solver_buses()` | `slack_bus_id_ac_solver_` |
| `Hvdc` | `fill_hvdc_droop_solver_data()` | `id_me_to_ac_solver_` |
| `VoltageControl` | `fill_voltage_control_solver_data()` | `id_ac_solver_to_me_`, + the two above |

`ac_pf` / `dc_pf` / `check_solution` pass those members straight to `pre_process_solver`, so they are always current there. The batch algorithms own local mapping vectors and pass those instead, against a private **copy** of the grid whose copy constructor `reset()`s all of it. `_pre_process_solver_impl` wrote only the **forward** map back onto the grid.

So `id_ac_solver_to_me_` and `slack_bus_id_ac_solver_` stayed empty — and empty is indistinguishable from *"this grid has no controller"*: `fill_voltage_control_solver_data` returned on its `nb_bus_solver == 0` guard and `get_free_vm_slack_solver_buses` on its `slack.empty()` one, both without a word. Confirmed by instrumentation:

```
batch:       id_ac_solver_to_me_.size()=0   id_me_to_ac_solver_.size()=4  slack_bus_id_ac_solver_.size()=0
single-shot: id_ac_solver_to_me_.size()=4   id_me_to_ac_solver_.size()=4  slack_bus_id_ac_solver_.size()=1
```

Note this is deterministic, not order-dependent: because the copy constructor resets the solver-side state, pre-solving the source grid does not help.

## The fix

`_pre_process_solver_impl` now syncs the reverse map and both slack vectors as well, AC and DC — completing an invariant the existing comment already claimed but implemented for one quarter of the state.

The `Hvdc` angle-droop extension was never affected: it only ever read the forward map, the one already synced. There is now a regression test pinning that too.

## Behaviour changes to be aware of

Both trade a silent wrong answer for an explicit one:

- A voltage-control configuration unsupported in v1 now **raises** in the batch path, as it always did single-shot. This is what broke `test_LSGrid_out_of_bounds.test_valid_calls_still_work`, fixed in its own commit: it used `set_gen_regulated_bus(0, 0)` as a stand-in for "a valid in-range id", but bus 0 of case14 is the slack, so the configuration is genuinely unsupported — `ac_pf` has always raised on it, and the `compute_Vs` call only passed because the batch discarded voltage control entirely. The test was asserting the bug. It now uses bus 9, a PQ load bus.
- Under `ContingencyAnalysis`'s `handle_disconnected_grid` mode, a contingency stranding a regulated bus is reported as a `DIVERGENCE` instead of converging with the regulation dropped. Bus masking is deliberately a value-only edit that must not touch the `J` pattern, so the bordered voltage row survives against a bus the mask pins and the system has no solution. I did **not** filter masked controllers out: that changes the system dimension and would break the symbolic-factorization reuse masking exists to provide. An honest non-result rather than a confident wrong one.

## Tests

- `src/tests/test_batch_voltage_control.cpp` — 7 cases: remote generator through `TimeSeries` and `ContingencyAnalysis`, a pre-solved source grid, two generators sharing one regulated bus, a voltage-mode SVC (flat and sloped), a free-`Vm` distributed-slack participant, an HVDC angle droop (the control), and a stranded-bus contingency. **6 of 7 fail without the fix**; the HVDC control passes both ways.
- `lightsim2grid/tests/test_voltage_control_batch.py` — case14, modelled on `test_hvdc_batch.py`. 8 assertion failures without the fix.

Each test solves the same grid single-shot and through the batch classes and requires the voltages to agree.

## Verification

- Full C++ suite green (4956 assertions / 156 cases) in Release, pinned to C++14, and under the `__DEBUG_ASSERTS=1` build CI uses.
- Full python suite run twice (1293 tests). All 133 remaining failures were attributed individually to this container's missing grid2op `data_test/`, `pypowsybl`, and KLU; the diff between the two runs is a single line — `test_valid_calls_still_work` disappearing.

Independent of #(the bus-reclassification PR): disjoint code, either merge order works.

---

### DIFF
--- a/lightsim2grid/tests/test_voltage_control_batch.py
+++ b/lightsim2grid/tests/test_voltage_control_batch.py
@@ -1,0 +1,204 @@
+# Copyright (c) 2026, RTE (https://www.rte-france.com)
+# See AUTHORS.txt
+# This Source Code Form is subject to the terms of the Mozilla Public License, version 2.0.
+# If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+# you can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
+# This file is part of LightSim2grid, LightSim2grid implements a c++ backend targeting the Grid2Op platform.
+
+"""The batch algorithms (``TimeSeriesCPP`` / ``ContingencyAnalysisCPP``) must apply
+remote voltage control -- a remote-regulating generator or a voltage-mode SVC --
+exactly as ``ac_pf`` does.
+
+They used not to. The ``VoltageControl`` NR extension is not built from the solver's
+arguments: it reaches back into the ``LSGrid`` through the algorithm's ``lsgrid_ptr``
+and reads that grid's member solver-side bus labelling. ``ac_pf`` passes its own
+members to ``pre_process_solver`` so they are always current, but the batch classes
+work on a private copy of the grid and pass their own local mapping vectors, and
+only the FORWARD map was written back. With the reverse map left empty,
+``fill_voltage_control_solver_data`` returned on its ``nb_bus_solver == 0`` guard --
+so the batch solved a grid with no regulation at all, converged, and returned
+plausible wrong voltages instead of raising.
+
+Companion C++ suite: ``src/tests/test_batch_voltage_control.cpp``.
+"""
+
+import warnings
+import unittest
+import numpy as np
+import pandapower.networks as pn
+
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore")
+    from lightsim2grid.gridmodel import init_from_pandapower
+from lightsim2grid.lightsim2grid_cpp import TimeSeriesCPP, ContingencyAnalysisCPP  # noqa: E402
+
+MAX_IT = 30
+TOL = 1e-11
+# SvcContainer.RegulationMode
+SVC_VOLTAGE_MODE = 1
+# gen 3 sits on bus 7; bus 9 is a PQ load bus no generator regulates locally.
+GEN_ID = 3
+REG_BUS = 9
+SVC_BUS = 4       # a PQ bus, so the SVC's own bus carries a reactive equation
+SVC_V_SET = 1.06
+
+
+def _case14():
+    with warnings.catch_warnings():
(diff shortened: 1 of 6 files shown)


lightsim2grid/tests/test_voltage_control_batch.py
```diff
@@ -0,0 +1,204 @@
...
+                                        lq_mat[step], nb_bus)
+                    self.assertGreater(Vref.shape[0], 0, f"{kind} ref step {step} diverged")
+                    self.assertLess(np.abs(Vs[step] - Vref).max(), 1e-7,
+                                    f"{kind} step {step}")
+
+
+class TestVoltageControlContingencyAnalysis(unittest.TestCase):
+    def test_contingency_matches_single_shot(self):
+        # powerlines away from the controller / regulated buses, so every
+        # contingency leaves the control problem itself well posed
```
BDonnot, line 149: How about a stress test disconnecting an element close to the controler / regulated bus ?

BDonnot (commented):
Ok

PR #177, opened 2026-08-08
Classify voltage-controlled buses by group before PV, and let stations join a group

BDonnot:
## The bug

A generator could not remotely regulate a bus that another generator already regulated **locally**. The grid was refused outright:

```
LSGrid::fill_voltage_control_solver_data: generator N regulates bus X
which has no voltage (Vm) unknown (it is a slack or an already-PV bus).
```

…even though the configuration is perfectly well posed. Two machines holding one bus simply form a control group: one bordered voltage row, one reactive unknown each, and a sharing row splitting the reactive duty by `qmax - qmin` — exactly what two *remote* controllers on one bus already did.

## Root cause: the order of bus classification

`GeneratorContainer::fillpv` implements only half the rule. It knows how to keep a controller's **own** bus out of PV (a remote regulator leaves its own bus PQ, `GeneratorContainer.cpp:279`), but nothing stopped a **local** regulator from claiming a bus somebody else regulated remotely.

That bus then had no `Vm` unknown, so `VoltageControl::register_in` got `vm_col == -1`, `declare_feature_entries` never declared `h_vm_[g]`, and for a GEN controller (no slope entry) the group's voltage row came out **structurally empty** — a singular Jacobian. Verified directly: relaxing the check *without* fixing the classification gives `ErrorType::SolverFactor`, not divergence. The check that refused the grid was reporting the misclassification, not a property of the configuration.

## The fix

Classification now runs the other way round:

- `LSGrid::get_group_controlled_buses()` collects every bus an active **remote** generator or an active voltage-mode SVC aims at. Works in grid ids and reads only input data, so it is valid while `fillpv_pq` is still building the solver labelling.
- `LSGrid::fillpv_pq` keeps those buses out of PV; the PQ loop picks them up.
- `fill_voltage_control_solver_data` enrols the local regulators standing on them as group members.
- `get_free_vm_slack_solver_buses` follows the same rule, so a local regulator on a group-controlled slack no longer counts as pinning it.

**The safety constraint worth reviewing closely:** buses regulated *only* by things standing on them are deliberately excluded. Several machines on one bus all regulating it locally is the common configuration, handled by the per-bus reactive redistribution in `GeneratorContainer::set_q`, and sweeping it into the bordered formulation would have been a large unintended behaviour change. The trigger is specifically the arrival of a *remote* controller (or an SVC, which is always a group controller). There are regression tests pinning that the multi-local-generator case still takes the PV path, builds no group, and gives bit-identical voltages.

## Second commit: HVDC converter stations

A VSC station with `voltage_regulator_on` pins its bus through the very same `fillpv` path, so it belongs in the group machinery on the same footing. The first commit could not express that (the group only knew `GEN` and `SVC`) and had to refuse such grids from `fillpv_pq` to avoid silently discarding the station's regulation. That exception is gone:

- `VoltageControlSolverData::Kind` gains `HVDC_SIDE_1` / `HVDC_SIDE_2`, carrying the hvdc **line** id in `elem_id` and the side in the kind.
- `fill_voltage_control_solver_data` enrols an active voltage-regulating station whose bus is group-controlled.
- `compute_results` writes the solved reactive output back to the station; the write-back switch now *rejects* an unknown kind rather than falling through to the SVC branch.

This composes cleanly because a station's sharing key is a reactive range in **MVAr**, the same currency as a generator's — so mixed groups are well defined. A group of a local generator, a remote generator and a station splits `Q` three ways by range (tested).

## Now working

- a remote controller onto a locally regulated bus;
- a remote controller onto a slack bus pinned by its own generator (the slack generator joins the group, MultiSlack grants the free `Vm`);
- a remote controller onto a slack bus that nothing pins — separately, by giving the regulated-bus check the `has_free_q` escape hatch the controller-bus check already had;
- an HVDC station sharing a regulated bus with generators, on either end of the line.

Disagreeing setpoints between a local and a remote regulator of one bus now surface as `conflicting voltage setpoints` rather than a bus-classification error.

## Still refused, on purpose

- **A remote regulator sharing its own bus with a local regulator.** This genuinely has no solution: whatever reactive power it injects at that bus is absorbed one-for-one by the machine holding the magnitude with free `Q`, so it has zero influence on its remote target. No reclassification can help; the "its OWN bus has no reactive (Q) equation" error still fires.
- **A voltage-mode SVC sharing a group.** Its key is a *susceptance* range, not a reactive range, and `NRSystem.hpp:591` records that the Phase 0 probes deliberately did not pin OLF's behaviour for mixed sharing. So for SVCs this PR delivers uniform classification and a better error, not new capability — worth stating plainly.

## Not covered

A converter station regulating a **remote** bus. Stations have no regulated-bus field at all today, so that would mean a new container member plus changes to `init_hvdc_lines`, the binary state, and the python bindings — a public API extension rather than a fix, so it is left out. Generators remain the only element that can target a remote bus.

## Tests

`src/tests/test_voltage_control_reclassify.cpp` (13 cases) and `lightsim2grid/tests/test_voltage_control_reclassify.py` (7 cases): the classification rule itself, each newly-supported configuration including the reactive split, the ones that must stay refused, and the regression guards for the ordinary PV path.

With the reclassification neutralised, 5 of 9 original cases fail and the 4 that pass are exactly the ones that should (multi-local-gen, the no-solution rejection, the SVC control, and the independent unpinned-slack relaxation). With station enrolment neutralised, the 3 station cases fail — on the *controller-count* assertion, i.e. the station gets silently dropped, which is the failure mode being guarded.

## Verification

Full C++ suite green (4850 assertions / 161 cases) in Release, under `__DEBUG_ASSERTS=1`, and pinned to C++14. Python: `test_hvdc_batch`, `test_voltage_control_{reclassify,remote_gen,svc}`, `test_multi_slack`, `test_ACPF`, `test_DCPF`, `test_SecurityAnlysis_cpp`, `test_LSGrid_out_of_bounds` all green.

Independent of #176: disjoint code, either merge order works.

---

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -123,6 +123,50 @@ TODO: Levenberg-Marquardt damping (a.k.a. Tikhonov-regularized Newton) : adding 
 
 [1.0.0] 2026-xx-yy
 --------------------
+- [FIXED] a generator could not remotely regulate a bus that another generator already
+  regulated *locally* -- the grid was refused outright with
+  ``LSGrid::fill_voltage_control_solver_data: generator N regulates bus X which has no voltage (Vm)
+  unknown`` -- even though the configuration is perfectly well posed. Two machines holding one bus
+  simply form a control group: one bordered voltage row, one reactive unknown each, and a sharing
+  row splitting the reactive duty in proportion to their ``qmax - qmin`` ranges, exactly as two
+  *remote* controllers on one bus already did.
+  The cause was the order in which buses are classified. ``GeneratorContainer::fillpv`` knew how to
+  keep a controller's OWN bus out of the PV set (a remote regulator leaves its own bus PQ), but
+  nothing stopped a LOCAL regulator from claiming a bus that somebody else regulated remotely. That
+  bus then had no ``Vm`` unknown, so ``VoltageControl::register_in`` got ``vm_col == -1`` for it, the
+  group's voltage row was never given its ``+1`` entry, and the row came out structurally EMPTY --
+  a singular Jacobian. The check that refused the grid was really reporting the misclassification,
+  not a property of the configuration.
+  Bus classification now runs the other way round: ``LSGrid::get_group_controlled_buses`` collects
+  every bus an active *remote* generator or an active voltage-mode SVC aims at, ``LSGrid::fillpv_pq``
+  keeps those out of PV, and ``fill_voltage_control_solver_data`` enrols the local regulators sitting
+  on them as members of the group. Buses regulated only by generators standing on them are
+  deliberately untouched: several machines sharing a bus and all regulating it locally remains the
+  ordinary PV case, with the per-bus reactive redistribution of ``GeneratorContainer::set_q``.
+  Three configurations start working as a result: a remote controller onto a locally regulated bus,
+  a remote controller onto a slack bus pinned by its own generator (the slack generator joins the
+  group and MultiSlack grants the free ``Vm``), and -- separately, by giving the regulated-bus check
+  the ``has_free_q`` escape hatch the controller-bus check already had -- a remote controller onto a
+  slack bus that nothing pins. Disagreeing setpoints between a local and a remote regulator of one
+  bus are now reported by the group's own check (``conflicting voltage setpoints``) rather than as a
+  bus-classification error.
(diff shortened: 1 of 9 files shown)


BDonnot (commented):
Ok

PR #178, opened 2026-08-08
Take the whole injection from the grid in TimeSeries, not a subset

BDonnot:
## The bug

`TimeSeriesCPP` solves each step against an **incomplete injection vector**. It does not use the `Sbus` the gridmodel builds — it rebuilds its own, per step, out of the four matrices `compute_Vs` receives (generator and static-generator **active** power, load active + reactive power) plus the hvdc term read from the grid.

`LSGrid::fillSbus_me` stamps considerably more, and everything else was silently dropped. Measured on a 4-bus feeder, one-step `TimeSeries` against `ac_pf` with the same injections:

| dropped | in `LSGrid` Sbus | worst \|ΔV\| |
|---|---|---|
| loads P+Q, gens P, sgens P, hvdc | ✓ covered | control: **0** |
| gen Q when `voltage_regulator_on == false` | `tmp += i*target_q` | **0.032 pu** |
| static generator Q | `{p, q}` | **0.052 pu** |
| storage P **and** Q | `-= {p, q}` — absent altogether | **0.066 pu** |
| REACTIVE_POWER-mode SVC Q | ✓ — absent altogether | **0.081 pu** |
| DC phase shifter | `hack_Sbus_for_dc_phase_shifter` | dc only |

For the storage and SVC rows the `TimeSeries` voltages came back *bit-identical to a grid without the element*: it contributed nothing at all. The powerflow converged happily on an injection the caller never asked for.

**DC is affected too.** `compute_one_powerflow` falls back to the gridmodel's `Pbus_` only when *no* complex `Sbus` is supplied, which is not the `TimeSeries` case — it takes `_Sbuses.row(i).real()`, so a storage unit's active power went missing in DC as well, along with the phase-shifter term.

`ContingencyAnalysisCPP` (and `SecurityAnalysis`, which wraps it) was **never affected**: it hands the gridmodel's own complete `Sbus_` straight to the solver. Only `TimeSeries` substitutes a reconstruction.

## The fix

`TimeSeries::_constant_sbus_pu` derives the remainder instead of enumerating it: the complete per-unit injection the gridmodel just built (`Sbus_` in ac, `Pbus_` in dc) **minus** the very reconstruction `compute_Vs` performs, evaluated at the gridmodel's own target values. The two cancel, and what is left is by construction everything the per-step matrices do not carry.

That choice is the point of the PR. This bug *is* an enumeration that fell behind — only the hvdc term was ever listed — so replacing it with a longer list would just reset the clock. Deriving it means an element type added to `fillSbus_me` later is picked up for free. It also handles the DC phase-shifter term correctly without special-casing, that one being deliberately in physical units (`fillSbus_me` applies it *after* the `sn_mva` division).

Also adds `OneSideContainer_PQ::get_target_q()`, the counterpart of the existing `get_target_p()`, needed to evaluate the load reactive share of the subtraction.

## Tests

`src/tests/test_timeseries_sbus.cpp` (9 cases) and `lightsim2grid/tests/test_timeseries_sbus.py` (3 cases) assert the real invariant: **fed the grid's own injections, a one-step time series must reproduce `ac_pf` / `dc_pf` exactly.** Each dropped element kind separately and all of them together, in ac and in dc, plus a multi-step case (the remainder is constant, so it has to reach every row, not just the first) and a control grid whose whole injection the per-step matrices already covered.

**Eight of the nine C++ cases fail without the fix**; the control is the one that passes. The python side includes a guard-the-guard case asserting the added elements actually move the operating point, so the comparison cannot pass vacuously — that trap is exactly what made this bug survive the existing `TimeSeries` tests, which run on grids where every missing term happens to be zero.

## Verification

Full C++ suite green — 5111 assertions / 178 cases — in Release, under `__DEBUG_ASSERTS=1`, and pinned to C++14. Python: `test_hvdc_batch`, `test_Computers`, `test_TimeSerie`, `test_SecurityAnlysis_cpp`, `test_voltage_control_batch`, `test_voltage_control_reclassify`, `test_ACPF`, `test_DCPF`, `test_LSGrid_out_of_bounds` all green.

## One adjacent issue, not fixed here

`GeneratorContainer::fillSbus` skips a *pseudo-off* PV generator (`target_p == 0` under `turnedoff_no_pv`), while PV classification is computed once from the grid's `target_p` and never revisited per step. A `gen_p` series that crosses zero can therefore disagree with the PV set `TimeSeries` fixed up front. Same file, different bug — left alone deliberately.

---

### DIFF
--- a/src/core/batch_algorithm/TimeSeries.cpp
+++ b/src/core/batch_algorithm/TimeSeries.cpp
@@ -183,11 +183,27 @@ CplxVect TimeSeries::_constant_sbus_pu(bool ac_solver_used) const
     const auto & s_generators = _grid_model.get_static_generators_as_data();
     const auto & loads = _grid_model.get_loads_as_data();
 
+    // The gridmodel's target vectors, presented to fill_SBus_* as the single-step
+    // matrices it expects, WITHOUT copying them: a 1 x n row-major matrix has exactly
+    // the memory layout of a contiguous n-vector, so an Eigen::Map over the vector's
+    // own storage is a view. Handing `.transpose()` over instead would not be -- the
+    // transpose of a column vector is not a RealMat, so binding it to the
+    // Eigen::Ref<const RealMat> parameter has to materialise it. The Refs are held in
+    // named locals rather than being called inline: that way the Map never depends on
+    // a temporary Ref's lifetime, whatever the accessors return.
+    const Eigen::Ref<const RealVect> gen_p_v = _grid_model.get_gen_target_p();
+    const Eigen::Ref<const RealVect> sgen_p_v = _grid_model.get_sgen_target_p();
+    const Eigen::Ref<const RealVect> load_p_v = _grid_model.get_load_target_p();
+    const Eigen::Ref<const RealVect> load_q_v = loads.get_target_q();
+    const Eigen::Map<const RealMat> gen_p(gen_p_v.data(), 1, gen_p_v.size());
+    const Eigen::Map<const RealMat> sgen_p(sgen_p_v.data(), 1, sgen_p_v.size());
+    const Eigen::Map<const RealMat> load_p(load_p_v.data(), 1, load_p_v.size());
+    const Eigen::Map<const RealMat> load_q(load_q_v.data(), 1, load_q_v.size());
+
+    // `accounted` does have to be its own buffer: it is written by fill_SBus_* and
+    // then scaled by sn_mva, which must NOT touch `res` (already per-unit). One
+    // vector-sized allocation per compute_Vs call, not per step.
     CplxMat accounted = CplxMat::Zero(1, nb_buses_solver_);
-    const RealMat gen_p = _grid_model.get_gen_target_p().transpose();
-    const RealMat sgen_p = _grid_model.get_sgen_target_p().transpose();
-    const RealMat load_p = _grid_model.get_load_target_p().transpose();
-    const RealMat load_q = loads.get_target_q().transpose();
     bool add_ = true;
     fill_SBus_real(accounted, generators, gen_p, id_me_to_solver_, add_);
     fill_SBus_real(accounted, s_generators, sgen_p, id_me_to_solver_, add_);
--- a/src/core/batch_algorithm/TimeSeries.cpp
+++ b/src/core/batch_algorithm/TimeSeries.cpp
@@ -80,17 +80,26 @@ int TimeSeries::compute_Vs(const Eigen::Ref<const RealMat> & gen_p,
     fill_SBus_real(_Sbuses, loads, load_p, id_me_to_solver_, add_);
     fill_SBus_imag(_Sbuses, loads, load_q, id_me_to_solver_, add_);
 
-    // add the (constant accross the steps) hvdc injections: fixed-setpoint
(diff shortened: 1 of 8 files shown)


src/core/batch_algorithm/TimeSeries.cpp
```diff
@@ -147,4 +156,49 @@ int TimeSeries::compute_Vs(const Eigen::Ref<const RealMat> & gen_p,
...
+    // own target values, so that the two cancel and only the rest is left
+    const auto & generators = _grid_model.get_generators_as_data();
+    const auto & s_generators = _grid_model.get_static_generators_as_data();
+    const auto & loads = _grid_model.get_loads_as_data();
+
+    CplxMat accounted = CplxMat::Zero(1, nb_buses_solver_);
+    const RealMat gen_p = _grid_model.get_gen_target_p().transpose();
+    const RealMat sgen_p = _grid_model.get_sgen_target_p().transpose();
+    const RealMat load_p = _grid_model.get_load_target_p().transpose();
+    const RealMat load_q = loads.get_target_q().transpose();
```
BDonnot: Why making copies here ? Eigen::Ref<const ...> or bare CPP reference would be better here ?

BDonnot (commented):
Check on copy

src/core/batch_algorithm/TimeSeries.cpp
BDonnot: Good catch — those four were genuine copies. Fixed in f72d7b0.

Worth noting that `Eigen::Ref<const ...>` on its own would **not** have fixed it: the transpose of a column vector is not a `RealMat`, so binding that expression to `fill_SBus_*`'s `Eigen::Ref<const RealMat>` parameter has to evaluate it into the `Ref`'s own storage. The copy would just have moved from an explicit local into the `Ref`.

What does work is a `Map`: a `1 x n` row-major matrix has exactly the memory layout of a contiguous `n`-vector, so mapping the vector's own storage as `1 x n` is a view, and the `Ref` then binds to that `Map` directly because layout and strides match.

```cpp
const Eigen::Ref<const RealVect> gen_p_v = _grid_model.get_gen_target_p();
const Eigen::Map<const RealMat>  gen_p(gen_p_v.data(), 1, gen_p_v.size());
```

I checked this rather than assumed it — the `Map`'s data pointer equals the container's, and it is *still* equal when observed from inside a function taking the `Eigen::Ref<const RealMat>` parameter, so no copy happens at either step:

```
Map is a view (data ptr equal): 1
inside Ref-taking fn: data==container? 1
```

The `Ref`s are held in named locals rather than called inline, so the `Map` never depends on a temporary `Ref`'s lifetime whatever the accessors are changed to return later. The zero-length case (a grid with no static generators — `data() == nullptr`, `size 0`) builds and binds fine, verified under the Eigen/libstdc++ assertion build where a bad `Map` would trip an assert.

One thing I deliberately left as a copy, now with a comment saying why: `accounted`. It is written by `fill_SBus_*` and then scaled by `sn_mva`, which must not touch `res` (already per-unit), so it needs its own buffer. It is one vector-sized allocation per `compute_Vs` call, not per step.

Full suite still green: 5127 assertions / 179 cases, in Release, under ASan + UBSan, and with assertions live.

---

BDonnot (commented):
Ok

PR #181, opened 2026-08-13
speed(dc batch): skip building complex V eagerly, reconstruct lazily

BDonnot:
## Summary

- DC batch algorithms (`TimeSeriesCPP` / `InjectionSweepCPP` / `ContingencyAnalysisCPP` / `ScenarioSweepCPP`) no longer build the full complex voltage every row. `BaseDCAlgo::compute_pf_dc`'s only real output is `theta` (the linear solve's result); the magnitude is a pure, unmodified echo of the caller's input. The batch sweep used to build the complex `V` anyway (a `std::polar`/hardware-sincos call per bus per row), only for the flow / current-limit-violation computations to immediately un-build it again with `.arg()`/`.abs()` — a wasted sincos-then-atan2 round trip.
- A new `BaseAlgo::set_lazy_v` toggle (DC-only, no-op for AC/plugin solvers) lets a DC solve skip `Vm_`/`V_` entirely and only produce `theta`. The per-row sweep (outside the "handle disconnected grid" masked path, which still needs per-row complex `V` for its violation checks and stays on the eager path) now accumulates `theta` into a plain real matrix and computes amps/power flows straight from it, with no `.arg()`.
- The (small, per-generator) voltage magnitude is reconstructed only if/when something actually asks for it — `get_voltages()` (lazily, cached on first call) or the amps flows — from the grid's own target voltages plus `modify_gen_v` if set, mirroring exactly what the row loop itself already used to seed each solve.
- Second commit fixes a correctness bug this surfaced: a row/bus that was never actually part of a converged solve (an islanding DC contingency that diverges, or a grid bus outside the solver's bus set) must reconstruct as exact complex `0`, not as a `0`-theta paired with a nonzero magnitude. Caught by CI (`test_ContingencyAnalysis_split.py`, `test_DCSecurityAnlysis.py`).

Benchmarked on a synthetic 2000-bus / 200-generator / 1800-load ring grid, `InjectionSweep` DC, 3000 steps, single-threaded:

| Stage | Before | After |
|---|---|---|
| `compute_Vs` (the solve) | 772 ms | 619 ms (**-20%**) |
| `compute_power_flows` | 408 ms | 204 ms (**-50%**) |
| `compute_flows` (amps) | 445 ms | 211 ms (**-53%**) |
| Total, if `get_voltages()` is never called | 1625 ms | 1035 ms (**-36%**) |

## Test plan

- [x] Full C++ Catch2 suite: 194 test cases / 5296 assertions, unchanged.
- [x] `lightsim2grid/tests/test_ContingencyAnalysis_split.py` (13/13) and `test_DCSecurityAnlysis.py` (17/17), the two files CI caught the regression in — verified locally with `pandapower` + `grid2op` installed.
- [x] Broader DC/batch-related Python test sweep (`InjectionSweep`, `TimeSeries`, `ScenarioSweep`, `ContingencyAnalysis`, DC*) — all green except pre-existing, environment-only gaps in this sandbox (no `pypowsybl`/`gymnasium`, no KLU solver build, missing bundled grid2op test data files) unrelated to this change.
- [x] Sanitizers CI workflow (ASan+UBSan / Eigen assertions, Python + C++ jobs) re-run automatically on push; the originally-reported failing job is expected green on the fix commit.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WSNsyua1QsDg9SLAtdBWBx)_

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1978,6 +1978,32 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
   vm vector, alongside the existing single-arg overload reading the grid's own
   ``target_vm_pu``). Left unset (the default), every row keeps using the grid's own
   ``target_vm_pu``, exactly as before this setter existed.
+- [IMPROVED] speed (DC batch algorithms -- ``TimeSeriesCPP`` / ``InjectionSweepCPP`` /
+  ``ContingencyAnalysisCPP`` / ``ScenarioSweepCPP``): a DC solve no longer builds the
+  complex voltage every row. ``BaseDCAlgo::compute_pf_dc`` only ever *outputs* the bus
+  angles (``theta``, the linear solve's actual result) -- the magnitude ``Vm_`` is a
+  pure, unmodified echo of the caller's input, and the complex ``V_`` it used to build
+  from both (a ``std::polar`` / hardware ``sincos`` call per bus) existed only so
+  downstream code could read it back. That downstream code (``compute_amps_flows`` /
+  ``compute_active_power_flows``) was then immediately un-building it again with
+  ``.arg()`` (and, for the current-limit checks, ``check_current_violations`` did the
+  same) -- a full sin/cos-then-atan2 round trip, per bus, per row, for values that were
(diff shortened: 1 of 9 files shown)


BDonnot (commented):
Ok

PR #183, opened 2026-08-22
Automatic per-family solver cache reuse (and the segfault behind unset_changes)

BDonnot:
## Why

`unset_changes()` was an opt-in performance footgun: most users never called it and silently paid for a full rebuild on every powerflow. It was also *unsafe* — it asserted "the cached solver data matches the grid" and nothing checked the claim.

Three sequences of documented API calls segfaulted (release wheels are `-O3 -DNDEBUG`, so this was an out-of-bounds read, not an error):

```
unset_changes(); ac_pf();           # nothing was ever built
dc_pf(); unset_changes(); ac_pf();  # what was built belongs to the other family
ac_pf(); unset_changes(); dc_pf();  # idem
```

The second one is why `LightSimBackend.runpf` carried a `self._last_dc` workaround with the comment *"otherwise might segfault"*, and the first is what `check_solution` guarded against locally.

## What changed

**Cache reuse is automatic, per solver family, and on by default.** Every powerflow marks its own family "in sync" on the way out, so `unset_changes()` is no longer needed — it is a no-op when reuse is enabled, and is kept for backward compatibility. Worth ~24% of the time of a powerflow on a 14-bus grid, with nothing to call.

**The AC and DC families no longer share any solver-side data.** `slack_weights`, the PV split and the PQ split were single members overwritten by whichever family solved last (unlike Ybus/Sbus/the id maps, already per family). That is what made a "nothing changed" claim unsound across families. Each family now owns its own, exposed as `get_ac_pv_solver` / `get_dc_pv_solver` and the `pq` / `slack_weights` equivalents. The family-less accessors answer for AC once an AC powerflow has run and fall back to DC — which also fixes them: they used to relabel whatever the last solve left behind with the *AC* bus mapping.

**Safety lives where the reuse decision is made.** `_pre_process_solver_impl` checks, in a handful of integer comparisons, that the data the flags describe is actually there, and falls back to a full rebuild otherwise. A wrong "nothing changed" can now cost time, never memory safety.

### New API

| | |
|---|---|
| `allow_cache_reuse(bool)` / `allow_ac_…` / `allow_dc_…` (+ `get_*`) | durable mode; off ⇒ that family rebuilds every solve |
| `prevent_cache_reuse()` / `prevent_ac_…` / `prevent_dc_…` | one-shot invalidation; new name of `tell_solver_need_reset()`, which still works |

The answer is identical either way, so `allow_*` is a debugging switch ("is this a caching artefact?") and an escape hatch for code that mutates the C++ containers directly. Both are now documented rather than tagged *"internal, do not use"*, with a new **Solver cache reuse** section in `docs/network.rst`.

### Divergence

The data a diverged solve built is a correct picture of the grid, so it is kept and stays reusable. The algorithm's own state (half-converged iterate, factorization of a system it gave up on) is reset, and the next powerflow of that family rebuilds its internals from the cached matrices.

### Serialization and copy

Nothing the solvers cache crosses a serialization boundary — `StateRes` carries the elements, everything else is derived and rebuilt. Already true; now pinned by tests, because the reset that enforces it carries a *"TODO see if it's worth the trouble NOT to do it"*. The reasoning is a security one: a cache read from a file is a second copy of state that `check_grid()` cannot verify against the elements it claims to describe. `set_state` also no longer reads the serialized bus-connectivity snapshot back into the cache snapshots.

`LSGrid.copy()` also starts cold. Unlike serialization that is a choice, not a requirement — pinned and documented so it can be revisited.

## Bugs found on the way (all pre-existing)

- **`set_sn_mva()` invalidated nothing**, though `sn_mva` is passed to `fillYbus`/`fillBdc` and divides `fillSbus_me`: a powerflow after it solved the previous base power's system. Found by the completeness sweep below.
- **`dc_pf()` returned a previous call's voltage magnitudes.** DC solves angles only, so its magnitudes are the ones it was handed — but `BaseDCAlgo` skipped `Vm_ = V.array().abs()` unless `_solver_control.has_v_changed()`, which asks whether the *grid's* setpoints changed and says nothing about the vector just passed in. Live today after every `unset_changes()`, which `LightSimBackend` calls after each step. Not cosmetic: `a_ka = √(P²+Q²)/(√3·Vm·vn_kv)`, so a stale `Vm` moved reported branch currents by **up to 4.7%** — and with them `rho` and DC contingency violations. The guard saved one `abs()` over `nb_bus` doubles; measured at 1888 and 9241 buses it is invisible next to the sparse solve.
- **A diverging powerflow** left the algorithm's internal state in place for the next call and discarded nothing.

## Tests

`src/tests/test_cache_reuse.cpp` covers three properties:

- **It is invisible** — a completeness sweep applies *every* mutating `LSGrid` method to a grid carrying a live cache and to one that rebuilds everything, in AC and in DC, requiring identical voltages. A setter that forgets to invalidate what it touched shows up here; that is how the two bugs above were found.
- **It is safe** — the sequences that used to segfault, plus `set_state`. Four of five sections crash without the guard.
- **It is controllable** — defaults, per-family marking, both switches, no cross-family write-through, algorithm change, divergence under every built-in AC algorithm, copy construction, and the `[serialization]` cold-start contract (removing the reset from `set_state` fails 12 assertions).

`lightsim2grid/tests/test_cache_reuse.py` covers the Python-facing API, and `TestDcThenAcSameBackend` in `test_DoNothingACDC.py` covers the backend sequence the removed `_last_dc` workaround used to protect.

**Verified:** 207 C++ test cases / 5737 assertions green under Release, ASan+UBSan, Eigen/libstdc++ assertions, and valgrind (`--leak-check=full --track-origins=yes`: 0 errors, no leaks). `test_solver_control.py` (139 tests) and the rest of the Python suite pass — the only failures in a full run are grid2op test-data files absent from the build machine, identical to the pre-change baseline.

## Breaking

- Code that mutates the grid through a path bypassing `LSGrid`'s own `change_*` / `deactivate_*` / `reactivate_*` methods used to get away with it as long as it never called `unset_changes()`. It must now say so with `prevent_cache_reuse()` (or the narrower `tell_recompute_ybus` / `tell_recompute_sbus`), or turn reuse off with `allow_cache_reuse(False)`.
- The AC/DC split of `slack_weights` / PV / PQ (C++ API; the family-less Python accessors keep working).
- `LightSimBackend._last_dc` (private) is removed along with its workaround.

https://claude.ai/code/session_01VVw4c45xkdJ7eRdALtsPkw

---
_Generated by [Claude Code](https://claude.ai/code/session_01VVw4c45xkdJ7eRdALtsPkw)_


BDonnot (commented):
Ok

PR #184, opened 2026-08-23
Solver-side data of a powerflow belongs to exactly one owner

BDonnot:
Two commits. The first is tests only. The second is the actual fix.

This started from the question "now that the cache is reused by default, can the consistency check go from the internal path?". Answer: no, and measuring why turned up a separate, real ownership bug in the batch path.

---

## Commit 1 — `test: pin the structural half of the cache-consistency guard`

**No production code.** One test case added to `src/tests/test_cache_reuse.cpp`.

### Why the check can't go

Cost of the guard, callgrind instruction counts on the exotic-elements IEEE14 grid, taken as the slope between 200 and 1200 powerflows so grid construction cancels:

| | with guard | without | delta |
|---|---|---|---|
| AC, one load changed then solve | 839 340 instr/pf | 839 290 | −51 (0.006 %) |
| DC, warm cache, nothing changed | 28 095 instr/pf | 28 054 | −42 (0.15 %) |

The DC row is the floor — the cheapest powerflow this library can run (2.8 µs here). Invisible in wall clock: run-to-run spread on one unchanged binary was several times the difference.

And it is still load-bearing. `unset_changes()` returns early only when **both** families cache automatically. Switching one off re-arms it, and it marks *both* families "in sync" — including the one still allowed to reuse, whose solver-side data may never have been built. All three of these segfault under `-O3 -DNDEBUG` with the structural comparisons removed:

```python
grid.allow_ac_cache_reuse(False); grid.unset_changes(); grid.dc_pf(...)
grid.allow_dc_cache_reuse(False); grid.unset_changes(); grid.ac_pf(...)
grid.ac_pf(...); grid.allow_ac_cache_reuse(False); grid.unset_changes(); grid.dc_pf(...)
```

### The coverage gap this closes

Every existing `[unset_changes]` section reaches it through `allow_cache_reuse(false)` — **both** families — so the family that runs is stopped by the guard's first term (`!allow_*_cache_reuse_`) before any structural comparison matters. **The whole suite passes with the structural half deleted.** The new test case leaves one family automatic, which is what actually exercises it: it fails with SIGSEGV without the guard and passes with it.

---

### What the bug is

`pre_process_solver` / `pre_process_dc_solver` serve two kinds of caller:

- **The grid itself** (`ac_pf` / `dc_pf` / `check_solution`) hands over its own per-family members. What comes out *is* the family cache, and this grid's algorithm consumes it.
- **A foreign builder** — the batch algorithms, via `BaseBatchSolverSynch::prepare_solver_input_base` — hands over containers it owns (`Ybus_`, `Sbus_`, `id_me_to_solver_`, …) and solves them with an algorithm of its own.

Nine containers make up the group. Six were parameters. **Three — `slack_weights`, `bus_pv`, `bus_pq` — were not:** they were bound to `slack_weights_ac_` / `bus_pv_ac_` / `bus_pq_ac_` (or the `_dc_` set) whatever the caller passed for the other six. That was the old first three lines of `_pre_process_solver_impl`.

Three consequences, all silent:

1. **A batch build wrote its pv-pq split and slack weights into the grid's own cache**, next to a `Ybus` / `Sbus` built for a different bus labelling. The result is the right size and passes every structural check — and the grid went on claiming that cache was reusable. Checked directly against the pre-fix core: `need_reset_solver()` stays **false** across a foreign build that just replaced the labelling.

2. **The cache-consistency guard sized one owner's containers against the other's.** Sizes agree far more often than labellings do, so a caller reusing its containers with a "nothing changed" control would have skipped `fillpv_pq` and stamped *this grid's* pv-pq split onto *its* system — converged, plausible, wrong. Unreachable today only because every batch path calls `tell_all_changed()` first. I.e. until the first one that tries to keep its `Ybus` across scenarios, which is the obvious next optimization in there.

3. **`_algo.reset()` and `_algo.tell_solver_control()` fired on the grid's own algorithm** for a solve it would never perform, discarding a factorization the grid's next powerflow would have reused.

**To be clear about blast radius:** `BaseBatchSolverSynch` holds `LSGrid _grid_model` *by value*, so nothing user-facing is broken today. This is preventive.

### What changed

Read in this order:

- **`src/core/LSGrid.cpp:1414-1454`** — the new ownership block at the top of `_pre_process_solver_impl`. It replaces the three `is_ac ? member : member` bindings. `_nb_own_cache_args` (defined at `src/core/LSGrid.hpp:2128`) counts how many of the nine containers are this grid's own — nine pointer comparisons summed branchlessly. Nine means the grid called us; zero means a foreign builder; anything in between throws rather than building a chimera. That throw is an in-tree invariant only: the members are private, so a mixed call cannot be constructed from outside the class.

- **the guard, just below** — gains `!own_cache ||` as its first term. `solver_control`, the connectivity snapshot `init_bus_status()` compares against, and the flags it raises all describe *this* grid and say nothing about a caller's containers. So a foreign build always rebuilds. Free — they all asked for it anyway — and it makes the mixed-owner fast path *impossible* rather than merely unreached.

- **the two algorithm blocks** — now gated on `own_cache`. The grid's algorithm is only touched when it is the one that will run.

- **`src/core/LSGrid.cpp:1624-1668`** — the publication block, reworked. A foreign build still has to write its labelling into the members, because the NR extensions read the members rather than what the solver was handed. **Note `bus_pq_ac_` in that list**: `fill_voltage_control_solver_data` needs it, and the old write-through supplied it *by accident*. So the write-through wasn't purely a mistake — it was load-bearing and undocumented. It is now explicit, complete, and paid for: `prevent_ac_cache_reuse()` / `prevent_dc_cache_reuse()` at the end, so the grid rebuilds rather than reusing something half its own and half the caller's.

- **`src/core/batch_algorithm/BaseBatchSolverSynch.hpp:508`** — the batch clears and passes its own `bus_pv_` / `bus_pq_` / `slack_weights_`, and no longer fetches them back out of `_grid_model` afterwards. That also removes the three copies its own `TODO copies are made here, which is not ideal` flagged.

- signature change on the two `pre_process_*` methods (three added parameters). **C++ only — neither is exposed to python**, and the plugin API (`ls2g_api.hpp`) does not reference them.

### Cost

Same callgrind method as above:

| | before | after | delta |
|---|---|---|---|
| AC powerflow | 839 340 instr/pf | 839 431 | +90 (+0.011 %) |
| DC, warm cache | 28 095 instr/pf | 28 194 | +99 (+0.35 %) |

The DC figure is the ceiling. Nothing shows in wall clock. Most of it is argument marshalling for three more reference parameters, not the ownership check itself.

### Tests

- `src/tests/test_cache_reuse.cpp:369` — a build into a caller's containers leaves the grid rebuilding rather than half-replaced (AC, DC, and that one family's foreign build leaves the other family's cache alone).
- `src/tests/test_batch_voltage_control.cpp:319` — the source grid still solves identically on both sides of a `TimeSeries` and a `ContingencyAnalysis`.

**Worth knowing before you read them:** that batch-level test **passes against the pre-fix core too**. Today the batch rebuilds a labelling identical to the grid's, so the mixture is benign — it guards the contract rather than reproducing a live failure. What does separate pre- from post-fix is the invariant in the `test_cache_reuse.cpp` sections: pre-fix, a foreign build left `need_reset_solver()` false.

### Verified

210 test cases / 5814 assertions pass under Release, under C++14 (`LS2G_CXX_STANDARD=14`), under ASan+UBSan, and under valgrind (0 errors, no leaks).

The python layer could **not** be exercised — no pybind11 / numpy on the machine this ran on. No python-visible API changed, but `lightsim2grid/tests/` has not been run against this.

---

---

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -149,6 +149,25 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
   a one-shot invalidation of what a family cached (as opposed to the ``allow_*`` mode above).
   ``prevent_cache_reuse()`` is the new name of ``tell_solver_need_reset()``, which still exists and
   behaves identically. Both are now documented rather than tagged "internal, do not use".
+- [FIXED] the solver-side data a powerflow is built from now belongs to exactly one owner.
+  ``LSGrid::pre_process_solver`` / ``pre_process_dc_solver`` (C++ only, not exposed to python)
+  serve two kinds of caller: the grid itself (``ac_pf`` / ``dc_pf`` / ``check_solution``, which
+  hand over their own members, so what comes out IS the family cache) and the batch algorithms
+  (``TimeSeries`` / ``ContingencyAnalysis`` / the sweeps, which own their containers and solve them
+  with an algorithm of their own). The slack weights and the PV / PQ split were exempt: they were
+  taken from the grid's members whatever the caller passed for everything else. So a batch build
+  wrote its own split into the grid's cache next to a ``Ybus`` / ``Sbus`` built for a different
+  labelling, and the cache-consistency guard compared one owner's container sizes against the
+  other's -- sizes agree far more often than labellings do, so a caller that reused its containers
+  with a "nothing changed" control would have skipped ``fillpv_pq`` and stamped the grid's PV / PQ
+  split onto its own system: converged, plausible, wrong. Nothing reached it (every batch asks for
+  a full rebuild), which is what made it worth closing before something did.
+  The three vectors are now parameters like the rest of the group, ``allow_cache_reuse`` fast paths
+  are reserved to the owner of a cache, mixing the two owners' containers in one call raises
+  instead of building a chimera, and a build into a caller's containers leaves the grid it was
+  called on rebuilding rather than reusing what is now half its own and half the caller's. This
+  changes the signature of the two ``pre_process_*`` methods (three added parameters) for C++
+  callers; nothing changes for python, and the numbers a batch returns are unchanged.
 - [BREAKING] the AC and the DC solver families no longer share ANY solver-side data.
   ``slack_weights``, the PV split and the PQ split were single members overwritten by whichever
   family solved last -- unlike ``Ybus`` / ``Sbus`` / the id maps, which were already per family.
--- a/src/core/LSGrid.cpp
+++ b/src/core/LSGrid.cpp
@@ -754,6 +754,9 @@ CplxVect LSGrid::ac_pf(const Eigen::Ref<const CplxVect> & Vinit,
                                     id_ac_solver_to_me_,
                                     slack_bus_id_ac_me_,
                                     slack_bus_id_ac_solver_,
+                                    slack_weights_ac_,
+                                    bus_pv_ac_,
+                                    bus_pq_ac_,
                                     is_ac,
                                     algo_controler_.ac_algo_controler());
 
@@ -1293,6 +1296,9 @@ CplxVect LSGrid::check_solution(const Eigen::Ref<const CplxVect> & V_proposed, b
                                     id_ac_solver_to_me_,
                                     slack_bus_id_ac_me_,
                                     slack_bus_id_ac_solver_,
+                                    slack_weights_ac_,
+                                    bus_pv_ac_,
+                                    bus_pq_ac_,
                                     is_ac, reset_solver,
                                     false);  // do NOT snap regulated buses to their target: we are testing V_proposed as-is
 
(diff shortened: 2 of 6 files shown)


BDonnot:
Closing in favour of #185, which fixes the same defect the other way round.

This PR kept the eighteen loose per-family members and added machinery to check they agree: three more parameters on `pre_process_solver`, a nine-way pointer-comparison count, and a throw when only some of them were the grid's own. That is a runtime check for something that should not be expressible in the first place.

#185 starts fresh from `dev_0.13.2` and makes the cached data one object (`SolverSideCache`), so the nine parts cannot be handed over separately at all. It deletes more than it adds — the ownership count and its throw are gone entirely, and so are the `if(ac) x_ac_ else x_dc_` bodies in `_mark_cache_valid` / `prevent_*_cache_reuse` / `init_bus_status`. The `unset_changes` test from here is carried over.

---

PR #185, opened 2026-08-23
A solver family's cached data is one object, not eighteen members

BDonnot:
One commit, fresh from `dev_0.13.2`. Replaces #184, which took the opposite approach (add checks that the eighteen loose members agree) and is closed.

## The problem, concretely

`pre_process_solver` fills in containers you hand it. Two callers hand it different ones:

```cpp
// LSGrid::ac_pf -- passes the grid's own members
pre_process_solver(Vinit, acSbus_, Ybus_ac_, id_me_to_ac_solver_, ...);

// BaseBatchSolverSynch::prepare_solver_input_base -- passes the BATCH's members
_grid_model.pre_process_solver(Vinit, Sbus_, Ybus_, id_me_to_solver_, ...);
```

Six of the nine outputs were parameters. **Three were not.** At the top of `_pre_process_solver_impl`:

```cpp
RealVect & slack_weights = is_ac ? slack_weights_ac_ : slack_weights_dc_;
SolverBusIdVect & bus_pv = is_ac ? bus_pv_ac_ : bus_pv_dc_;
SolverBusIdVect & bus_pq = is_ac ? bus_pq_ac_ : bus_pq_dc_;
```

Hardcoded to the grid's members whatever the caller passed for the rest. One call, output split across two objects. The tell was already sitting in the batch:

```cpp
bus_pv_ = _grid_model.get_ac_pv_solver();   // reach back into the grid to collect
bus_pq_ = _grid_model.get_ac_pq_solver();   // the other half of my own output
slack_weights_ = _grid_model.get_ac_slack_weights_solver();
// TODO copies are made here, which is not ideal
```

Three consequences, all silent:

1. **A batch build wrote its pv-pq split and slack weights into the grid's cache**, next to a `Ybus` built for a different bus labelling — and the grid went on claiming that mixture was reusable. Checked against `dev_0.13.2`: `need_reset_solver()` stays **false** across such a build.
2. **The reuse guard sized one owner's containers against the other's.** Sizes agree far more often than labellings do, so a caller reusing its containers with a "nothing changed" control would have skipped `fillpv_pq` and stamped *this grid's* split onto *its* system — converged, plausible, wrong.
3. **`_algo.reset()` / `tell_solver_control()` fired on the grid's own algorithm** for a solve it would never perform, discarding a factorization its next powerflow would have reused.

**Nothing reaches any of this today.** The batch works on a private `LSGrid` copy and always calls `tell_all_changed()` first. Both are accidents; neither is written down as a requirement, and the obvious next optimisation in there (keep `Ybus` across scenarios) removes the second one.

## The fix: make them one object

`src/core/SolverSideCache.hpp`, split by **type**, not lifetime:

| | contents |
|---|---|
| `SolverBusLayout` | labelling, slack, pv-pq split, connectivity snapshot, `allow_reuse`, `algo_needs_rebuild` — everything whose type doesn't mention the family's scalar |
| `SolverSideCache<T>` | the above, plus `mat` and `inj`. `cplx_type` → AC (Ybus/Sbus), `real_type` → DC (Bbus/Pbus) |

The split exists so the batch — which runs AC *or* DC, never both — can name the family-agnostic half: 82 of its 150 call sites don't care which family, and go through `active_layout()` instead of a ternary at every use.

`LSGrid` holds `ac_cache_` / `dc_cache_`. `pre_process_solver` / `pre_process_dc_solver` take **one cache reference**. There is no way to hand either function half of one.

## What this deletes rather than adds

- The reuse guard is `cache.is_usable(nb_bus)` — one place, instead of eight comparisons written inline against a mix of two objects' members.
- "Is this my cache?" is `&c == &ac_cache_`, two overloads picked by the family's own type. #184 needed nine pointer comparisons and a throw on a partial match; that whole function is gone.
- `_mark_cache_valid` / `prevent_*_cache_reuse` / `init_bus_status` lose their `if(ac) x_ac_ else x_dc_` bodies.
- `BaseBatchSolverSynch`'s eleven loose members become its own two caches; the read-back above, and its three copies, are gone.

It's also the extension point: remote / shared voltage-control layouts, HVDC droop data — add a field and it inherits the lifetime, invalidation and consistency rules of everything already there.

`DualAlgoControl` deliberately stays put. It's threaded through every element-container mutator signature and is already correctly per-family.

## Behaviour changes (all in the foreign-build path)

- A build into a caller's cache **never reuses**. `solver_control`, the snapshot `init_bus_status()` raises its flags against, and those flags all describe *this* grid and say nothing about someone else's cache.
- It **retires** this grid's cache for that family afterwards. The labelling and the split still have to be published — the NR extensions read them back through `lsgrid_ptr` rather than from what the solver was handed, **including `bus_pq`**, which `fill_voltage_control_solver_data` needs and which the old write-through supplied *by accident*. Publishing the matrix too would mean copying it, so what remains is a view for the extensions, not a cache: snapshot cleared, control raised, next own powerflow rebuilds.
- The grid's own algorithm is no longer reset or reconfigured for a solve it will never perform.

## ⚠️ ABI

**This changes `LSGrid`'s member layout.** The note that sat on `_forced_ref_slack_bus_id` ("declared LAST so existing member offsets are unchanged — ABI-stable for the gpusim2grid cross-module `LSGrid` cast") cannot survive regrouping scattered members. `docs/solver_plugin.rst` already requires the same version of headers at runtime ("different `BaseAlgo` layout ... undefined behaviour"), and 1.0.0 already carries several `[BREAKING]` entries — so this is flagged in the changelog rather than worked around. **gpusim2grid must be rebuilt against these headers.**

## Cost — measured

Callgrind slope, 200→1200 powerflows on the exotic-elements IEEE14 grid (so construction cancels):

| | dev_0.13.2 | this branch | delta |
|---|---|---|---|
| AC powerflow | 839 340 instr/pf | 839 541 | +200 (+0.024 %) |
| DC, warm cache | 28 095 instr/pf | 28 309 | +214 (+0.76 %) |

The DC row is the floor — the cheapest powerflow this library can run.

Roughly two thirds of that is **not this change**: a per-function diff of the two profiles puts **+74/pf in `compute_results_tsc_rxha_no_amps`** and **+60/pf in an `Eigen::Ref` helper**, neither of which the diff touches — inlining decisions that moved when the translation unit was recompiled. Wall clock shows nothing: run-to-run spread on one unchanged binary (2464 → 2744 ns on the same DC case) exceeds the gap between the two binaries.

## Tests

- **`test_cache_reuse.cpp`, "the structural half … `unset_changes`"** — the existing `[unset_changes]` sections all reach it through `allow_cache_reuse(false)`, i.e. *both* families off, so the family that runs is stopped by `is_usable`'s first line (`if(!allow_reuse)`) and those sections pass with the rest of `is_usable` deleted. Leaving **one** family automatic is what exercises the structural checks. Verified: gutting `is_usable` to `return allow_reuse;` makes it **SIGSEGV**.
- **`test_cache_reuse.cpp`, "a build into someone else's cache …"** — a foreign build fills the caller's cache completely, publishes what the extensions read, and retires the grid's. The `need_reset_solver()` assertion is the one that separates this from `dev_0.13.2`.
- **`test_batch_voltage_control.cpp`** — the source grid solves identically on both sides of a `TimeSeries` and a `ContingencyAnalysis`.

**210 test cases / 5815 assertions** pass under Release, C++14 (`LS2G_CXX_STANDARD=14`), ASan+UBSan, and valgrind (0 errors, no leaks).

The python layer could **not** be exercised locally (no pybind11 / numpy on the machine this ran on) — CircleCI's `test_legacy_*` jobs cover it. No python-visible API changed; `pre_process_solver` was never bound, and every `&LSGrid::` symbol the bindings reference still resolves.

---

---

### DIFF
--- a/src/core/LSGrid.cpp
+++ b/src/core/LSGrid.cpp
@@ -739,7 +739,6 @@ CplxVect LSGrid::ac_pf(const Eigen::Ref<const CplxVect> & Vinit,
     // pre process the data to define a proper jacobian matrix, the proper voltage vector etc.
     bool is_ac = true;
     CplxVect V = pre_process_solver(Vinit,
-                                    ac_cache_,
                                     algo_controler_.ac_algo_controler());
 
     // start the solver
@@ -1267,14 +1266,13 @@ CplxVect LSGrid::check_solution(const Eigen::Ref<const CplxVect> & V_proposed, b
     // default-constructed (empty) size, and `fill_hvdc_droop_solver_data`'s
     // `id_me_to_solver[bus_id]` lookup (bus ids in the hundreds/thousands) then read
     // out of bounds -- a silent, hard-to-reproduce segfault instead of a clean rebuild.
-    // (the cache-consistency guard in _pre_process_solver_impl now catches this case too,
+    // (the cache-consistency guard in _pre_process_own_cache now catches this case too,
     // and the two families' variants of it; this stays as the local, cheaper statement of
     // what check_solution itself may assume)
     if(ac_cache_.may_be_reused() && ac_cache_.is_consistent(substations_.nb_bus())){
         reset_solver.tell_none_changed();
     }
     CplxVect V = pre_process_solver(V_proposed,
-                                    ac_cache_,
                                     reset_solver,
                                     false);  // do NOT snap regulated buses to their target: we are testing V_proposed as-is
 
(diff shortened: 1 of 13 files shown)


src/core/LSGrid.cpp
```diff
@@ -671,41 +672,41 @@ void LSGrid::init_bus(unsigned int n_sub,
...
 {
     if(reset_ac){
-        id_me_to_ac_solver_ = SolverBusIdVect();
-        id_ac_solver_to_me_ = GlobalBusIdVect();
-        slack_bus_id_ac_solver_ = SolverBusIdVect();
-        Ybus_ac_ = Eigen::SparseMatrix<cplx_type>();
+        ac_cache_.id_me_to_solver = SolverBusIdVect();
+        ac_cache_.id_solver_to_me = GlobalBusIdVect();
+        ac_cache_.slack_bus_id_solver = SolverBusIdVect();
+        ac_cache_.mat = Eigen::SparseMatrix<cplx_type>();
```
BDonnot: Why not ac_cache.clear() ?

src/core/LSGrid.cpp
```diff
@@ -671,41 +672,41 @@ void LSGrid::init_bus(unsigned int n_sub,
...
-    dc_algo_needs_rebuild_ = false;
+    ac_cache_.inj = CplxVect();
+    dc_cache_.inj = RealVect();
+    ac_cache_.bus_pv = SolverBusIdVect();
+    ac_cache_.bus_pq = SolverBusIdVect();
+    dc_cache_.bus_pv = SolverBusIdVect();
+    dc_cache_.bus_pq = SolverBusIdVect();
+    dc_cache_.slack_weights = RealVect();
+    ac_cache_.algo_needs_rebuild = false;  // the algorithms themselves are reset below
+    dc_cache_.algo_needs_rebuild = false;
```
BDonnot: Same why not ac_cache.clear() and dc_cache.clear()

src/core/LSGrid.cpp
```diff
@@ -671,41 +672,41 @@ void LSGrid::init_bus(unsigned int n_sub,
...
 
     if(reset_dc){
-        id_me_to_dc_solver_ = SolverBusIdVect();
-        id_dc_solver_to_me_ = GlobalBusIdVect();
-        slack_bus_id_dc_solver_ = SolverBusIdVect();
-        Bbus_dc_ = Eigen::SparseMatrix<real_type>();
+        dc_cache_.id_me_to_solver = SolverBusIdVect();
+        dc_cache_.id_solver_to_me = GlobalBusIdVect();
+        dc_cache_.slack_bus_id_solver = SolverBusIdVect();
+        dc_cache_.mat = Eigen::SparseMatrix<real_type>();
```
BDonnot: dc_cache.clear() ?

src/core/LSGrid.cpp
```diff
@@ -671,41 +672,41 @@ void LSGrid::init_bus(unsigned int n_sub,
...
-    slack_bus_id_ac_me_ = GlobalBusIdVect();  // slack bus id, gridmodel number
-    slack_bus_id_ac_solver_ = SolverBusIdVect();  // slack bus id, solver number
-    slack_bus_id_dc_me_ = GlobalBusIdVect();
-    slack_bus_id_dc_solver_ = SolverBusIdVect();
-    slack_weights_ac_ = RealVect();
+    ac_cache_.slack_bus_id_me = GlobalBusIdVect();  // slack bus id, gridmodel number
+    ac_cache_.slack_bus_id_solver = SolverBusIdVect();  // slack bus id, solver number
+    dc_cache_.slack_bus_id_me = GlobalBusIdVect();
+    dc_cache_.slack_bus_id_solver = SolverBusIdVect();
+    ac_cache_.slack_weights = RealVect();
```
BDonnot: no need if the cache has been cleared above, is it ?

BDonnot (commented):
Ok double check the number of instructions is not caused by a missed -  Eigen::Ref<const XXX> & or const Eigen::Ref<const XXX> &, or non const one.

Plus some other issues in comment

BDonnot:
Pushed `1fd0c7a`, which answers the four `clear()` comments, and re-measured the instruction count to answer the review.

## 1. `LSGrid::reset()` — the four `clear()` comments

You were right on all four, and for the same reason: the body was still the pre-1.0.1 one with the member names substituted. Ten assignments, three blocks, doing by hand exactly what `SolverSideCache::clear()` does. Now:

```cpp
if(reset_ac) ac_cache_.clear();
if(reset_dc) dc_cache_.clear();
```

The `ac_cache_.slack_weights = RealVect();` you flagged at the bottom was the visible symptom — that line and nine like it are gone.

Two things the old body did that were *not* the same as `clear()`, both artifacts of the members being loose rather than choices:

- **The flags cleared half a cache.** `reset_ac` gated the labelling and the matrix; the injections, the pv-pq split and the slack were cleared for **both** families whatever the flags said. A non-reset family was left holding a labelling and a matrix with no slack and no split — a state that fails `is_consistent()`, so it was unreachable rather than harmful, but the flag now means what its name says. `reset` is `protected` and every caller in the tree passes `(true, true, true)`, so nothing observable changes.
- **`algo_needs_rebuild` was cleared unconditionally**, under the comment *"the algorithms themselves are reset below"* — which is only true when `reset_solver` is set. It moved inside that branch. With `reset_solver` false the algorithms keep their internals and that flag is the only thing left that would tell them to rebuild from the cache.

210 test cases / 5830 assertions pass, Release and Debug/C++14 (the latter with the powerflow path's debug assertions live).

## 2. The instruction count — is it a missed `Eigen::Ref`?

**No.** Measured rather than argued, and the answer is that the PR is not slower here at all.

Setup: gcc 13.3, Release, callgrind, exotic-elements IEEE14, slope over 200 → 1200 powerflows so construction cancels. One caveat: **no KLU on this machine** (SuiteSparse absent, so SparseLU), which is why the absolute AC figure differs from the table in the PR body.

| | dev_1.0.1 | `da92ed9` (refactor only) | PR head |
|---|---|---|---|
| AC powerflow | 640 001 instr/pf | 639 974 (**−27**) | 639 482 (**−520**, −0.081 %) |
| DC, warm cache | 27 658 instr/pf | 27 693 (**+35**, +0.13 %) | 27 654 (**−4**, −0.014 %) |

The per-function diff is the part that actually answers your question. Over 1200 powerflows, head vs `dev_1.0.1`:

| function | AC | DC |
|---|---|---|
| `_pre_process_solver_impl` | −59 974 (−50.0/pf) | −58 774 (−49.0/pf) |
| `pre_process_solver` / `pre_process_dc_solver` | −7 200 (−6.0/pf) | −6 000 (−5.0/pf) |
| `ac_pf` / `dc_pf` | −12 000 (−10.0/pf) | −14 400 (−12.0/pf) |
| **every other `Eigen::Ref`-taking function** | **±0** | **±0** |

That last row is the load-bearing one. `column_dfs`, `column_bmod`, `LU_kernel_bmod`, `SparseLULinearSolver::solve`, `NRSystem::_residual`, `_get_amps`, `v_kv_from_vpu`, `v_deg_from_va`, `_get_results_back_to_orig_nodes`, `compute_results_tsc_rxha_no_amps` — **not one instruction moved** in any of them, across all three builds. A dropped `const Eigen::Ref<const X> &` materialises the Ref at the call, which is a heap allocation plus a copy per call; it would have shown up in exactly those functions, and none of them budged.

The only three symbols whose cost changes are the three whose signature the PR deliberately changes, and all three get **cheaper**. The wrappers in particular are the parameter count: 10 arguments down to 4, so fewer register and stack moves per call.

Confirmed statically as well — the PR adds, removes and changes **zero** `Eigen::Ref` parameters. The only `Eigen::Ref` lines in the diff are the six python-facing accessors (`get_*_pv_solver_numpy`, `get_*_slack_weights_solver`), whose signatures are byte-identical on both sides; only the member they name changed. The non-const `Eigen::Ref<X>`-by-value ones (`fillSbus_me`, `compute_flows`, `fill_SBus_real`, …) are untouched by the diff. In the other direction the PR *removes* a copy: `const RealVect gm_bus_sw = _grid_model.get_ac_slack_weights_solver()` in `BaseBatchSolverSynch` — that was materialising the whole vector out of the `Eigen::Ref` on every batch prep, and it is gone with the read-back.

**One thing to flag:** the "Cost — measured" table in the PR description (+200 AC / +214 DC) does not reproduce here, and it predates `2dac026` ("stop copying the bus-status vector"), which is worth ~−47/pf on `_pre_process_solver_impl` on its own. The refactor commit in isolation is −27/pf AC and +35/pf DC on this box. Also of note: `compute_results_tsc_rxha_no_amps`, which the description blames for +74/pf, is bit-identical across all three builds here — which supports its claim that that was an inlining artifact of that particular build rather than a real cost, but does mean the description's table is now stale for the head commit. Say the word and I'll rewrite that section with these numbers.

---

BDonnot (commented):
Ok

PR #188, opened 2026-08-30
Bus connectivity IS the element counts, and nothing else

BDonnot:
Rebased onto `dev_1.0.1` after #185 merged, so this is 5 commits on top of it — no stack.

## The problem

`std::vector<bool> bus_status_` said the same thing as "does any element sit on this bus", a second time. Two statements of one fact can disagree, and this pair did:

```cpp
// LSGrid::init_bus_status(), before
substations_.disconnect_all_buses();
powerlines_.reconnect_connected_buses(substations_);
shunts_.reconnect_connected_buses(substations_);
trafos_.reconnect_connected_buses(substations_);
generators_.reconnect_connected_buses(substations_);
loads_.reconnect_connected_buses(substations_);
sgens_.reconnect_connected_buses(substations_);
storages_.reconnect_connected_buses(substations_);
hvdc_lines_.reconnect_connected_buses(substations_);
// svcs_ is not on this list
```

`SvcContainer` is a `OneSideContainer_PQ` and inherits a perfectly good `reconnect_connected_buses`. It was simply left off. **A bus whose only element was an SVC read as disconnected, got no solver id, and silently dropped out of the system it should have been in.** An SVC injects reactive power; its bus belongs there.

That is what a hand-written list does. It was also O(all elements) on every powerflow that rebuilt, it was serialized (so a crafted file could make it contradict the elements it claims to describe), and it could be set by hand — `deactivate_bus` — to something the elements contradicted.

## The fix

`SubstationContainer` keeps, per bus, **how many elements hold it alive**. A bus is in the solved system iff that count is non-zero, so the two transitions that matter (`0 -> 1`, `1 -> 0`) are exactly the ones that change the *dimension* of the system and renumber every bus after them. Those, and only those, raise `tell_dimension_changed`.

| | before | after |
|---|---|---|
| `is_bus_connected(b)` | `bus_status_[b]` | `nb_elements_per_bus_[b] > 0` |
| `nb_connected_bus()` | walk the whole vector | one integer, moved by one at the crossings |
| `get_bus_status()` | `const std::vector<bool>&` into a member | built from the counts, **by value** |
| `init_bus_status()` | disconnect all + walk 8 containers | nothing, in the steady state |
| dimension-change detection | two O(nb_bus) walks comparing against a per-family photograph | the crossing raises the flag as it happens, O(1) |

**Which buses an element holds is stated once**, in `GenericContainer::contribute_to_buses`. The recount, and every mutator's `-1 … mutate … +1` bracket, are built from that one predicate — so the incremental and the rebuilt answer *cannot* disagree about a rule. The rule is genuinely per container, which is why it is virtual:

- one-sided elements hold their bus iff active;
- a line / trafo is gated by `status_global_` **first**, and only then does each side count — a line end cannot know that, so it cannot live in `OneSideContainer`;
- an HVDC line has no global gate (one station on while the other is off is legitimate) and simply delegates to its two sides. **Behaviour left exactly as it was**, per review.

Only `tell_dimension_changed` moves here. `ybus_change_sparsity_pattern` / `ybus_values_changed` stay where the shunt / trafo / line containers decide them.

## Deleted rather than kept

- **`reconnect_connected_buses`, all five containers.** Dead once the counts drive the status, and it was the second statement of the rule — the one that drifted.
- **`OneSideContainer`'s "a connected element's bus must be active"** check in `check_grid`. An active element on bus `b` *is* an element holding `b`, so `counts[b] > 0`: a tautology. It could only ever fire before the counts were established, which says nothing about the grid.
- **`GenericContainer::_generic_deactivate/_generic_reactivate(GlobalBusId, SubstationContainer&)`** — no caller left.

## Deprecated no-ops

`LSGrid.deactivate_bus` / `reactivate_bus` (and `SubstationContainer`'s `disconnect_all_buses` / `reconnect_bus` / `disconnect_bus` / `reset_bus_status`) have nothing left to set. Kept so existing code keeps importing — the pandapower and powermodels loaders and `LightSimBackend` all call them.

**This changes no powerflow result**, because they were already all but inert: whatever they wrote, the next powerflow rebuilt the status from the elements and threw it away. A bus with elements on it came straight back; a bus without them was already out. What actually took an out-of-service bus out of the system was the loaders disconnecting its **elements**, and that still works.

## Recovery, so drift cannot outlive an invalidation

Incremental bookkeeping rots silently — one missed decrement and the count is wrong for the rest of the grid's life. So the counts are recounted from the elements whenever there is nothing to protect: a freshly built grid, `set_state` / `load_binary`, and any `init_*` that replaces a whole element container (which no `+1 / -1` can see).

## ⚠️ Binary format 4 -> 6

Nothing about bus connectivity is serialized any more: `LSGrid`'s state loses the AC family's bus-connectivity photograph (5), `SubstationContainer`'s loses `bus_status_` (6). A restored grid counts its buses from its elements, whose own status **is** in the file — so the connectivity round-trips exactly and a file cannot state anything else.

`test_check_grid`'s "rejects a `bus_status` shorter than the bus count" goes with the field: it poisoned a length a file can no longer state. Replaced from the other side by "a restored grid counts its own buses", which changes the topology, round-trips, and checks the restored grid agrees bus for bus.

## Tests

`src/tests/test_bus_element_count.cpp` is new. **All 41 mutators** that can change bus membership, each on a cold grid *and* on one that has already solved, against the strong assertion: after ANY mutation, the incremental counts equal what a recount from the elements produces. Plus the no-op cases (moving an element to the bus it is already on; deactivating an already-deactivated element) — grid2op sends a topology vector every step, most of which asks for the bus an element is already on, and a transient `1 -> 0 -> 1` would read as "the dimension changed" and cost a full rebuild.

`connected_bus_count_is_exact()` states the one remaining invariant and is checked, not asserted in a comment: `init_bus_status()` asserts it, and the sweep checks it after every mutator, **before** the recount that would re-establish it and hide drift.

A guard on the list itself: `CHECK(all_bus_mutations().size() == 41u)`, so a new mutating method that is not added there stops the build's conscience rather than silently escaping coverage.

## What it is worth

Callgrind, one load moved between two buses per solve (so `init_bus_status()` actually runs), 21 solves each:

| grid (sub × busbar) | nb_bus | before | after | Δ | per bus |
|---|---|---|---|---|---|
| 1000 × 1 | 1 000 | 16 378 616 | 16 365 863 | −0.08% | −12.8 |
| 1000 × 12 | 12 000 | 16 990 176 | 16 832 723 | −0.93% | −13.1 |
| 5000 × 1 | 5 000 | 121 603 588 | 121 539 360 | −0.05% | −12.8 |
| 5000 × 12 | 60 000 | 124 767 349 | 123 979 613 | −0.63% | −13.1 |

Divided by `nb_bus`: **12.8, 13.1, 12.8, 13.1** — one flat constant, the same at every size and every element count. The cost removed was proportional to the number of buses, not to the work the solve does, so it is the sparsely-filled grids that gain: 5000 substations at 12 busbars each is a 60 000-entry vector rebuilt to solve a 5 000-bus system.

In wall clock it does not show at any of the four sizes; the solve dominates and run-to-run spread is a few percent. **The case for this change is the shape, and the SVC bug it makes unspeakable.** The instructions are a bonus, not the argument.

## Verification

**215 test cases / 7327 assertions** pass under Release, C++14 (`LS2G_CXX_STANDARD=14`), Debug (assertions live), ASan+UBSan, and valgrind (0 errors, no leaks).

The python layer could **not** be exercised here — no pandapower / pypowsybl / grid2op on this machine. CI covers it.

## ⚠️ One thing this PR cannot fix

`lightsim2grid/tests/binary_format_fixture/case14_sandbox_format4.lsb` is a **format-4** file and was not regenerated at the 4 -> 5 bump, so `TestBinaryLayoutUnchanged` fails independently of this work. Regenerating needs grid2op:

```
python -m lightsim2grid.tests.test_binary_serialization regen
```

then rename the file and `FIXTURE_PATH` to `..._format6.lsb`.

---

---


src/core/LSGrid.hpp
```diff
@@ -773,7 +862,7 @@ class LS2G_API LSGrid final
          * 
          * @return int 
          */
-        int nb_connected_bus() const {return substations_.nb_connected_bus();}
+        int nb_connected_bus() const {_ensure_bus_counts(); return substations_.nb_connected_bus();}
```
BDonnot, line 865: I am not a fan here. Except for the first call before a powerflow, we know exactly this in O(1) for all successive checks. This is weird to force an O(n_bus) each time

src/core/LSGrid.hpp
```diff
@@ -788,7 +877,9 @@ class LS2G_API LSGrid final
         [[nodiscard]] const SGenContainer & get_static_generators() const {return sgens_;}
         [[nodiscard]] const SvcContainer & get_svcs() const {return svcs_;}
         [[nodiscard]] const ShuntContainer & get_shunts() const {return shunts_;}
-        [[nodiscard]] const std::vector<bool> & get_bus_status() const {return substations_.get_bus_status();}
+        /// which buses are in the solved system; built from the element counts, see
+        /// SubstationContainer::get_bus_status (returns BY VALUE, no longer a reference)
+        [[nodiscard]] std::vector<bool> get_bus_status() const {_ensure_bus_counts(); return substations_.get_bus_status();}
```
BDonnot, line 882: Same comment as for nb_connected_buses, should be O(1) once init

src/core/SolverSideCache.hpp
```diff
@@ -129,8 +137,8 @@ struct SolverBusLayout
         if(slack_weights.size() != nb_solver) return false;
         // every bus is pv, pq, or slack: the split can never outnumber the system
         if(bus_pv.size() + bus_pq.size() > static_cast<std::size_t>(nb_solver)) return false;
-        // a snapshot that does not cover the grid means "retired", see last_bus_status
-        if(last_bus_status.size() != nb_bus_grid) return false;
+        // built for another grid, or retired: see built_for_nb_bus
+        if(built_for_nb_bus != nb_bus_grid) return false;
         return true;
     }
 };
```
BDonnot, line 144: I thought the solver sode cache would keep a "hash", a size_t, increased by 1 each time there is a "0->1" or "1->0" (and reset to 0 each time the cache is manually reset) and this "hash" would identify uniquely the topology. Either it's the same as the grid last seen and in this case nothing needs to be done, or this is not (something has been updated) and in the case -> rebuild

src/core/element_container/OneSideContainer.hpp
```diff
@@ -297,9 +343,132 @@ class OneSideContainer : public GenericContainer
...
+                if(subid_.size() != nb()){
+                    std::ostringstream exc_;
+                    exc_ << "OneSideContainer::update_topo: cannot reconnect element " << el_id
+                         << " to a bus: subid_ has " << subid_.size() << " entries but this "
+                         << "container currently has " << nb() << " elements (set_subid was "
+                         << "called for a different element count -- call it again after "
+                         << "re-initializing this container).";
+                    throw std::runtime_error(exc_.str());
+                }
+                int sub_id = subid_(el_id);
```
BDonnot: const int maybe

src/core/element_container/OneSideContainer.hpp
```diff
@@ -314,98 +483,18 @@ class OneSideContainer : public GenericContainer
...
-                }
+                // an entry the caller did not touch mutates nothing, so it must not be
+                // bracketed either: taking a contribution away and putting it straight
+                // back leaves the counts right, but a bus held by this element alone
+                // would transiently hit 0 and report a crossing that never happened.
+                if(!has_changed(checked_pos_topo_vect(el_id, nb_topo))) continue;
+                // ONE bracket for the whole entry: reactivating an element and then
+                // moving it is a single change of which bus it holds, not two.
+                _apply_and_track_buses(el_id, substations, solver_control, [&]{
+                    res[el_id] = update_topo_one_el_no_bus_tracking(el_id, has_changed,
```
BDonnot, line 494: Why no_bus_tracking here ?

src/core/help_fun_msg.cpp
```diff
@@ -4579,6 +4579,11 @@ const std::string DocLSGrid::total_bus = R"mydelimiter(
 
 const std::string DocLSGrid::nb_connected_bus = R"mydelimiter(
     Returns (>0 integer) the number of connected buses on the powergrid (ignores the disconnected bus).
+
+    .. versionchanged:: 1.0.0
```
BDonnot: It's 1.0.1 now

src/core/help_fun_msg.cpp
```diff
@@ -5514,14 +5519,47 @@ const std::string DocLSGrid::get_bus_vn_kv = R"mydelimiter(
 
 )mydelimiter";
 
+const std::string DocLSGrid::deactivate_bus = R"mydelimiter(
+    .. deprecated:: 1.0.0
```
BDonnot: It's 1.0.1

src/core/help_fun_msg.cpp
```diff
@@ -5514,14 +5519,47 @@ const std::string DocLSGrid::get_bus_vn_kv = R"mydelimiter(
...
+    back and a bus without them was already out. Removing the effect therefore changes no
+    powerflow result.
+
+    To take a bus out of the powerflow, disconnect the **elements** on it
+    (``deactivate_load``, ``deactivate_gen``, ``deactivate_powerline``, ...).
+
+)mydelimiter";
+
+const std::string DocLSGrid::reactivate_bus = R"mydelimiter(
+    .. deprecated:: 1.0.0
```
BDonnot: 1.0.1

src/core/help_fun_msg.cpp
```diff
@@ -5514,14 +5519,47 @@ const std::string DocLSGrid::get_bus_vn_kv = R"mydelimiter(
...
 const std::string DocLSGrid::get_bus_status = R"mydelimiter(
     Whether each bus ("gridmodel" numbering) is currently connected -- part of at least one
-    active element or busbar coupling, so contributing an unknown to the powerflow.
+    active element, so contributing an unknown to the powerflow.
 
     There is no dedicated python class for a single bus (unlike loads, generators, etc.): this
     raw per-bus vector, together with :func:`get_bus_vn_kv`, is the only way to inspect bus-level
     state directly.
 
+    .. versionchanged:: 1.0.0
```
BDonnot: 1.0.1

src/tests/test_bus_element_count.cpp
```diff
@@ -0,0 +1,534 @@
...
+        {"change_bus1_powerline",       [](LSGrid & g){ g.change_bus1_powerline_python(7, 2); }},
+        {"change_bus2_powerline",       [](LSGrid & g){ g.change_bus2_powerline_python(8, 9); }},
+        {"deactivate_trafo",            [](LSGrid & g){ g.deactivate_trafo(1); }},
+        {"reactivate_trafo",            [](LSGrid & g){ g.deactivate_trafo(1); g.reactivate_trafo(1); }},
+        {"deactivate_trafo_side1",      [](LSGrid & g){ g.deactivate_trafo_side1(0); }},
+        {"deactivate_trafo_side2",      [](LSGrid & g){ g.deactivate_trafo_side2(0); }},
+        {"reactivate_trafo_side1",      [](LSGrid & g){ g.deactivate_trafo_side1(0); g.reactivate_trafo_side1(0); }},
+        {"reactivate_trafo_side2",      [](LSGrid & g){ g.deactivate_trafo_side2(0); g.reactivate_trafo_side2(0); }},
+        {"change_bus1_trafo",           [](LSGrid & g){ g.change_bus1_trafo_python(0, 3); }},
+        {"change_bus2_trafo",           [](LSGrid & g){ g.change_bus2_trafo_python(0, 3); }},
```
BDonnot, line 121: This should also be tested with the sync_status_both_side non default.

The change_bus should also be tested with -1 (and not only for 2 sides elements but for all)

src/tests/test_bus_element_count.cpp
```diff
@@ -0,0 +1,534 @@
...
+}
+
+}  // namespace
+
+TEST_CASE("update_topo counts the buses it takes out of the system",
+          "[SubstationContainer][bus_count]")
+{
+    // The batch entry point grid2op drives every single step, and the one the
+    // mutator sweep above cannot reach. It is not a thin wrapper over the mutators:
+    // it calls `resolve_status`, which sets a branch's `status_global_` -- the gate
```
BDonnot, line 415: This depends on synch_status_both side, so these tests should be done this flag at non default

BDonnot (commented):
Things to change I think

BDonnot (commented):
Ok

PR #189, opened 2026-09-04
Powerflow performance pass, and a cache that cannot outlive a failure

BDonnot:
A profiling-driven pass over the C++ powerflow path, then the correctness work that came out of it. Every performance item was measured before it was kept: a git worktree baseline at the previous commit, matched build configs, callgrind instruction counts (deterministic) plus a dedicated wall-clock instrument, and full-precision result dumps compared bit-for-bit. Several candidates measured *worse* and were reverted rather than committed — they are not in this branch.

Rebased on top of #188, which this branch merges (`2dfaa18`).

## Performance

Measured on `case9241pegase` unless stated, NR + KLU, callgrind `Ir`.

| change | effect |
|---|---|
| Feed `Contrib` to `setFromTriplets` directly (`4aa6a85`) | drops a whole redundant `cntrb` → `triplets` copy |
| Build J's CSC arrays directly in `build_J_sparsity` (`4c4398d`) | no triplets, no temporary transposed matrix; counting sort with duplicate collapse folded into the transpose |
| Same treatment for the DC slack-free matrix (`08e8b60`) | built in place, two passes over `ref_mat` |
| `fillYbus` via the same assembly | reuses the `setFromTriplets` alternative rather than copy-pasting it |
| Shunt and branch results on real/imag parts (`93088bc`, `77f09e0`) | avoids the NaN-recovery branch `std::complex::operator*` carries after every product; bit-identical for finite inputs |
| Fuse `v_kv_from_vpu` + `v_deg_from_va` (`780baf5`) | one walk instead of two |
| One-pass `_get_amps` with no temporaries (`e8e2e1e`) | `compute_results` −6.7% Ir, −14% wall |
| dS assembly gets both derivatives from one product (`4260235`) | |
| Stop zeroing J, stop rebinding a `Ref` per coefficient (`4fab621`) | |
| FDPF stops rediscovering `(Vm, Va)` with a `hypot` and an `atan2` (`bf2eaf2`) | |
| Bounds-check ids we generated only in debug builds (`3ae9427`, `1bd626d`, `1d68a29`) | see below |

### Internal vs user-facing bounds checks

`_check_in_range` (always active) is now distinguished from `_check_in_range_internal` (`#ifndef NDEBUG`). Only ids the library generated itself are compiled out; **no check reachable from the python API is**. Verified from the binaries rather than by reading: message-string counts in Release vs Debug libraries, plus a release-built probe confirming 6/6 user-facing rejections still throw.

`5a76ac5` also removed the last throw sites from inside the bus-counting bracket, which recovered the full 4.9M instructions per solve that #188's `try`/`catch` had cost — not through virtualization (disassembly showed 1196 vs 1205 instructions, 4 indirect calls each) but through the unwind edge, which changes GCC's scheduling of unrelated inlined code in the same TU. Bisected to that single construct; reverting only the `try`/`catch` restored the old figure to the instruction.

### A powerflow that throws can no longer leave a cache claiming to be in sync

A solve rebuilds its family's cache in place — bus labelling, Ybus/Sbus, the PV/PQ split, slack weights, a factorization. A throw part-way through left a mixture of the old grid and the new one behind, with the change flags still saying whatever they said before the call.

`ac_pf` / `dc_pf` now run the whole solve against a **copy** of the change tracking and hold the grid itself at "everything changed, both families" for the duration; the copy becomes the grid's change tracking only at a publication statement after `compute_results`, which a throw never reaches. Deliberately not a `try`/`catch` — an unwind edge through this code is not free, as above — while the copy is 24 bools: **+21 instructions per solve**, against 206M.

Input validation at the top of `ac_pf` / `dc_pf` (Vinit size, `max_iter`/`tol`, a droop grid handed to a solver that cannot do droop) stays *outside* that bracket: it runs before anything is touched, so a rejected call still costs the caller's cache nothing.

`build_solver_input` / `build_dc_solver_input` got the same treatment from the other side: they retire this grid's cache *before* copying the caller's labelling into it rather than after.

### The per-bus element counts are rebuilt when the cache cannot be reused

Everything a solve builds is derived from the elements, here and now — except the counts, which are carried forward `+1`/`-1` from every mutation the grid has ever seen. `init_bus_status()` only *established* them when they had never been counted; an armed-but-drifted count was inherited for the life of the grid.

That matters more than a stale cache would, because since #188 connectivity **is** the counts: a phantom bus enlarges the solved system and shifts every bus id after it, and nothing downstream can notice — an off-by-one count reads exactly like a real one.

They are now rebuilt from the elements whenever the answer to "may I reuse what was built for this grid?" is no. Hooked to that, **not** to `redo_all`: the latter also fires on `has_dimension_changed()`, which an ordinary topology change raises on every grid2op step, and an O(all elements) recount per step is precisely the cost these counts exist to avoid.

| path | before | after |
|---|---|---|
| cache reuse (grid2op steady state) | 619,932,638 | 619,932,635 (−3 over three solves) |
| full rebuild | 863,811,268 | 879,669,208 (+5.29M/solve, +1.8%) |

All of the delta is `recompute_bus_element_counts` (15,857,772 of 15,857,940).

Together with the point above, this is what closes the loop on a throw out of a *mutator*: the backend calls `tell_solver_need_reset()` on any exception, so the next solve recounts from the elements and the corruption dies there.

## Tests

New C++ cases, each checked against the parent commit's library to confirm it actually fails there:

- A plugin solver that **converges on its first call and throws on every later one**, run twice on the same grid — nothing is swapped between the two calls, so the state after the second is the throw's doing. Before: the grid still answered `nothing_changed()` on both families. Also: a solver throwing from `compute_pf` (AC and DC), a throw raised *after* the algorithm ran, and the success path still publishing exactly what it consumed.
- A drift staged by hand — no public API can produce one any more — and checked that it dies at the next invalidation: a phantom bus (lost decrement, changes connectivity), a count one too low (lost increment, latent until the bus empties one element early), the DC family, and the negative case that an ordinary reusing solve does *not* pay for the recount.
- New `test_fdpf_algorithm.cpp`, plus additions to `test_powerflow_algorithm.cpp`.

**229 test cases / 590,856 assertions pass in C++17 Release, C++14 Release, and Debug with assertions active.** 16/16 grid × algorithm result dumps are bit-identical across the performance work.

Python: 507 passed / 97 skipped / 461 subtests. The remaining failures and errors in this container are all missing optional dependencies (`grid2op`, `pypowsybl`), not assertion failures.

## Notes for review

- **`LSGrid`'s member layout changes**, so anything casting an `LSGrid` across a module boundary (`gpusim2grid`) must be rebuilt against these headers — which the plugin ABI policy in `docs/solver_plugin.rst` already requires. Nothing changes for python.
- `process_results` gained a parameter (protected; `ac_pf`/`dc_pf` are its only callers) and the `_mark_cache_valid(bool)` shorthand is gone — it marked the *member* control, which is now exactly the bug the working-copy protocol removes.
- Still open, and deliberately not in this branch: `fillYbus` position-caching for the `need_recompute_ybus`-without-pattern-change path (14.1M of `setFromTriplets`), cheapening `get_free_vm_slack_solver_buses` (5.61M, called twice per solve) and `fill_voltage_control_solver_data` (4.10M), and the bus power mismatch computed twice per solve (~2.1M).

https://claude.ai/code/session_013nEocx4vydkgBNQKNzRfwy

---
_Generated by [Claude Code](https://claude.ai/code/session_013nEocx4vydkgBNQKNzRfwy)_


src/core/LSGrid.hpp
```diff
@@ -2305,8 +2307,13 @@ class LS2G_API LSGrid final
...
         // (res = _get_results_back_to_orig_nodes(...)) to total_bus() size, which can
         // differ from the caller's initial (often empty) res.
+        // `solve_control` is the working copy of this family's change tracking that
+        // `ac_pf` / `dc_pf` run the whole solve against (see the "exception safety"
+        // note on ac_pf). It is what gets marked "in sync" here -- never the member,
+        // which stays "everything changed" until the caller publishes this copy back
+        // on the success path.
         void process_results(bool conv, CplxVect & res, const Eigen::Ref<const CplxVect> & Vinit, bool ac,
-                             SolverBusIdVect & id_me_to_solver);
+                             SolverBusIdVect & id_me_to_solver, AlgoControl & solve_control);
```
BDonnot, line 2343: If algo control does not change, mark it const (const everywhere is better)

src/core/element_container/GenericContainer.hpp
```diff
@@ -147,32 +147,24 @@ class LS2G_API GenericContainer : public BaseConstants
...
+            // _check_new_bus_id), so a call the grid refuses never reaches this
+            // bracket and the counts it would have left short are never touched.
+            //
+            // This used to be a try / catch that put the contribution back on the way
+            // out of an exception. It was not free: an unwind edge through this header
+            // made GCC keep every std::vector<bool> access in fillYbus live across it,
+            // for 4.9M instructions per rebuild solve of case9241pegase, in a function
+            // that never calls any of this. An exception from deeper inside a mutation
+            // can still leave the counts short; that is a grid which must be rebuilt
+            // and its caches dropped, not one to carry on solving with.
```
BDonnot: The new defense mechanism should be explained here (something that throws put the grid in a state where everything needs to be computed)

src/core/LSGrid.cpp
```diff
@@ -1411,15 +1451,51 @@ CplxVect LSGrid::_build_into_cache(
...
+    // avoid. A step that only moves elements around keeps the incremental counts,
+    // and `init_bus_status()`'s assertion is what watches them in debug builds.
+    // The two branches are the same operation at two strengths, not two operations:
+    // `init_bus_status()` is `_ensure_bus_counts()` -- recount only if the counts were
+    // never armed -- plus a debug assertion. Calling the recount directly does strictly
+    // more, and the assertion is the one thing it skips, which right afterwards cannot
+    // fail: recompute_bus_element_counts() ends with recount_connected_buses(), and
+    // `connected_bus_count_is_exact()` is that same "count the non-empty buses" loop
+    // compared against what it just wrote.
+    if (force_full_rebuild || solver_control.need_reset_solver()){
```
BDonnot: I'm not a huge fan.

need_reset_solver is not "need_reset_cache". Most of the time the cache is not poisoned so it forces its rebuild for nothing. Maybe the addition of another  flag in solver_control above, like "cache_maybe_poisoned" or something would be clearer and avoid the unnecessary rebuild of a consistent cache when solver needs to be reset.

src/core/powerflow_algorithm/BaseAlgo.cpp
```diff
@@ -172,18 +172,22 @@ RealVect BaseAlgo::_evaluate_Fx(
...
+    // once each through the index arrays, so materialising them would only add
+    // two full-length copies.
     // build and fill the result
     RealVect res(npv + 2*npq);
-    res.segment(0,npv) = real_(pv);
-    res.segment(npv,npq) = real_(pq);
-    res.segment(npv+npq, npq) = imag_(pq);
+    res.segment(0,npv) = mis.real()(pv);
+    res.segment(npv,npq) = mis.real()(pq);
+    res.segment(npv+npq, npq) = mis.imag()(pq);
```
BDonnot, line 192: If this function is called by newton Raphson the current implementation is wrong. It misses for example the "extension" for remote voltage control and SVC at least.
Or maybe these are post processed elsewhere?

src/core/powerflow_algorithm/BaseAlgo.cpp
```diff
@@ -235,19 +239,17 @@ RealVect BaseAlgo::_evaluate_Fx(
...
     // build and fill the result
     RealVect res(npv + 2 * npq + 1); // slack adds one component hence the '+1' also bellow)
-    res(0) = real_(slack_id);  // slack bus is first variable
-    res.segment(1, npv) = real_(pv);
-    res.segment(npv + 1, npq) = real_(pq);
-    res.segment(npv + npq + 1, npq) = imag_(pq);
+    res(0) = std::real(mis(slack_id));  // slack bus is first variable
+    res.segment(1, npv) = mis.real()(pv);
+    res.segment(npv + 1, npq) = mis.real()(pq);
+    res.segment(npv + npq + 1, npq) = mis.imag()(pq);
```
BDonnot, line 254: Same question as above, is this used by newton Raphson?

src/core/powerflow_algorithm/BaseFDPFAlgo.hpp
```diff
@@ -21,6 +21,44 @@ template<class LinearSolver, FDPFMethod XB_BX>
...
+        // unreachable from a converging solve -- a trajectory that overshoots a
+        // magnitude ends up diverging, and a diverged solve clears its state.
+        static void canonicalise_vm_va(Eigen::Ref<RealVect> Vm, Eigen::Ref<RealVect> Va)
+        {
+            // a negative magnitude is the same phasor turned by pi
+            Va.array() = (Vm.array() < my_zero_).select(Va.array() + my_pi, Va.array());
+            Vm = Vm.cwiseAbs();
+            // subtracts an exact zero whenever the angle is already in range,
+            // which is the ordinary case
+            Va.array() -= my_two_ * my_pi * (Va.array() * (my_one_ / (my_two_ * my_pi))).round();
```
BDonnot: Why we do that everytime? Once at the end should be enough now ?

src/core/powerflow_algorithm/BaseFDPFAlgo.hpp
```diff
@@ -21,6 +21,44 @@ template<class LinearSolver, FDPFMethod XB_BX>
...
+        // atan2's signed-zero rules produced for (+-0, +-0); this keeps the
+        // incoming angle. Neither can reach a converged answer, because the
+        // caller's next step divides the mismatch by that zero and fails its
+        // allFinite() check.
+        //
+        // Kept as a named static (rather than inline in has_converged) so the
+        // identity it rests on can be tested directly: the branches below are
+        // unreachable from a converging solve -- a trajectory that overshoots a
+        // magnitude ends up diverging, and a diverged solve clears its state.
+        static void canonicalise_vm_va(Eigen::Ref<RealVect> Vm, Eigen::Ref<RealVect> Va)
```
BDonnot: Can this be promoted ro BaseAlgorithm and the same trick be used for ? Newton Raphson too ?

src/core/powerflow_algorithm/BaseFDPFAlgo.hpp
```diff
@@ -99,17 +139,34 @@ class BaseFDPFAlgo final: public BaseAlgo
...
+        void evaluate_mismatch_into(const EigenRefConstCplxSpMat     &  Ybus,
+                                    const Eigen::Ref<const CplxVect> & V,
+                                    const Eigen::Ref<const CplxVect> & Sbus,
+                                    size_t /*slack_id*/,  // id of the ref slack bus
+                                    real_type slack_absorbed,
+                                    const Eigen::Ref<const RealVect> & slack_weights)
+        {
+            ybus_v_.noalias() = Ybus * V;
+            mis_ = V.array() * ybus_v_.array().conjugate() - Sbus.array() + slack_absorbed * slack_weights.array();
+        }
```
BDonnot, line 121: Can this be promoted to BaseAlgorithm too and reused by other solver type (if needed, not sure however it's on the path of other algorithm...)
Anyway having mis_ as a default algorithm member and filled by the concrete implementation sounds reasonable.

src/core/powerflow_algorithm/GaussSeidelAlgo.cpp
```diff
@@ -36,6 +36,9 @@ bool GaussSeidelAlgo::compute_pf(
     // TODO SLACK (for now i put all slacks as PV, except the first one)
     Eigen::VectorXi my_pv = retrieve_pv_with_slack(slack_ids, pv);
 
+    // row-major view + diagonal of Ybus, used by every sweep below
+    refresh_ybus_cache(Ybus);
```
BDonnot, line 61: Should be done only if ybus_values_changed or ybus_changed_sparsity_pattern of dimensions_changed or solver_need_reset, no ?

src/core/powerflow_algorithm/GaussSeidelAlgo.cpp
```diff
@@ -36,6 +36,9 @@ bool GaussSeidelAlgo::compute_pf(
...
+
     V_ = V;
     Vm_ = V_.array().abs();  // update Vm and Va again in case
     Va_ = V_.array().arg();  // we wrapped around with a negative Vm
```
BDonnot, line 65: On fdpf above a faster formula was found for that. And I suggested promoting it to BaseAlgorithm to be reusable. 
This could be reused here I think.

src/core/powerflow_algorithm/GaussSeidelAlgo.cpp
```diff
@@ -81,9 +84,20 @@ bool GaussSeidelAlgo::compute_pf(
...
 
+void GaussSeidelAlgo::refresh_ybus_cache(const EigenRefConstCplxSpMat & Ybus)
+{
+    ybus_rowmajor_ = Ybus;  // column-major -> row-major (one O(nnz) transpose)
+    const Eigen::Index n = Ybus.rows();
+    if(ybus_diag_.size() != n) ybus_diag_.resize(n);
+    ybus_diag_.setZero();
+    for(int col = 0; col < Ybus.outerSize(); ++col)
+        for(EigenRefConstCplxSpMat::InnerIterator it(Ybus, col); it; ++it)
+            if(it.row() == it.col()) ybus_diag_.coeffRef(it.row()) = it.value();
```
BDonnot: No need to continue searching in the inner iterator. A "continue" statement might be good here.

src/core/powerflow_algorithm/NRAlgo.tpp
```diff
@@ -120,14 +120,21 @@ bool NRAlgo<LinearSolver, NRSystem>::compute_pf(
...
-        _system.apply_step(coeff * F);
+        // scale F in place rather than passing the `coeff * F` expression:
+        // binding it to apply_step's Eigen::Ref<const RealVect> parameter made
+        // Eigen materialise the expression into a heap temporary on every
+        // iteration. F is dead here -- it is overwritten by the new mismatch
+        // just below -- so scaling it in place is free, and the full-Newton
+        // case (coeff == 1, the NoScaling policy) does not even pay the scan.
+        if (coeff != static_cast<real_type>(1.)) F *= coeff;
+        _system.apply_step(F);
         timer_Va_Vm_ += timer_va_vm.duration();
```
BDonnot: There might have been an issue at some point, but this timing is not correctly placed.

It should be placed where va and vm are computed not where the vector F is scaled. 
This should be counted toward the timer below (timer mismatch) I believe.
Va and vm should time the conversation from va and vm to v then to va and vm (which should use the formula on fdpf here too I assume)

src/core/powerflow_algorithm/NRSystem.hpp
```diff
@@ -84,11 +85,24 @@ class LS2G_API Contrib final
...
        int ybus_k() const {return ybus_k_;}
        dSdX whichmat() const {return whichmat_;}
 
+       // Eigen's "triplet" protocol: build_J_sparsity hands the contributions
+       // straight to setFromTriplets instead of copying them into a vector of
+       // Eigen::Triplet just to attach the zero. That pass builds the sparsity
+       // pattern only, so every value is zero; fill_J writes the numbers.
+       int row() const {return jrow_;}
+       int col() const {return jcol_;}
+       real_type value() const {return static_cast<real_type>(0.);}
```
BDonnot: Can this be constexpr in cpp14 ?

src/core/powerflow_algorithm/NRSystem.hpp
```diff
@@ -1113,10 +1220,23 @@ class NRSystem<Base, Rest...>
     Eigen::Index                                            Sbus_size_;
     Eigen::Map<const CplxVect> _Sbus_view() const { return Eigen::Map<const CplxVect>(Sbus_data_ptr_, Sbus_size_); }
 
+    // V = Vm * exp(i.Va), written into V_out (resized only when needed).
+    static void _reconstruct_V_into(CplxVect& V_out,
+                                    const Eigen::Ref<const RealVect>& Va,
+                                    const Eigen::Ref<const RealVect>& Vm);
```
BDonnot, line 1258: Shouldn't this be the same as the FDPF implementation, promoted to BaseAlgorithm? Looks like there are 3 or 4 implementations of this now.

src/core/powerflow_algorithm/NRSystem.tpp
```diff
@@ -81,48 +99,141 @@ inline void NRSystem<Base, Rest...>::build_J_sparsity()
...
+        const int p = head[cntrb[e].jrow()]++;
+        csr_col[p] = cntrb[e].jcol();
+        csr_src[p] = e;
+    }
 
-    const int dim_J = ledger_.size();
-    J_.resize(dim_J, dim_J);
-    J_.setFromTriplets(triplets.begin(), triplets.end());
-    J_.makeCompressed();
+    // ---- 2. CSR -> CSC: count the columns, duplicates collapsed -----------
```
BDonnot, line 139: Why going first CSR then CSC. Can't we do CSC directly ?

src/core/powerflow_algorithm/NRSystem.tpp
```diff
@@ -131,33 +242,30 @@ inline void NRSystem<Base, Rest...>::fill_J()
...
+    // them ever claim the same coefficient, see _assign_ds), so they bring
+    // their own coefficients up to date whatever was there before. What still
+    // accumulates is the feature entries, because a component may legitimately
+    // add to a dS coefficient -- an hvdc droop slope adds to the dP/dtheta of
+    // its end buses -- and those need a zero to start from.
+    //
+    // Hence the order: zero the feature positions, let the dS pass overwrite
+    // the ones it shares, then add the feature values on top. Values are
+    // unchanged: `0 + x` is `x`.
+    for (int pos : feature_pos_)
```
BDonnot, line 260: There could even be two types of feature (type constexpr, known at compile time) the one that needs 0ing and the other. For example distributed slack does not need to be zeroed. Once set values does not change I think.

src/core/powerflow_algorithm/NRSystem.tpp
```diff
@@ -227,12 +389,21 @@ inline void NRSystem<Base, Rest...>::fill_internal_variables()
...
+    assert(res.size() == n);
     if(dx_zero_cache_.size() != n) dx_zero_cache_ = RealVect::Zero(n);
-    return _residual(V_, dx_zero_cache_);
+    _residual_into(res, V_, dx_zero_cache_);
+}
+
+template <typename... Rest>
+inline RealVect NRSystem<Base, Rest...>::mismatch() const
+{
+    RealVect res(static_cast<Eigen::Index>(total_state_variables()));
```
BDonnot, line 404: Mismatch could be cached in BaseAlgorithm. Fdpf reuse the same strategy.
The size of the mismatch is not the same here and in fdpf however which is fine. Derived class could be responsible for the mismatch size.
This Would avoid other system calls here I believe.

BDonnot, line 404: Mismatch could be cached in BaseAlgorithm. Fdpf reuse the same strategy.
The size of the mismatch is not the same here and in fdpf however which is fine. Derived class could be responsible for the mismatch size.
This Would avoid other system calls here I believe.

src/core/powerflow_algorithm/NRSystem.tpp
```diff
@@ -260,53 +431,87 @@ inline void NRSystem<Base, Rest...>::apply_step(const Eigen::Ref<const RealVect>
...
+    // same expression as before, but assigned instead of returned, so the
+    // per-call nb_bus complex temporary is gone (Eigen resizes V_out only when
+    // the dimension actually changed, so nothing is allocated after the first
+    // call of a topology).
     const cplx_type m_i = BaseConstants::my_i;
-    return Vm.array() * (Va.array().cos().template cast<cplx_type>()
-                         + m_i * Va.array().sin().template cast<cplx_type>());
+    V_out = Vm.array() * (Va.array().cos().template cast<cplx_type>()
+                          + m_i * Va.array().sin().template cast<cplx_type>());
 }
```
BDonnot, line 474: The fdpf formula might be reused here I believe. And it has been proven faster. Needs to be A/B tested though

src/core/powerflow_algorithm/NRSystem.tpp
```diff
@@ -260,53 +431,87 @@ inline void NRSystem<Base, Rest...>::apply_step(const Eigen::Ref<const RealVect>
...
-                   - _Sbus_view().array();
+    // per-bus complex power mismatch: V .* conj(Ybus V) - Sbus.
+    // Ybus * V_t goes into its own persistent buffer first: left inside the
+    // coefficient-wise expression it is a sparse-times-dense product, which
+    // Eigen can only evaluate into a heap temporary -- one allocation per call,
+    // and this runs at least twice per NR iteration (once per backtracking
+    // trial of a line search).
+    assert(res.size() == static_cast<Eigen::Index>(total_state_variables()));
+    ybus_v_cache_.noalias() = Ybus_ref_ * V_t;
+    mis_cache_ = V_t.array() * ybus_v_cache_.array().conjugate()
```
BDonnot: The more I read this the more I think both real and complex mismatch should be defined in BaseAlgorithm (as members) and filled by the derived classes.
So that it's even usable later on the code (eg for post processing)

BDonnot (commented):
Some more refactoring. 

Do not hesitate to A/B test and not just assume everything here needs to be implemented.

There are some things that change the architecture (like having both mismatch, the real one eg used as rhs for fdpf and nr, and the complex one that is used on every DC Algo at the algorithm level so that each can reuse it (and that it can be reused afterwards).
I would like ro have the cost (or benefit) from this change particularly (and if it does not cost too much to have it implemented)

PR #190, opened 2026-09-05
Audit the cached powerflow path, and fix what it turned up

BDonnot:
An instruction-count audit of the **cached** powerflow path — a solve that follows a successful one, which is what every grid2op step after the first actually runs — plus the fixes it led to.

## The harness (`benchmarks/cache_profiling/`)

A standalone C++ driver (no python, no pybind11, linked straight against `src/core`) that lets callgrind **collect only the `ac_pf` / `dc_pf` calls themselves** — loading, warm-up solves and the mutation between two steps all run with collection off, so every figure is the cost of one powerflow. `run_profile.sh` drives one run per (grid, phase); `summarize.py` parses the raw callgrind output into a per-solve breakdown and a per-call-site ledger.

`ab_test.sh` / `ab_wallclock.sh` / `compare_traces.py` A/B one candidate change: build the tree as it is, build it again with a patch applied, run every (grid, phase) under both, and report instruction counts side by side **and whether the two builds return the same answer** — the driver writes every solve's iteration count and full complex voltage vector at 17 significant digits, so "same answer" is measured, not assumed. Both restore the tree with `git checkout` on exit.

Baseline, case30 / case118 / case1354pegase / case9241pegase: an ordinary cached step costs **177k / 864k / 11.6M / 111.3M** instructions — 2.3–4.0x cheaper than the first solve, 1.7–2.4x cheaper than the same solve with the cache disabled. On the largest grid 93% of that is the algorithm, 55% inside `klu_refactor`.

## Performance (each A/B'd, answers compared)

| change | case30 | case118 | case1354 | case9241 |
|---|---:|---:|---:|---:|
| `1/\|V\|` from `Vm_` instead of a `hypot` pass over `V_` | -2.1% | -2.5% | -2.2% | -1.6% |
| read each branch status bit once, not five times | -1.5% | -1.2% | -1.0% | -0.9% |
| take the per-bus mismatch off the algorithm | -2.1% | -2.1% | — | -1.4% |

The first moves a rounding (identical iteration counts on all 24 pairs; voltages apart by at most 1.4e-12 pu, four to six orders below the tolerance both were accepted at). The second is bookkeeping — **bit-identical, 0 ulp**. The third removes the `Ybus·V` product from `compute_results` entirely, and is primarily a correctness change (below).

## Correctness

- **NaN in published results.** Two elements sharing a bus could publish `NaN` and `-0` for their reactive output: the split divided `(max_q - min_q)` spans that overflow to `+inf` for the `±DBL_MAX` limits a converter writes for "unbounded". The formula was six lines copied into two containers — the split now happens once, in `LSGrid::compute_results`, with an equal-share fallback for non-finite ranges (the same answer it already gave when every range is zero). Finite ranges bit-identical.
- **Double-served elements.** An element whose reactive output the algorithm solved for was served twice on the converter-station side, *and* inflated the totals its neighbours were divided by. Both containers now read one mask, built once per solve from the algorithm's own controller list — one source of truth instead of a rule re-derived differently in each.
- **`init_q_vector` removed.** The per-bus reactive totals were three `nb_bus` vectors accumulated over every generator and station on the grid, on every solve, for a question that concerns only buses carrying two or more locally-regulating machines. The split is now per bus, over a flat list; a bus with one machine computes no total at all. Nothing in the path is proportional to grid size, and no name of that family is left in the tree.
- **Every divergence reported `NotInitError` after 0 iterations.** `process_results` resets a diverged algorithm — it must — and `reset()` means "never been run". The real diagnosis was destroyed one line before any caller could read it, including through `LightSimBackend`'s *"Divergence of AC powerflow. Detailed error: …"*. Error and iteration count are now preserved across the reset.
- **The exotic fixture asked for 180 MW/deg of angle droop on a link rated 20 MW** — it reaches its cap in 0.106° of angle difference, so it is a switch rather than a droop, and moving the slack generator to any of five buses failed with `TooManyIterations`. Now 1.8 MW/deg (full ±20 MW over ~10.6°). Regenerated, not hand-edited; the control was run first — regenerating the *unchanged* fixture reproduces the committed header byte-for-byte on pypowsybl 1.16.1.

## New contract

`BaseAlgo::fills_bus_mismatch()` + `FILLS_BUS_MISMATCH`, alongside the existing capability flags: does this algorithm leave a usable per-bus mismatch behind? True for Newton-Raphson and Fast-Decoupled; Gauss-Seidel now writes it too. **Defaults false**, so a plugin built against an older header is unaffected, and one that claims the capability is checked once per solve. Documented in `docs/solver_plugin.rst`.

## Tests

`src/tests/test_kcl_from_results.cpp` checks Kirchhoff's current law on the **published results** rather than on the solver: it walks the containers, reads the numbers a python user reads back, sums them per bus with each container's own sign convention, and requires zero — no Ybus, no Sbus, no mismatch vector. Calibrated on plain case14 before being pointed at the exotic fixture (the calibration earned its keep — it caught the checker having the HVDC stations backwards).

Suite: **237 test cases, 590,975 assertions, all passing**, nothing tagged as an expected failure.

## Declined findings

Four of the audit's findings were declined by the maintainer (carrying the slack state over between solves, initialising AC from the previous solution rather than DC, J-refactoring policy, voltage-control caching). They are recorded next to their measurements in the README so the numbers stay useful, without reading as unheeded advice. No dead A/B patch is kept in the tree: the one candidate worth landing did so properly, in `LSGrid::compute_results`.

https://claude.ai/code/session_01Q7f2UvXcBDk6EGnqtT6b8R


CHANGELOG.rst
```diff
@@ -139,6 +139,214 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
...
+  of the exotic fixture. It is a bug, not an unsupported arrangement: such a station is a PV
+  source like any generator, two voltage-regulating generators on the slack bus solve in four
+  iterations, and the assembled system has the right shape (20 unknowns, 20 equations, the
+  same as a configuration that converges). Ruled out: the angle droop (one of the four lines
+  has none), a setpoint conflict (making them agree changes nothing), the linear solver (KLU,
+  SparseLU and the single-slack Newton all fail) and the starting point (it diverges
+  warm-started from the exact solution of the unmodified grid). From that exact solution every
+  residual falls to ~1e-7 by the third iteration and then ONE -- the reactive balance at that
+  bus -- grows by a steady factor per iteration until the iterate runs away, which points at a
+  Jacobian entry that does not match its residual at a bus whose voltage has no column.
```
BDonnot: Should be removed now I think

CHANGELOG.rst
```diff
@@ -139,6 +139,214 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
...
+  matter what actually went wrong. ``LSGrid::process_results`` resets the algorithm on
+  divergence -- which it must, to drop a half-converged iterate and a factorization of a
+  system it gave up on -- but ``BaseAlgo::reset()`` puts the algorithm back to "never been
+  run", i.e. ``err_ = NotInitError`` and ``nr_iter_ = 0``, and the divergence path is
+  precisely where a caller reads those back. Every real diagnosis (``SingularMatrix``,
+  ``TooManyIterations``, ``InifiniteValue``, ``SolverFactor``, ...) was destroyed before
+  anyone could see it, including through ``LightSimBackend``'s
+  "Divergence of AC powerflow. Detailed error: ...". The error and the iteration count are
+  now captured and restored across the reset (``BaseAlgo::set_nb_iter``, next to the existing
+  ``set_error``); the next solve overwrites both before using them, so nothing else changes.
```
BDonnot: Already written no ?

CHANGELOG.rst
```diff
@@ -139,6 +139,214 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
...
+  published reactive power **NaN** and the generator's **-0**: the proportional split in
+  ``ConverterStationContainer::set_q`` / ``GeneratorContainer::set_q`` computes
+  ``(max_q_me - min_q_me + eps) / (max_q_bus - min_q_bus + n * eps)``, and the stations of
+  hvdc lines 1 and 2 carry +/- DBL_MAX as their limits (what the pypowsybl converter writes
+  for "unbounded", and the fixture is a capture of a real conversion), so both spans overflow
+  to ``+inf`` and the ratio is ``inf / inf``. The generator, measured against the same
+  infinite denominator, gets ``90 / inf = 0``. Moving the SLACK generator onto that bus does
+  not solve at all (``ErrorType::NotInitError``, zero iterations). A third arrangement -- a
+  generator on the bus a voltage-mode SVC regulates -- cannot be built in v1 (an SVC must be
+  the only controller of its bus), which is recorded as its own test so the gap is explicit.
```
BDonnot: Both fixed now

CHANGELOG.rst
```diff
@@ -139,6 +139,214 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
...
+  ``benchmarks/cache_profiling/README.md`` for the full tables and for six measured savings the
+  audit found, the two largest being (a) the cached path always runs one full Newton iteration
+  even when the grid did not change, because ``MultiSlack::update_state`` re-derives the slack
+  state from ``real(Sbus.sum())`` on every solve instead of carrying it over -- the same solve
+  with ``NRSing_KLU`` converges in **zero** iterations and costs 71% less on case9241pegase --
+  and (b) ``LightSimBackend``'s default ``initdc=True``, which on a warm-started step doubles
+  the cost of a case9241pegase step (221,384,553 vs 111,291,208 instructions) because the DC
+  solution is a worse starting point than the previous AC one. Nothing in the library is
+  changed by this entry: it is an audit, and the candidate fixes were measured on throwaway
+  builds.
```
BDonnot: That entry is written twice no ?

benchmarks/cache_profiling/results_ab_inv_vm.txt
```diff
@@ -0,0 +1,66 @@
...
+============ A/B (wall clock, best of 9) ============
+grid             phase          A (ms)       B (ms)     delta
+case118          idem        0.0356587    0.0350253    -1.78%
+case118          inj          0.072053    0.0697153    -3.24%
+case1354pegase   idem          0.63141     0.613793    -2.79%
+case1354pegase   inj           1.50871      1.46888    -2.64%
+case30           idem       0.00991241   0.00982926    -0.84%
+case30           inj         0.0157342    0.0155277    -1.31%
+case9241pegase   idem          5.70382       5.5814    -2.15%
+case9241pegase   inj           14.5562      14.0722    -3.33%
```
BDonnot: No value in git. Remive

benchmarks/cache_profiling/results_ab_compute_results.txt
```diff
@@ -0,0 +1,50 @@
...
+============ A/B (wall clock, best of 9) ============
+grid             phase          A (ms)       B (ms)     delta
+case118          idem        0.0350448    0.0341631    -2.52%
+case118          inj         0.0699859    0.0692382    -1.07%
+case1354pegase   idem         0.615381     0.610102    -0.86%
+case1354pegase   inj            1.4684      1.46127    -0.49%
+case30           idem       0.00979329   0.00957716    -2.21%
+case30           inj         0.0154853    0.0152443    -1.56%
+case9241pegase   idem           5.5536      5.49528    -1.05%
+case9241pegase   inj           14.1849      14.0967    -0.62%
```
BDonnot: Value un git ???

benchmarks/cache_profiling/results_summary.txt
```diff
@@ -0,0 +1,872 @@
...
+         120,244  LSGrid::_build_into_cache
+         101,680  LSGrid::_get_results_back_to_orig_nodes
+          79,847  LSGrid::fillpv_pq
+          64,718  LSGrid::init_converter_bus_id
+          63,295  LSGrid::fill_voltage_control_solver_data
+          29,008  LSGrid::get_free_vm_slack_solver_buses
+          23,248  LSGrid::get_group_controlled_buses
+          18,506  <file scope>
+             195  LSGrid::compute_results
+             100  LSGrid::ac_pf
```
BDonnot: Added value in git ?

I start to get why this PR is infinite ..  between the double entries in the changelog and now this endless files that are tied to  a specific run for a specific commit and habe likelyhood of at best 0 to be updated each changes...

BDonnot (commented):
Heavy clean up required before I continue

CHANGELOG.rst
BDonnot: Removed. The whole CHANGELOG section was reset to `dev_1.0.1` and rewritten from scratch in 384108b, so nothing marked `[TODO]` or `[KNOWN LIMIT]` for something this branch later fixed survives.

---

CHANGELOG.rst
BDonnot: Yes, it was. Fixed in 384108b — the section is now 11 entries with no repetition.

---

CHANGELOG.rst
BDonnot: Right — both are fixed in this branch, so they are no longer described as open. The rewritten entries state the final state only: the NaN split and the double-served reactive output are plain `[FIXED]` entries now.

---

CHANGELOG.rst
BDonnot: Three times, in fact — the `NotInitError` entry. That is what came of appending to the CHANGELOG as the work went along instead of writing it once at the end. Gone in 384108b.

---

benchmarks/cache_profiling/results_ab_inv_vm.txt
BDonnot: Deleted in 384108b. The A/B result that matters (the `1/|V|` from `Vm_` change) is in the CHANGELOG entry and in the README's table; the raw console dump is reproducible with `ab_test.sh` and does not belong in git.

---

benchmarks/cache_profiling/results_ab_compute_results.txt
BDonnot: None. Deleted in 384108b, along with the other two `results_*.txt` dumps and the `patches/` A/B patch for a change that has since landed properly.

---

benchmarks/cache_profiling/results_summary.txt
BDonnot: Fair, and you are right about the mechanism: those files are a snapshot of one run of one commit, nobody would ever refresh them, and they would be wrong from the next merge onwards. All three are deleted in 384108b (988 lines), together with `patches/`.

I did a pass over the rest of the branch rather than waiting for you to reach it:

- CHANGELOG: the whole section reset to `dev_1.0.1` and rewritten — 11 entries, no duplicate, no `[TODO]`/`[KNOWN LIMIT]` for something fixed later in the same branch.
- `benchmarks/cache_profiling/README.md`: 452 → 144 lines. The per-finding essays, the "what is not worth attacking" section and the running corrections narrative are gone; what is left is what the harness measures, how to run it, and the result tables. It no longer points at the deleted patch.
- `test_kcl_from_results.cpp` and the `inv_vm_cache_` comment in `NRSystem.tpp`: running commentary trimmed.

No functional change in that commit — full suite still green (590,975 assertions, 237 test cases).

---

src/core/LSGrid.cpp
```diff
@@ -2120,21 +2201,164 @@ void LSGrid::compute_results(bool ac){
...
+                        hvdc_lines_.set_station_voltage_control_q(sh.elem_id, 1, q);
+                        break;
+                    default:
+                        hvdc_lines_.set_station_voltage_control_q(sh.elem_id, 2, q);
+                        break;
+                }
+            }
+            first = last;
+        }
+    }
```
BDonnot: This should be refactored in its own function (even multiple ones) and probably not a large block here

BDonnot (commented):
One small fix

src/core/LSGrid.cpp
BDonnot: Done in 3bcab23. `compute_results` went from 285 lines to 61, and the block you flagged is six helpers, each carrying its own reasoning instead of a wall of comment inline:

| helper | what it does |
|---|---|
| `_fill_bus_mismatch_ac` | what the elements at each bus produced — from the algorithm's mismatch, or re-derived when it leaves none |
| `_fill_bus_mismatch_dc` | the same in DC: the imbalance shared among the participating slack buses |
| `_mark_q_solved_by_algo` | who the algorithm solved the reactive output for, from the controller list it hands back |
| `_collect_q_residual_shares` | who instead takes a share of their bus' reactive residual, as a flat list |
| `_split_q_residual_per_bus` | give each of them its share, one bus at a time |
| `_write_back_controller_q` | publish what the algorithm solved for |

`QShare` moves to the header next to the declarations. The "unknown voltage controller kind" throw was written out twice; it is now `_throw_unknown_controller_kind`, which names the function it actually fired in rather than always saying `compute_results`.

Pure code motion, same order of operations: 590,975 assertions in 237 test cases, the same count as on the previous commit.

---

BDonnot (commented):
Ok

PR #191, opened 2026-09-06
Derive the voltage-control plan once per powerflow, in one class

BDonnot:
Follow-up to #188 / #189: the remaining piece of the cross-solve cache, the "fancy" voltage-controller features (remote voltage control, several machines regulating one bus, SVCs).

## What was there

Three derived sets described those controllers, and each was re-derived where it happened to be needed:

| set | derived by |
|---|---|
| which buses a control GROUP regulates (grid ids) | `LSGrid::fillpv_pq` |
| which slack buses keep a free Vm unknown (solver ids) | `Base::update_state` |
| the controller list itself | `VoltageControl::update_state`, which re-ran the free-Vm slack pass of its own first |

Four walks of every generator of the grid per solve, each building `std::set`s as it went — and three independent chances for the layers to disagree, because each walk read the containers again and nothing said they had to agree.

They were never three answers: the controller list is derived from the free-Vm slack set, which is derived from the group layout.

## 1. A class

`src/core/VoltageControlPlan.{hpp,cpp}` — one object, **four** layers, ~250 lines moved off `LSGrid` (`fillpv_pq` included, after review):

| layer | entry point | needs |
|---|---|---|
| 1 — `group_controlled_buses()` | `build_groups` | generators + SVCs only |
| 2 — the pv/pq split | `build_pv_pq` | layer 1 + labelling + slack |
| 3 — `free_vm_slack_buses()` | `build_free_vm_slack` | layers 1–2 |
| 4 — `controllers()` | `build_controllers` | layers 1–3 |

`LSGrid::get_group_controlled_buses` / `get_free_vm_slack_solver_buses` / `fill_voltage_control_solver_data` keep their signatures and behaviour, error messages included; they now build a throw-away plan and are documented as the on-demand form, for callers outside a solve.

## 2. In the cache, once per powerflow

The plan is a member of `SolverBusLayout` — it is expressed in that cache's bus labelling **and** its pv-pq split, which is why it belongs next to them rather than in the extensions that read it. A plan carried across a relabelling is not stale data, it is a different grid.

A foreign build (the batch algorithms) now publishes the plan it just built alongside the labelling, instead of leaving the extensions to re-derive one from it.

## 3. `AlgoControl` + propagation in the element modifiers

Two predicates, both on `AlgoControl` and both read by `_build_into_cache`, so what the powerflow asks and what a test asks cannot drift:

```cpp
bool need_recompute_pv_pq() const noexcept {
    return need_reset_solver_ || change_dimension_ ||
           slack_participate_changed_ || pv_changed_ || pq_changed_;
}
bool need_recompute_voltage_control() const noexcept {
    return need_recompute_pv_pq() || voltage_control_changed_;
}
```

Since the split **is** layer 2 of the plan, "the split must be rebuilt" and "the plan must be rebuilt" are the same question one layer apart: who is in a control group, and which bus it regulates, are exactly the inputs the split reads. So the new `tell_voltage_control_changed()` carries only the one input the split does *not* read — a voltage **setpoint**, which moves no bus's pv/pq class at all — and is raised from exactly two places (`GeneratorContainer::change_v_nothrow`, `ConverterStationContainer::change_v`), on the AC family only and only for an element that actually regulates.

The review also caught three pre-existing over-invalidations, all fixed here: `set_regulated_bus` raising `tell_recompute_sbus()` (`fillSbus` never reads that field) and `tell_pv_changed()` unconditionally, and the converter station's `_deactivate` / `_reactivate` raising `tell_pv_changed()` for non-regulating stations.

## 4. An algorithm that cannot do voltage control now says so

`BaseAlgo::supports_remote_voltage_control()` existed and had no caller. Fast-decoupled and Gauss-Seidel hold no `NRSystem`, so on a grid with control groups they converged **0.36 pu away from the Newton-Raphson answer**, missing setpoints by 0.127 pu — silently. `ac_pf` now refuses such a grid with the full list of offending generators, SVCs and stations; for an algorithm that cannot, no plan is built at all (layer 2 produces the classical split, layers 1/3/4 are skipped), which also stops it paying for work it never reads.

Two smaller fixes found on the way: `change_algorithm(const std::string&)` never called `init_fdpf_coeffs()`, and the profiling driver's `MAX_ITER = 10` diverges for FDPF (~50 needed) — it now derives the budget from `BaseAlgo::is_fdpf()`.

## Measurements

callgrind, KLU, `-O3`, `tol = 1e-8`; every solve's iteration count and full complex voltage vector compared between the two builds — **all traces bit-identical, on every phase**.

| grid | `inj` (an ordinary step) | `idem` (the floor) | `topo` (a line toggled) |
|---|---:|---:|---:|
| case30 | **-4.9%** | -7.9% | -2.3% |
| case118 | **-8.3%** | -16.9% | -4.9% |
| case118_fancy | **-7.0%** | -18.6% | -4.0% |
| case1354pegase | **-3.2%** | -7.3% | -1.9% |
| case1354pegase_fancy | **-2.6%** | -3.9% | -1.7% |
| case9241pegase | **-2.2%** | -5.4% | -1.4% |
| case9241pegase_fancy | **-0.4%** | -0.8% | -0.5% |

`idem` and `inj` save the *same* number of instructions to the last one: what is removed is a fixed per-solve cost, before the Newton loop starts. `inj` reads smaller only because it is a 1.5x–2.7x bigger solve. About 1.9k of the figure is an unrelated find: `AlgorithmSelector` took its `error_msg` by `const std::string &`, and every call site passes a literal longer than libstdc++'s SSO buffer — a malloc and a free per solve for a string only the error path reads.

Part of the `topo` column is the term the review disputed. `need_recompute_pv_pq()` listed `ybus_change_sparsity_pattern_`; the reviewer disagreed and asked for a test rather than an argument, so the `[pv_pq]` cases in `test_cache_reuse.cpp` reach a state where one term is raised and the other five are not, solve warm, reset, solve cold and compare — and the predicate was then built both ways:

```
dropping ybus_change_sparsity_pattern_  ->  41 assertions, all pass   -> dropped
dropping slack_participate_changed_     ->  1 of 2 cases fails        -> kept
```

Worth **-0.9% / -1.8% / -0.7% / -0.4%** on its own (case30 / case118 / case1354pegase / case9241pegase, `topo`, same tree with the term restored).

`make_grids.py` also writes `_fancy` variants of case118 / case1354pegase / case9241pegase — pairs of generators re-pointed at a common neighbouring load bus plus voltage-mode SVCs. The plain pandapower cases have no remote control and no SVC at all, so nothing about the bordered block showed up in the #190 audit.

## Two findings, neither in scope

- **On case9241pegase_fancy, 86% of a cached solve is `klu_refactor`** (55% on the plain case). Forty two-member groups add 80 reactive-injection columns, each coupling a generator's bus to a regulated bus a branch away, and KLU's ordering pays the fill-in.
- The singular configurations hit while writing the tests (a local PV regulator co-existing with a remote group) reproduce **identically on the pre-change binary**. That is the open TODO entry, not a regression.

## Tests

- C++: **249 test cases / 591,174 assertions green**, including the new `src/tests/test_voltage_control_plan_cache.cpp` (every input of the plan mutated on a grid that has already solved, so a stale plan is available to be wrongly reused, *and* on a fresh grid built in the final state — a reused stale plan does not throw and does not look wrong, it converges on the previous scenario's controllers) and the new `[pv_pq]` section of `test_cache_reuse.cpp`.
- Python: 188 passed / 65 skipped across the tests touching this area. The full suite's remaining failures are all missing `grid2op` `data_test` fixtures in this container and reproduce without the change.

https://claude.ai/code/session_01Qz1LmyVeo1nnhrJvLJv9iY

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qz1LmyVeo1nnhrJvLJv9iY)_

### DIFF
--- a/src/core/Utils.hpp
+++ b/src/core/Utils.hpp
@@ -267,16 +267,29 @@ class AlgoControl final
          *
          * Every term is a reason the split itself changes: the system was rebuilt from
          * scratch (`need_reset_solver_`, which `tell_cache_maybe_poisoned()` implies),
-         * the bus set changed (`change_dimension_`), the bus LABELLING changed
-         * (`ybus_change_sparsity_pattern_`, which is what makes
-         * `LSGrid::init_converter_bus_id` run), the slack set moved
+         * the bus set changed (`change_dimension_`), the slack set moved
          * (`slack_participate_changed_`, and the slack is not PV), or a bus changed
          * class (`pv_changed_` / `pq_changed_`).
+         *
+         * `ybus_change_sparsity_pattern_` is deliberately NOT a term, though it looks
+         * like one: it is the flag for "the bus LABELLING may have moved". It is only
+         * ever raised by a BRANCH-side mutation (reconnecting a line or a trafo,
+         * moving one of its ends), and every one of those goes through
+         * `GenericContainer::_apply_and_track_buses`, which raises `change_dimension_`
+         * exactly when such a mutation empties or fills a bus -- which is exactly when
+         * the labelling moves. If no bus crossed, `id_me_to_solver` is unchanged, and a
+         * branch is neither a voltage controller nor a slack, so the split it produces
+         * is identical. The term was there, and dropping it was measured, not argued:
+         * see the `[pv_pq]` cases in `test_cache_reuse.cpp`, which put the grid in a
+         * state where this flag is the ONLY term raised and check the reused split
+         * against a cold one. With the term dropped they still pass; the same
+         * experiment run on `slack_participate_changed_` fails, which is why that one
+         * stays.
          *
          * Read by `LSGrid::_build_into_cache`, which is the only thing that rebuilds it.
          */
         [[nodiscard]] bool need_recompute_pv_pq() const noexcept {
-            return need_reset_solver_ || change_dimension_ || ybus_change_sparsity_pattern_ ||
+            return need_reset_solver_ || change_dimension_ ||
                    slack_participate_changed_ || pv_changed_ || pq_changed_;
         }
 
--- a/src/core/Utils.hpp
+++ b/src/core/Utils.hpp
@@ -193,23 +193,21 @@ class AlgoControl final
         void tell_ybus_some_coeffs_zero(){ybus_some_coeffs_zero_ = true;}
         void tell_one_el_changed_bus(){one_el_change_bus_ = true;}
         /**
-         * Something the voltage-control plan is made of has changed: which buses a
-         * control GROUP regulates, who is in a group, or a setpoint / sharing key a
-         * group's bordered rows carry.
+         * A voltage SETPOINT a control group's bordered rows carry has moved: a
+         * regulating generator's or hvdc converter station's target magnitude.
          *
-         * Raised by the element modifiers that can move any of those -- a generator's
-         * regulated bus, its voltage setpoint, its regulator being connected or
-         * disconnected, an SVC or an hvdc converter station changing bus or status.
-         * Deliberately its OWN flag rather than a reuse of `tell_pv_changed()`: the
-         * two questions really do differ. A remote-regulating generator changing its
-         * setpoint changes no bus's pv/pq class at all (that is the whole point of the
-         * bordered formulation: the regulated bus stays PQ), and conversely most of
-         * what makes a bus PV or PQ has nothing to do with a control group. Folding
-         * one into the other would either rebuild the pv-pq split for nothing or --
-         * the dangerous direction -- reuse a plan whose setpoints have moved.
+         * Deliberately narrow, and deliberately NOT raised for everything the
+         * voltage-control plan is made of. Who is in a group, and which bus a group
+         * regulates, are the same inputs the pv/pq split reads -- and since that split
+         * IS a layer of the plan (VoltageControlPlan::build_pv_pq), a change to one is
+         * a change to the other, already carried by `pv_changed_` and friends. See
+         * `need_recompute_pv_pq()`. What those flags do NOT carry is a setpoint: moving
+         * a remote regulator's target changes no bus' pv/pq class at all -- that is the
+         * point of the bordered formulation, the regulated bus stays PQ -- so it needs
+         * a flag of its own, and this is it.
          *
-         * See `need_recompute_voltage_control()` for the question a powerflow
-         * actually asks, and VoltageControlPlan for what gets rebuilt.
+         * If you add an input to the plan that the pv/pq split does not read, raise
+         * this from the modifier that moves it.
          */
         void tell_voltage_control_changed(){voltage_control_changed_ = true;}
         /**
@@ -265,30 +263,48 @@ class AlgoControl final
         bool has_voltage_control_changed() const {return voltage_control_changed_;}
 
         /**
+         * Must the pv/pq split be rebuilt?
+         *
+         * Every term is a reason the split itself changes: the system was rebuilt from
+         * scratch (`need_reset_solver_`, which `tell_cache_maybe_poisoned()` implies),
+         * the bus set changed (`change_dimension_`), the bus LABELLING changed
+         * (`ybus_change_sparsity_pattern_`, which is what makes
+         * `LSGrid::init_converter_bus_id` run), the slack set moved
+         * (`slack_participate_changed_`, and the slack is not PV), or a bus changed
+         * class (`pv_changed_` / `pq_changed_`).
+         *
+         * Read by `LSGrid::_build_into_cache`, which is the only thing that rebuilds it.
+         */
+        [[nodiscard]] bool need_recompute_pv_pq() const noexcept {
+            return need_reset_solver_ || change_dimension_ || ybus_change_sparsity_pattern_ ||
+                   slack_participate_changed_ || pv_changed_ || pq_changed_;
+        }
+
+        /**
          * Must the voltage-control plan be rebuilt, or does the one the previous solve
          * of this family left in its cache still describe the grid?
          *
-         * A plan is expressed in ONE bus labelling and ONE pv-pq split, so it is
-         * invalidated both by what changes its own inputs (`voltage_control_changed_`)
-         * and by anything that re-labels the buses or re-splits them underneath it --
-         * a plan carried across a relabelling is not stale data, it is a different
-         * grid. Hence the union below rather than the single flag.
+         * The pv/pq split is a LAYER of that plan (VoltageControlPlan::build_pv_pq), so
+         * whatever rebuilds the split rebuilds the plan -- that is the first half, and
+         * it is not a grab-bag of loosely-related flags but literally the same question
+         * asked one layer down. Whoever is in a control group, and which bus it
+         * regulates, are exactly the inputs the split reads; there is no way to change
+         * one without changing the other.
          *
-         * What is deliberately NOT in it is the set of changes an ordinary grid2op
-         * step makes: moving a load's P and Q raises `need_recompute_sbus_` and
+         * The second half is what the split does NOT read: a setpoint. See
+         * `tell_voltage_control_changed()`.
+         *
+         * What is deliberately in NEITHER half is the set of changes an ordinary
+         * grid2op step makes: moving a load's P and Q raises `need_recompute_sbus_` and
          * nothing else, so the plan survives such a step untouched -- which is the
          * whole point of caching it.
          *
-         * All-or-nothing on purpose: the plan's three layers are rebuilt together or
-         * not at all. Refreshing only the setpoints of an otherwise unchanged group
-         * layout (the `v_changed_` case) is a finer question this does not try to
-         * answer yet.
+         * All-or-nothing on purpose: the plan's layers are rebuilt together or not at
+         * all. Refreshing only the setpoints of an otherwise unchanged group layout is
+         * a finer question this does not try to answer yet.
          */
         [[nodiscard]] bool need_recompute_voltage_control() const noexcept {
-            return voltage_control_changed_ || change_dimension_ || need_reset_solver_ ||
-                   pv_changed_ || pq_changed_ || slack_participate_changed_ ||
-                   slack_weight_changed_ || one_el_change_bus_ ||
-                   ybus_change_sparsity_pattern_ || v_changed_ || cache_maybe_poisoned_;
+            return need_recompute_pv_pq() || voltage_control_changed_;
         }
 
     private:    
--- a/src/core/Utils.hpp
+++ b/src/core/Utils.hpp
@@ -108,7 +108,8 @@ class AlgoControl final
             ybus_some_coeffs_zero_(true),
             ybus_change_sparsity_pattern_(true),
             one_el_change_bus_(true),
-            cache_maybe_poisoned_(true)
+            cache_maybe_poisoned_(true),
+            voltage_control_changed_(true)
             {};
 
         ~AlgoControl() noexcept = default;
@@ -127,6 +128,7 @@ class AlgoControl final
             ybus_change_sparsity_pattern_ = true;
             one_el_change_bus_ = true;
             cache_maybe_poisoned_ = true;
+            voltage_control_changed_ = true;
         }
 
         /**
@@ -145,7 +147,7 @@ class AlgoControl final
                    !need_recompute_sbus_ && !need_recompute_ybus_ && !v_changed_ &&
                    !slack_weight_changed_ && !ybus_some_coeffs_zero_ &&
                    !ybus_change_sparsity_pattern_ && !one_el_change_bus_ &&
-                   !cache_maybe_poisoned_;
+                   !cache_maybe_poisoned_ && !voltage_control_changed_;
         }
 
         void tell_none_changed(){
@@ -162,6 +164,7 @@ class AlgoControl final
             ybus_change_sparsity_pattern_ = false;
             one_el_change_bus_ = false;
             cache_maybe_poisoned_ = false;
+            voltage_control_changed_ = false;
         }
 
         // the dimension of the Ybus matrix / Sbus vector has changed (eg. topology changes)
@@ -189,6 +192,26 @@ class AlgoControl final
         // might need to trigger some recomputation of some solvers (eg NR based ones)
         void tell_ybus_some_coeffs_zero(){ybus_some_coeffs_zero_ = true;}
         void tell_one_el_changed_bus(){one_el_change_bus_ = true;}
+        /**
+         * Something the voltage-control plan is made of has changed: which buses a
+         * control GROUP regulates, who is in a group, or a setpoint / sharing key a
+         * group's bordered rows carry.
+         *
+         * Raised by the element modifiers that can move any of those -- a generator's
+         * regulated bus, its voltage setpoint, its regulator being connected or
+         * disconnected, an SVC or an hvdc converter station changing bus or status.
+         * Deliberately its OWN flag rather than a reuse of `tell_pv_changed()`: the
+         * two questions really do differ. A remote-regulating generator changing its
+         * setpoint changes no bus's pv/pq class at all (that is the whole point of the
+         * bordered formulation: the regulated bus stays PQ), and conversely most of
+         * what makes a bus PV or PQ has nothing to do with a control group. Folding
+         * one into the other would either rebuild the pv-pq split for nothing or --
+         * the dangerous direction -- reuse a plan whose setpoints have moved.
+         *
+         * See `need_recompute_voltage_control()` for the question a powerflow
+         * actually asks, and VoltageControlPlan for what gets rebuilt.
+         */
+        void tell_voltage_control_changed(){voltage_control_changed_ = true;}
         /**
          * The per-bus element counts may no longer be what the elements say.
          *
@@ -238,6 +261,35 @@ class AlgoControl final
         bool has_one_el_changed_bus() const {return one_el_change_bus_;}
         // see tell_cache_maybe_poisoned()
         bool cache_maybe_poisoned() const {return cache_maybe_poisoned_;}
+        // see tell_voltage_control_changed()
+        bool has_voltage_control_changed() const {return voltage_control_changed_;}
+
+        /**
+         * Must the voltage-control plan be rebuilt, or does the one the previous solve
+         * of this family left in its cache still describe the grid?
+         *
+         * A plan is expressed in ONE bus labelling and ONE pv-pq split, so it is
+         * invalidated both by what changes its own inputs (`voltage_control_changed_`)
+         * and by anything that re-labels the buses or re-splits them underneath it --
+         * a plan carried across a relabelling is not stale data, it is a different
+         * grid. Hence the union below rather than the single flag.
+         *
+         * What is deliberately NOT in it is the set of changes an ordinary grid2op
+         * step makes: moving a load's P and Q raises `need_recompute_sbus_` and
+         * nothing else, so the plan survives such a step untouched -- which is the
+         * whole point of caching it.
+         *
+         * All-or-nothing on purpose: the plan's three layers are rebuilt together or
+         * not at all. Refreshing only the setpoints of an otherwise unchanged group
+         * layout (the `v_changed_` case) is a finer question this does not try to
+         * answer yet.
+         */
+        [[nodiscard]] bool need_recompute_voltage_control() const noexcept {
+            return voltage_control_changed_ || change_dimension_ || need_reset_solver_ ||
+                   pv_changed_ || pq_changed_ || slack_participate_changed_ ||
+                   slack_weight_changed_ || one_el_change_bus_ ||
+                   ybus_change_sparsity_pattern_ || v_changed_ || cache_maybe_poisoned_;
+        }
 
     private:    
         bool change_dimension_;
@@ -253,6 +305,7 @@ class AlgoControl final
         bool ybus_change_sparsity_pattern_;  // sparsity pattern of ybus changed (and so are its coeff), or ybus change of dimension
         bool one_el_change_bus_;  // whether one element has change of bus (or being reconnected / disconnected)
         bool cache_maybe_poisoned_;  // the per-bus element counts may have drifted: see tell_cache_maybe_poisoned()
+        bool voltage_control_changed_;  // an input of the voltage-control plan moved: see tell_voltage_control_changed()
 };
 
 /**
--- a/src/core/VoltageControlPlan.cpp
+++ b/src/core/VoltageControlPlan.cpp
@@ -13,6 +13,7 @@
 #include <stdexcept>
 
 #include "BaseConstants.hpp"
+#include "element_container/GenericContainer.hpp"
 #include "element_container/GeneratorContainer.hpp"
 #include "element_container/HvdcLineContainer.hpp"
 #include "element_container/SvcContainer.hpp"
@@ -30,9 +31,16 @@ void VoltageControlPlan::clear() noexcept
 // layer 1: which buses a GROUP regulates (grid ids, no labelling needed)
 // ---------------------------------------------------------------------------
 void VoltageControlPlan::build_groups(const GeneratorContainer & generators,
-                                      const SvcContainer & svcs)
+                                      const SvcContainer & svcs,
+                                      bool supports_voltage_control)
 {
     group_reg_buses_.clear();
+    // An algorithm with no bordered block cannot honour a group, and taking a bus
+    // out of PV for one it will not build leaves that bus' magnitude pinned by
+    // nothing at all. Leaving the set empty is what gives layer 2 the classical
+    // split; refusing the grid outright, when it really does have controllers, is
+    // LSGrid::ac_pf's job (see list_unsupported) and happens before this runs.
+    if(!supports_voltage_control) return;
     // an ACTIVE remote-regulating generator: gen_is_voltage_controller() already
     // means "connected, regulator on, not pseudo-off, and regulating a bus that is
     // NOT its own". A purely local regulator therefore never lands here, which is
@@ -54,7 +62,104 @@ void VoltageControlPlan::build_groups(const GeneratorContainer & generators,
 }
 
 // ---------------------------------------------------------------------------
-// layers 2 and 3
+// layer 2: the pv/pq split
+// ---------------------------------------------------------------------------
+void VoltageControlPlan::build_pv_pq(const std::vector<const GenericContainer *> & pv_sources,
+                                     const SolverBusIdVect & id_me_to_solver,
+                                     const GlobalBusIdVect & id_solver_to_me,
+                                     const SolverBusIdVect & slack_bus_id_solver,
+                                     SolverBusIdVect & bus_pv_out,
+                                     SolverBusIdVect & bus_pq_out) const
+{
+    const int nb_bus = static_cast<int>(id_solver_to_me.size());  // number of bus in the solver!
+    std::vector<int> bus_pq;
+    bus_pq.reserve(nb_bus);
+    std::vector<int> bus_pv;
+    bus_pv.reserve(nb_bus);
+    std::vector<bool> has_bus_been_added(nb_bus, false);
+
+    bus_pv_out = SolverBusIdVect();
+    bus_pq_out = SolverBusIdVect();
+
+    // the classical part: every container says which buses it pins
+    for(const GenericContainer * container : pv_sources){
+        if(container == nullptr) continue;
+        container->fillpv(bus_pv, has_bus_been_added, slack_bus_id_solver, id_me_to_solver);
+    }
+
+    // ... and layer 1 takes back the ones a control GROUP regulates: their magnitude
+    // is set by the group's bordered voltage row, so they must keep their own Vm
+    // unknown (and hence their Q equation). See the doc on this function for the
+    // configuration this fixes. Whatever pinned such a bus through the PV path -- a
+    // local generator, or a voltage-regulating hvdc converter station -- is enrolled
+    // as a member of the group by build_controllers instead.
+    if(!group_reg_buses_.empty()){
+        std::vector<int> bus_pv_kept;
+        bus_pv_kept.reserve(bus_pv.size());
+        for(int bus_id_solver : bus_pv){
+            const int bus_id_me = id_solver_to_me[bus_id_solver].cast_int();
+            if(bus_id_me < 0 || !group_reg_buses_.count(bus_id_me)){
+                bus_pv_kept.push_back(bus_id_solver);
+                continue;
+            }
+            has_bus_been_added[bus_id_solver] = false;  // let the PQ loop take it
+        }
+        bus_pv.swap(bus_pv_kept);
+    }
+
+    // TODO remove the order here..., i could be faster in this piece of code
+    // (looping once through the buses)
+    for(int bus_id = 0; bus_id < nb_bus; ++bus_id){
+        if(GenericContainer::is_in_vect(bus_id, slack_bus_id_solver.to_int_vector())) continue;  // slack bus is not PQ either
+        if(has_bus_been_added[bus_id]) continue; // a pv bus cannot be PQ
+        bus_pq.push_back(bus_id);
+        has_bus_been_added[bus_id] = true;  // don't add it a second time
+    }
+    bus_pv_out = SolverBusIdVect(bus_pv.size(), SolverBusId(0));
+    for(int i = 0; i < static_cast<int>(bus_pv.size()); ++i){
+        bus_pv_out(i) = SolverBusId(bus_pv[i]);
+    }
+    bus_pq_out = SolverBusIdVect(bus_pq.size(), SolverBusId(0));
+    for(int i = 0; i < static_cast<int>(bus_pq.size()); ++i){
+        bus_pq_out(i) = SolverBusId(bus_pq[i]);
+    }
+}
+
+// ---------------------------------------------------------------------------
+// the guard: what an algorithm without the bordered block must refuse
+// ---------------------------------------------------------------------------
+VoltageControlPlan::Unsupported
+VoltageControlPlan::list_unsupported(const GeneratorContainer & generators,
+                                     const SvcContainer & svcs,
+                                     const HvdcLineContainer & hvdc_lines) const
+{
+    Unsupported res;
+    if(group_reg_buses_.empty()) return res;  // nothing needs the bordered block
+
+    const int nb_gen = static_cast<int>(generators.nb());
+    for(int gen_id = 0; gen_id < nb_gen; ++gen_id){
+        if(generators.gen_is_voltage_controller(gen_id)) res.gen_ids.push_back(gen_id);
+    }
+    const int nb_svc = static_cast<int>(svcs.nb());
+    for(int svc_id = 0; svc_id < nb_svc; ++svc_id){
+        if(svcs.svc_is_voltage_controller(svc_id)) res.svc_ids.push_back(svc_id);
+    }
+    // a station is never an offender on its own: it pins its own bus the classical
+    // way unless a group already claims that bus, in which case it is enrolled and
+    // the user needs to know it is affected too
+    const int nb_hvdc = static_cast<int>(hvdc_lines.nb());
+    for(int hvdc_id = 0; hvdc_id < nb_hvdc; ++hvdc_id){
+        for(int side = 1; side <= 2; ++side){
+            if(!hvdc_lines.station_is_voltage_controller(hvdc_id, side)) continue;
+            const int bus = hvdc_lines.get_station_bus(hvdc_id, side).cast_int();
+            if(group_reg_buses_.count(bus)) res.station_ids.push_back(std::make_pair(hvdc_id, side));
+        }
+    }
+    return res;
+}
+
+// ---------------------------------------------------------------------------
+// layers 3 and 4
 // ---------------------------------------------------------------------------
 void VoltageControlPlan::build_solver_side(const GeneratorContainer & generators,
                                            const SvcContainer & svcs,
@@ -92,7 +197,7 @@ void VoltageControlPlan::build_free_vm_slack(const GeneratorContainer & generato
     std::set<int> locally_vfixed;
     // ... except that a local regulator on a bus a control GROUP regulates does not
     // pin it: it is enrolled as a member of that group instead (see build_groups and
-    // the reclassification in LSGrid::fillpv_pq), and the group's voltage row needs
+    // the reclassification in build_pv_pq), and the group's voltage row needs
     // the free Vm this grants.
     const int nb_gen = static_cast<int>(generators.nb());
     const GlobalBusIdVect & gen_buses = generators.get_buses();
@@ -125,7 +230,7 @@ void VoltageControlPlan::build_controllers(const GeneratorContainer & generators
     if(nb_bus_solver == 0) return;  // see build_free_vm_slack
 
     // PQ membership: a bus owns a Q equation AND a Vm unknown iff it is a PQ bus
-    // (PV buses have only theta/P, the slack none). `bus_pq` is set by fillpv_pq.
+    // (PV buses have only theta/P, the slack none). `bus_pq` is layer 2's own output.
     std::vector<bool> is_pq(nb_bus_solver, false);
     for(int k = 0; k < static_cast<int>(bus_pq.size()); ++k){
         const int b = bus_pq(k).cast_int();
@@ -201,7 +306,7 @@ void VoltageControlPlan::_collect_gen_controllers(const GeneratorContainer & gen
         // The regulated bus needs a Vm unknown for the bordered row to act on. An
         // ordinary PQ bus has one; so does a slack bus that nothing pins locally (same
         // escape hatch as the controller bus just above -- `has_free_q`). A bus that is
-        // PV despite being group-controlled cannot occur any more (fillpv_pq
+        // PV despite being group-controlled cannot occur any more (layer 2
         // reclassifies it), so what is left here is a bus pinned by something that
         // cannot be enrolled.
         if(!is_pq[reg_solver] && !has_free_q[reg_solver]){
@@ -279,7 +384,7 @@ void VoltageControlPlan::_collect_station_controllers(const HvdcLineContainer & 
     // the voltage-regulating hvdc converter stations. A VSC station with
     // voltage_regulator_on pins its own bus through the PV path, exactly like a local
     // generator, so it is a controller only when a GROUP regulates that bus instead
-    // (fillpv_pq then kept the bus out of PV and it needs its members). Its sharing
+    // (layer 2 then kept the bus out of PV and it needs its members). Its sharing
     // key is a reactive range in MVAr -- the same currency as a generator's -- so a
     // mixed generator/station group shares reactive power correctly, unlike an SVC
     // (whose key is a susceptance range; hence the SVC-alone restriction below).
--- a/src/core/VoltageControlPlan.cpp
+++ b/src/core/VoltageControlPlan.cpp
@@ -1,0 +1,395 @@
+// Copyright (c) 2026, RTE (https://www.rte-france.com)
+// See AUTHORS.txt
+// This Source Code Form is subject to the terms of the Mozilla Public License, version 2.0.
+// If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+// you can obtain one at http://mozilla.org/MPL/2.0/.
+// SPDX-License-Identifier: MPL-2.0
+// This file is part of LightSim2grid, LightSim2grid implements a c++ backend targeting the Grid2Op platform.
+
+#include "VoltageControlPlan.hpp"
+
+#include <cmath>
+#include <sstream>
+#include <stdexcept>
+
+#include "BaseConstants.hpp"
+#include "element_container/GeneratorContainer.hpp"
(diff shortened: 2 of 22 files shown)


src/core/Utils.hpp
```diff
@@ -238,6 +261,35 @@ class AlgoControl final
...
+         * All-or-nothing on purpose: the plan's three layers are rebuilt together or
+         * not at all. Refreshing only the setpoints of an otherwise unchanged group
+         * layout (the `v_changed_` case) is a finer question this does not try to
+         * answer yet.
+         */
+        [[nodiscard]] bool need_recompute_voltage_control() const noexcept {
+            return voltage_control_changed_ || change_dimension_ || need_reset_solver_ ||
+                   pv_changed_ || pq_changed_ || slack_participate_changed_ ||
+                   slack_weight_changed_ || one_el_change_bus_ ||
+                   ybus_change_sparsity_pattern_ || v_changed_ || cache_maybe_poisoned_;
```
BDonnot: Ok for :

Voltage_control_changed_
Change_dimension
Need_reset_solver
cache_maybe_poisoined

Doubt for:
- pq_changed 
- pv_changed


Really convinced with:

- Slack_participate_changed -> what has the slack to do with it ?
- v_changed -> v is pre processed by the LSgrid.

This does definitely not belong here:

- One_el_change_bus -> raw indicator raised if one generic element changed its bus. In the idea to check whether the remapping of solver bus should be checked or not. Nothing to do here.
- ybus_change_sparsity_pattern -> this is raised when a trafo change its tap for example. Nothing to do here.
- slack_weight_chanhed -> I thought the new feature handled only the voltage. Why is slack weight needed here ?

src/core/VoltageControlPlan.cpp
```diff
@@ -0,0 +1,395 @@
...
+    free_vm_slack_buses_.clear();
+    controllers_.clear();
+}
+
+// ---------------------------------------------------------------------------
+// layer 1: which buses a GROUP regulates (grid ids, no labelling needed)
+// ---------------------------------------------------------------------------
+void VoltageControlPlan::build_groups(const GeneratorContainer & generators,
+                                      const SvcContainer & svcs)
+{
```
BDonnot, line 36: Why no HVDC stations here ? Some stations regulates voltages (eg VSC with voltage regulation on)

src/core/VoltageControlPlan.hpp
```diff
@@ -0,0 +1,216 @@
...
+ * nothing said they had to agree.
+ *
+ * Built as one object they cannot: layer 2 is built from layer 1's own result,
+ * layer 3 from layer 2's, and the whole thing is built once and then read.
+ *
+ * ---------------------------------------------------------------------------
+ * THE THREE LAYERS, AND WHAT EACH NEEDS
+ *
+ *  1. `group_controlled_buses()` -- GRID bus ids. Reads only input data
+ *     (generators, SVCs), never the labelling, which is what lets `fillpv_pq`
```
BDonnot: Still don't get why no HVDC here. An already present bug ?

src/core/element_container/ConverterStationContainer.cpp
```diff
@@ -153,6 +153,9 @@ void ConverterStationContainer::change_v(int station_id, real_type new_v_pu, Dua
     {
         solver_control.ac_algo_controler().tell_v_changed();
         solver_control.dc_algo_controler().tell_v_changed();
+        // v_set of the group this station belongs to, when a group claims its bus.
+        // AC only: a DC solve has no voltage control, hence no plan to invalidate.
+        solver_control.ac_algo_controler().tell_voltage_control_changed();
```
BDonnot: Why ? Only value changed in this case. And value is already tagged above

src/core/element_container/ConverterStationContainer.cpp
```diff
@@ -373,6 +376,7 @@ bool ConverterStationContainer::_deactivate(int el_id, DualAlgoControl & solver_
     solver_control.dc_algo_controler().tell_recompute_sbus();
```
BDonnot, line 380: Maybe a "bug" already there, but are the stations in sbus? I thought they were not

src/core/element_container/ConverterStationContainer.cpp
```diff
@@ -373,6 +376,7 @@ bool ConverterStationContainer::_deactivate(int el_id, DualAlgoControl & solver_
...
     solver_control.ac_algo_controler().tell_pv_changed();
     solver_control.dc_algo_controler().tell_pv_changed();
+    solver_control.ac_algo_controler().tell_voltage_control_changed();  // a disconnected station is no longer a group member
```
BDonnot: This should at least depends on the station type no ? Some stations are NOT regulating voltages so no need to raise all that in this case. Or do I miss something?

src/core/element_container/ConverterStationContainer.cpp
```diff
@@ -382,6 +386,7 @@ bool ConverterStationContainer::_reactivate(int el_id, DualAlgoControl & solver_
     solver_control.dc_algo_controler().tell_recompute_sbus();
```
BDonnot: Same as above : o thought stations were not in sbus

src/core/element_container/ConverterStationContainer.cpp
```diff
@@ -382,6 +386,7 @@ bool ConverterStationContainer::_reactivate(int el_id, DualAlgoControl & solver_
...
     solver_control.ac_algo_controler().tell_pv_changed();
     solver_control.dc_algo_controler().tell_pv_changed();
+    solver_control.ac_algo_controler().tell_voltage_control_changed();  // ... and a reconnected one is one again
```
BDonnot: Same as above: only some specific types of stations are regulating voltages, and only if proper flag is set

src/core/element_container/ConverterStationContainer.cpp
BDonnot, line 404: Are station in sbus ?

BDonnot, line 401: Input this comment here because I don't know where to put it otherwise.

 But can the user modify the regulated bus of a station ? 
Of so then it's probably good to have it raise the proper flags.
If not then it's probably something to mark as Todo on the Changelog (todos are written at the top of it)

Same question about the possibility to change "voltage_regulator_on". And same consequences

src/core/element_container/GeneratorContainer.cpp
```diff
@@ -349,13 +349,19 @@ void GeneratorContainer::_change_p(int gen_id, real_type new_p, bool /*my_status
...
+            // no voltage control, hence no plan to invalidate.
+            solver_control.ac_algo_controler().tell_voltage_control_changed();
            }
     }
 }
 
 bool GeneratorContainer::_deactivate(int el_id, DualAlgoControl & solver_control) {
     if(!status_[el_id]) return false;  // nothing to do if it was already deactivated
     solver_control.ac_algo_controler().tell_recompute_sbus(); solver_control.dc_algo_controler().tell_recompute_sbus();
+    solver_control.ac_algo_controler().tell_voltage_control_changed();  // a disconnected generator regulates nothing
```
BDonnot: Only if it was regulating on the first place

src/core/element_container/GeneratorContainer.cpp
```diff
@@ -365,6 +371,7 @@ bool GeneratorContainer::_deactivate(int el_id, DualAlgoControl & solver_control
 bool GeneratorContainer::_reactivate(int el_id, DualAlgoControl & solver_control) {
     if(status_[el_id]) return false;  // nothing to do if gen already connected
     solver_control.ac_algo_controler().tell_recompute_sbus(); solver_control.dc_algo_controler().tell_recompute_sbus();
+    solver_control.ac_algo_controler().tell_voltage_control_changed();  // ... and a reconnected one regulates again
```
BDonnot: Only if it was regulating on the first place

src/core/element_container/GeneratorContainer.cpp
```diff
@@ -392,6 +399,7 @@ void GeneratorContainer::change_v_nothrow(int gen_id, real_type new_v_pu, DualAl
     if (abs(target_vm_pu_(gen_id) - new_v_pu) > _tol_equal_float)
     {
         solver_control.ac_algo_controler().tell_v_changed(); solver_control.dc_algo_controler().tell_v_changed();
+        solver_control.ac_algo_controler().tell_voltage_control_changed();  // v_set of the group this generator belongs to
```
BDonnot: I don't see why here.

src/core/element_container/GeneratorContainer.cpp
```diff
@@ -416,6 +424,7 @@ bool GeneratorContainer::_change_bus(int el_id, GridModelBusId new_bus_id, DualA
     }
     solver_control.ac_algo_controler().tell_recompute_sbus(); solver_control.dc_algo_controler().tell_recompute_sbus();
     solver_control.ac_algo_controler().tell_one_el_changed_bus(); solver_control.dc_algo_controler().tell_one_el_changed_bus();
+    solver_control.ac_algo_controler().tell_voltage_control_changed();  // the controller bus, and a LOCAL regulator's regulated bus, both moved
```
BDonnot: Not if not regulating

src/core/element_container/GeneratorContainer.hpp
```diff
@@ -114,16 +114,34 @@ class LS2G_API GeneratorContainer final: public OneSideContainer_PQ, public Iter
...
+                // is_pseudo_off() answers false for a slack generator whatever its
+                // active power, and a pseudo-off generator is not a voltage
+                // controller -- so taking the slack role changes the control groups.
+                // Only the role: once it is held, the WEIGHT is irrelevant there
+                // (is_pseudo_off returns on gen_slackbus_ before it looks at it), and
+                // this is called for every generator on every step when the
+                // distributed slack is on, so raising it unconditionally would retire
+                // the voltage-control plan on every step for nothing. AC only: a DC
+                // solve has no voltage control, hence no plan to invalidate.
+                solver_control.ac_algo_controler().tell_voltage_control_changed();
```
BDonnot: What slack has to do with any of that ?

src/core/element_container/GeneratorContainer.hpp
```diff
@@ -114,16 +114,34 @@ class LS2G_API GeneratorContainer final: public OneSideContainer_PQ, public Iter
...
-            if(abs(gen_slack_weight_[gen_id]) > _tol_equal_float){ solver_control.ac_algo_controler().tell_slack_weight_changed(); solver_control.dc_algo_controler().tell_slack_weight_changed(); }
+            const bool was_slack = gen_slackbus_[gen_id];
+            const bool had_weight = abs(gen_slack_weight_[gen_id]) > _tol_equal_float;
+            if(was_slack){ solver_control.ac_algo_controler().tell_slack_participate_changed(); solver_control.dc_algo_controler().tell_slack_participate_changed(); }
+            if(had_weight){ solver_control.ac_algo_controler().tell_slack_weight_changed(); solver_control.dc_algo_controler().tell_slack_weight_changed(); }
+            // giving up the slack role, or the weight that stood in for it, can make
+            // this generator pseudo-off and so stop it being a voltage controller.
+            // Conditional for the same reason as add_slackbus above.
+            if(was_slack || had_weight) solver_control.ac_algo_controler().tell_voltage_control_changed();
```
BDonnot: Still don't get the link between slack and voltage control

src/core/element_container/GeneratorContainer.hpp
```diff
@@ -198,11 +216,13 @@ class LS2G_API GeneratorContainer final: public OneSideContainer_PQ, public Iter
...
+            solver_control.ac_algo_controler().tell_voltage_control_changed();  // turnedoff_gen_pv_ gates gen_is_voltage_controller
             turnedoff_gen_pv_=false;  // turned off generators are not pv. This is NOT the default.
             }  
         void turnedoff_pv(DualAlgoControl & solver_control){
             solver_control.ac_algo_controler().tell_slack_participate_changed(); solver_control.dc_algo_controler().tell_slack_participate_changed();
             solver_control.ac_algo_controler().tell_slack_weight_changed(); solver_control.dc_algo_controler().tell_slack_weight_changed();
+            solver_control.ac_algo_controler().tell_voltage_control_changed();  // turnedoff_gen_pv_ gates gen_is_voltage_controller
             turnedoff_gen_pv_=true;  // turned off generators are pv. This is the default.
             }  
         bool get_turnedoff_gen_pv() const {return turnedoff_gen_pv_;}
```
BDonnot, line 224: I put it here because i don't know where to put it in the diff...

But is there something to remove a generator from voltage regulation (coltage_regulation_on_ or something).
If so I think it's missing some flags.
If not a Todo needs to be added in the Changelog (at the top)

Same remark for changing the regulating bus id

src/core/element_container/GeneratorContainer.hpp
```diff
@@ -227,6 +247,7 @@ class LS2G_API GeneratorContainer final: public OneSideContainer_PQ, public Iter
             if(regulated_bus_id_(gen_id) != bus_id){
                 regulated_bus_id_(gen_id) = bus_id;
                 solver_control.ac_algo_controler().tell_pv_changed();  // groups are rebuilt on topology init
+                solver_control.ac_algo_controler().tell_voltage_control_changed();  // this IS the group layout
```
BDonnot: Only if it regulates voltage

src/core/element_container/GeneratorContainer.hpp
```diff
@@ -227,6 +247,7 @@ class LS2G_API GeneratorContainer final: public OneSideContainer_PQ, public Iter
...
                 solver_control.ac_algo_controler().tell_recompute_sbus();
                 solver_control.dc_algo_controler().tell_pv_changed();  // groups are rebuilt on topology init
```
BDonnot: Only if it regulates voltage

src/core/element_container/GeneratorContainer.hpp
```diff
@@ -227,6 +247,7 @@ class LS2G_API GeneratorContainer final: public OneSideContainer_PQ, public Iter
             if(regulated_bus_id_(gen_id) != bus_id){
                 regulated_bus_id_(gen_id) = bus_id;
                 solver_control.ac_algo_controler().tell_pv_changed();  // groups are rebuilt on topology init
+                solver_control.ac_algo_controler().tell_voltage_control_changed();  // this IS the group layout
                 solver_control.ac_algo_controler().tell_recompute_sbus();
```
BDonnot: Why ?

src/core/element_container/SvcContainer.cpp
BDonnot, line 261: Same question I don't know where to put...

1) possibility to change regulated bus
2) possibility to change the regulation mode of SVC

If not present -> add a Todo
Of so, some flags are probably missing

BDonnot (commented):
Lots of changes to make and things that are unclear

Also some tests might need to be added for the flag raised in the grid modifiers.

src/core/Utils.hpp
BDonnot: You were right on essentially all of it, and moving `fillpv_pq` into the plan is what makes it obvious. Fixed in 7ec0ba3.

The pv/pq split is now **layer 2 of the plan**, so "the split must be rebuilt" and "the plan must be rebuilt" are the same question asked one layer apart. Who is in a control group, and which bus it regulates, are *exactly* the inputs the split reads — so `pv_changed_` / `slack_participate_changed_` already carried all of it, and me re-listing their causes was duplication.

There are now two predicates:

```cpp
bool need_recompute_pv_pq() const noexcept {
    return need_reset_solver_ || change_dimension_ || ybus_change_sparsity_pattern_ ||
           slack_participate_changed_ || pv_changed_ || pq_changed_;
}
bool need_recompute_voltage_control() const noexcept {
    return need_recompute_pv_pq() || voltage_control_changed_;
}
```

Point by point:

* `slack_weight_changed_`, `one_el_change_bus_` — **gone**. You were right, neither belongs.
* `v_changed_` — **gone**. Replaced by the narrow `voltage_control_changed_`, which is now raised from exactly two places (`GeneratorContainer::change_v_nothrow`, `ConverterStationContainer::change_v`), down from fifteen.
* `slack_participate_changed_` — **kept, but as a pv/pq term, not a voltage one**: the slack is not PV, so moving the slack set moves the split. The voltage link you doubted (a slack generator is never `is_pseudo_off`, hence always a controller) is real but is *carried by that same flag*, which is why `add_slackbus` / `remove_slackbus` no longer raise anything voltage-specific.
* `ybus_change_sparsity_pattern_` — **kept, but as a pv/pq term**: it is what makes `init_converter_bus_id` run, i.e. it re-labels the buses, and the split is expressed in that labelling. You're right that a tap change has nothing to do with voltage control; it does have something to do with the split.
* `pv_changed_` / `pq_changed_` (your "doubt") — **kept, and now clearly justified**: they *are* the split.
* `change_dimension_`, `need_reset_solver_`, `cache_maybe_poisoned_` — kept (the last implies `need_reset_solver_`).

`_build_into_cache` now reads those two and nothing else, so what the powerflow asks and what a test asks cannot drift apart.

---

src/core/VoltageControlPlan.cpp
BDonnot: Not an oversight, but the code did not say so — now it does (7ec0ba3), both here and on the declaration.

A station **has no regulated bus of its own**: `ConverterStationContainer` stores no `regulated_bus_id_`, so a regulating station always regulates the bus it stands on. It therefore can never be the *remote* controller that makes a bus need the bordered treatment — which is what this function is looking for. It pins its own bus through the ordinary PV path (`ConverterStationContainer::fillpv`, which skips `!voltage_regulator_on_`), exactly like a local generator.

A station becomes a *member* of a group only when a group formed by one of the two loops here already claims its bus, and that enrolment is `build_controllers`' job (`_collect_station_controllers`) — keyed on the set this function returns. Adding stations here would be circular as well as wrong.

Remote regulation **by** a station is simply not modelled, which is a real gap rather than a bug; added to the changelog TODO along with the other missing setters.

---

src/core/VoltageControlPlan.hpp
BDonnot: Same answer as on the `.cpp` — see [that thread](https://github.com/Grid2op/lightsim2grid/pull/191#discussion_r3942900486). Short version: not a bug. A station has no regulated bus of its own (the container stores none), so it can only ever regulate the bus it stands on, through the ordinary PV path — it can never *create* a group, only join one somebody else created, which `build_controllers` does. Both ends now say so explicitly in 7ec0ba3, and "remote regulation by a station is not modelled" is now a changelog TODO.

---

src/core/element_container/ConverterStationContainer.cpp
BDonnot: This one I kept, but your objection made me narrow it and it is now the *only* reason the flag still exists.

`tell_v_changed()` above is not enough because `v_changed_` is no longer part of `need_recompute_voltage_control()` — you were right that it did not belong in that union (see [the Utils.hpp thread](https://github.com/Grid2op/lightsim2grid/pull/191#discussion_r3942898769)). And a setpoint really is the one input of the plan that nothing else carries: moving a remote regulator's target changes **no bus' pv/pq class at all** — that is the point of the bordered formulation, the regulated bus stays PQ — so `pv_changed_` never fires and the plan would be reused with a stale `v_set`.

Narrowed to only fire when the station actually regulates, since `target_vm_pu_` of one that does not is never read:

```cpp
if(voltage_regulator_on_[station_id]) solver_control.ac_algo_controler().tell_voltage_control_changed();
```

There is a test for it, and it fails if the term is dropped from the predicate.

---

src/core/element_container/ConverterStationContainer.cpp
BDonnot: Checked: **yes, stations are in Sbus.** `ConverterStationContainer::fillSbus_station` stamps every connected station — its active power always, plus the reactive personality when it is *not* a voltage regulator (an LCC / power-factor station). So `tell_recompute_sbus()` here is correct and stays, for a regulating station too.

Added that as a comment on the line in 7ec0ba3, since it is clearly not obvious from the call site.

(The `tell_pv_changed()` two lines below is a different story — see [the next thread](https://github.com/Grid2op/lightsim2grid/pull/191#discussion_r3942941262).)

---

src/core/element_container/ConverterStationContainer.cpp
BDonnot: You're right, and this one was wrong **before** this branch — `tell_pv_changed()` was unconditional here. `ConverterStationContainer::fillpv` skips `!voltage_regulator_on_[station_id]`, so a non-regulating station pins no bus and its status cannot move the split. Fixed in 7ec0ba3:

```cpp
// a station IS in Sbus (fillSbus_station stamps its active power, and its reactive
// personality when it does not regulate), so its status always moves the injections
solver_control.ac_algo_controler().tell_recompute_sbus();
solver_control.dc_algo_controler().tell_recompute_sbus();
// ... but only a REGULATING one pins a bus (fillpv skips the others), so only that
// one moves the pv/pq split -- and with it the voltage-control plan built on it
if(voltage_regulator_on_[el_id]){
    solver_control.ac_algo_controler().tell_pv_changed();
    solver_control.dc_algo_controler().tell_pv_changed();
}
```

My `tell_voltage_control_changed()` is gone from here entirely: with the split as layer 2 of the plan, the gated `tell_pv_changed()` already rebuilds it. Same in `_reactivate`.

---

src/core/element_container/ConverterStationContainer.cpp
BDonnot: Yes they are — `fillSbus_station` stamps every connected station (active power always, reactive too when it does not regulate). Same answer as [the `_deactivate` thread](https://github.com/Grid2op/lightsim2grid/pull/191#discussion_r3942939455); the line is now commented to that effect, and `_reactivate` points at `_deactivate` for it.

---

src/core/element_container/ConverterStationContainer.cpp
BDonnot: Agreed — `_reactivate` is fixed the same way as `_deactivate` in 7ec0ba3: `tell_pv_changed()` now only for `voltage_regulator_on_[el_id]` (that is exactly `ConverterStationContainer::fillpv`'s own gate), `tell_recompute_sbus()` unconditional because a station is in Sbus either way, and my `tell_voltage_control_changed()` removed as redundant with the gated pv flag.

---

src/core/element_container/ConverterStationContainer.cpp
BDonnot: Yes — see [above](https://github.com/Grid2op/lightsim2grid/pull/191#discussion_r3942939455). `fillSbus_station` stamps them, so `tell_recompute_sbus()` on a bus change is right and stays.

---

BDonnot: Checked both: **neither is settable at runtime**, so it is the Todo branch, not the missing-flag branch. Added to the changelog TODO in 7ec0ba3.

* a station's **regulated bus** does not exist as a concept — `ConverterStationContainer` stores no `regulated_bus_id_` at all, so a regulating station always regulates the bus it stands on. Remote regulation *by* a station is not modelled; that is its own TODO line.
* a station's **`voltage_regulator_on_`** has no setter: it is written by `init()` / `set_state()` only, i.e. you have to rebuild the container through `init_hvdc_lines`.

Both are missing *capabilities* rather than missing flags — there is no setter to raise a flag from — and the TODO says that whoever adds the setter must raise `tell_pv_changed()` (it moves who pins a bus, and so the pv/pq split and the plan built on it) with it.

Same shape for the generator and the SVC, all four collected in one TODO block at the top of the changelog.

---

src/core/element_container/GeneratorContainer.cpp
BDonnot: Right — and once gated on "was it regulating", the line becomes redundant, so it is **removed** rather than moved (7ec0ba3).

`if(voltage_regulator_on_[el_id]){ tell_pv_changed(); }` is already on the very next line, and with the pv/pq split now being layer 2 of the voltage-control plan, that flag rebuilds the plan too. A generator that was not regulating was never in a control group, so disconnecting it changes nothing there.

Same in `_reactivate`, `_change_bus` and `_change_p`.

---

src/core/element_container/GeneratorContainer.cpp
BDonnot: Same as [`_deactivate`](https://github.com/Grid2op/lightsim2grid/pull/191#discussion_r3942955127): removed in 7ec0ba3. The `if(voltage_regulator_on_[el_id]) tell_pv_changed()` right below already covers it, now that the split is a layer of the plan.

---

src/core/element_container/GeneratorContainer.cpp
BDonnot: This is `change_v_nothrow`, and it is the one place the flag genuinely earns its keep — but your comment made me narrow it, and it is now one of only two sites left (down from fifteen).

Why `tell_v_changed()` above is not enough: `v_changed_` is no longer part of `need_recompute_voltage_control()`, because you were right that it did not belong in that union. And a setpoint is precisely the input the pv/pq split does **not** read — moving a remote regulator's target changes no bus' class at all (the regulated bus of a group stays PQ), so `pv_changed_` never fires and the plan would be reused with a stale `v_set` in its bordered voltage row. That is a converged, wrong answer.

Now gated, since `target_vm_pu_` of a non-regulating generator is never read:

```cpp
if(voltage_regulator_on_[gen_id]) solver_control.ac_algo_controler().tell_voltage_control_changed();
```

Pinned by `test_voltage_control_plan_cache.cpp` ("a group member's voltage setpoint"), which fails if the term is dropped from the predicate.

---

src/core/element_container/GeneratorContainer.cpp
BDonnot: Agreed — removed in 7ec0ba3. `_change_bus` already has `if(voltage_regulator_on_[el_id]) tell_pv_changed()` just below, which is both the right gate and (now that the split is layer 2 of the plan) enough to rebuild the plan. A generator that does not regulate carries no group membership to move.

---

src/core/element_container/GeneratorContainer.hpp
BDonnot: There *is* a link, but you're right that it does not justify a line here — **removed** in 7ec0ba3.

The link, for the record: `gen_is_voltage_controller()` gates on `is_pseudo_off()`, and `is_pseudo_off()` answers false for a slack generator whatever its active power (`if (gen_slackbus_[gen_id]) return false;`). So a p == 0 regulating generator that takes the slack role *starts* being a controller.

But `tell_slack_participate_changed()` on the line above already fires in exactly that branch, and `slack_participate_changed_` is part of `need_recompute_pv_pq()` — which, now that the split is layer 2 of the plan, rebuilds the plan. So my line was pure duplication.

It was also the one I was least happy with: `LightSimBackend._handle_dist_slack` calls `update_slack_weights` on every step of a distributed-slack episode, so it had to be carefully conditional not to retire the plan every step. Deleting it removes that worry entirely. The test that pinned the behaviour ("re-declaring the same slack weights reuses the plan") stays and still passes.

Same in `remove_slackbus`, and in `turnedoff_pv` / `turnedoff_no_pv` for the same reason.

---

src/core/element_container/GeneratorContainer.hpp
BDonnot: Removed in 7ec0ba3 — see [the `add_slackbus` thread](https://github.com/Grid2op/lightsim2grid/pull/191#discussion_r3942964748) for the (real but redundant) link. `remove_slackbus` is back to exactly what it was before this branch: `tell_slack_participate_changed()` when the role goes, `tell_slack_weight_changed()` when the weight does, nothing else.

---

src/core/element_container/GeneratorContainer.hpp
BDonnot: Checked both:

* **regulating bus id** — `set_regulated_bus` exists (`LSGrid::set_gen_regulated_bus`) and does raise what it must. It was raising slightly *too much*, which your other two comments caught; fixed in 7ec0ba3.
* **`voltage_regulator_on_`** — no setter at all. It is written by `init()` / `set_state()` only, so a caller cannot take a machine out of voltage regulation, or put it back, without rebuilding the container through `init_generators_full`. Not a missing flag (there is no setter to raise one from) but a missing capability → **Todo**, added at the top of the changelog.

The TODO block collects all four of these (generator and station `voltage_regulator_on_`, SVC `regulation_mode_` and `regulated_bus_id_`, plus "remote regulation by a station is not modelled") and says explicitly that whoever adds a setter must raise `tell_pv_changed()` — or `tell_voltage_control_changed()` where it only moves a setpoint — with it.

---

src/core/element_container/GeneratorContainer.hpp
BDonnot: Agreed, and this one was wrong **before** this branch too — the `tell_pv_changed()` was unconditional. `regulated_bus_id_` is only ever read for a generator that regulates (it decides whether the machine pins its own bus or joins a group at another one); for one that does not, the field is inert. 7ec0ba3:

```cpp
if(regulated_bus_id_(gen_id) != bus_id){
    regulated_bus_id_(gen_id) = bus_id;
    // Only a generator that actually regulates reads this field: [...] For a
    // non-regulating generator the value is stored and nothing else changes.
    if(voltage_regulator_on_[gen_id]){
        solver_control.ac_algo_controler().tell_pv_changed();
        solver_control.dc_algo_controler().tell_pv_changed();
    }
}
```

The value is still stored unconditionally — only the flags are gated. My `tell_voltage_control_changed()` here is gone as redundant with the (now correct) pv flag.

---

src/core/element_container/GeneratorContainer.hpp
BDonnot: Same fix — the DC `tell_pv_changed()` is inside the same `if(voltage_regulator_on_[gen_id])` now (7ec0ba3). See [the AC one](https://github.com/Grid2op/lightsim2grid/pull/191#discussion_r3942977711).

---

src/core/element_container/GeneratorContainer.hpp
BDonnot: Good catch — there is no reason. **`tell_recompute_sbus()` removed** from `set_regulated_bus` in 7ec0ba3 (both families).

`GeneratorContainer::fillSbus` never reads `regulated_bus_id_`. What it stamps is `target_p_mw_`, plus `target_q_mvar_` only when `!voltage_regulator_on_`:

```cpp
tmp = {target_p_mw_(gen_id), 0.};
if(!voltage_regulator_on_[gen_id]){
    tmp += my_i * target_q_mvar_(gen_id);   // gen is pq if voltage regulation is off
}
```

So for a regulating generator the injection is the same wherever it regulates, and for a non-regulating one the field is not read at all. The flag was pure over-invalidation — pre-existing, from before this branch.

---

src/core/element_container/SvcContainer.cpp
BDonnot: Checked: **neither is settable at runtime** for an SVC, so it is the Todo branch. Added to the changelog TODO in 7ec0ba3.

* `regulation_mode_` (OFF / VOLTAGE / REACTIVE_POWER): written by `init()` / `set_state()` only, so an SVC cannot be switched between modes after `init_svcs`.
* `regulated_bus_id_`: same — the only place it is touched afterwards is `_change_bus`, which drags a *local* SVC's regulated bus along with its own bus. There is no way to point an SVC at a different bus.

Both are missing capabilities rather than missing flags, and the TODO says the setter must come with `tell_pv_changed()` (an active voltage-mode SVC is always a group controller, so either change moves the pv/pq split and the plan built on it).

Meanwhile the three flags I had added in this file are **removed**: `_deactivate` / `_reactivate` / `_change_bus` already raise `tell_pv_changed()` inside their `regulation_mode_(svc_id) == VOLTAGE` branch, which is both the right gate and — now that the pv/pq split is layer 2 of the voltage-control plan — enough to rebuild it.

---

BDonnot:
## Review answered in 7ec0ba3

You were right that moving `fillpv_pq` into the plan changes the answers — but mostly by making your objections *more* right, not less. The pv/pq split is now **layer 2 of the plan**, so "the split must be rebuilt" and "the plan must be rebuilt" are the same question asked one layer apart, and the flags I had been listing were already carried by `pv_changed_` / `slack_participate_changed_`.

**The predicate.** Two now, both on `AlgoControl`, and the second is visibly the first plus one term:

```cpp
bool need_recompute_pv_pq() const noexcept {
    return need_reset_solver_ || change_dimension_ || ybus_change_sparsity_pattern_ ||
           slack_participate_changed_ || pv_changed_ || pq_changed_;
}
bool need_recompute_voltage_control() const noexcept {
    return need_recompute_pv_pq() || voltage_control_changed_;
}
```

`slack_weight_changed_`, `one_el_change_bus_` and `v_changed_` are gone. `slack_participate_changed_` and `ybus_change_sparsity_pattern_` stay, but as *pv/pq* terms — the slack is not PV, and the sparsity flag is what re-labels the buses — which is a reason you can check rather than a claim about voltage control. `_build_into_cache` reads those two and nothing else, so what the powerflow asks and what a test asks cannot drift.

**The propagation: 15 call sites → 2.** `tell_voltage_control_changed()` now means one thing only — a voltage **setpoint** moved — and lives in `GeneratorContainer::change_v_nothrow` and `ConverterStationContainer::change_v`, gated on `voltage_regulator_on_`. That is the one input the split does not read: moving a remote regulator's target changes no bus' class at all. Everywhere else you flagged (`_deactivate`, `_reactivate`, `_change_bus`, `_change_p`, `add_slackbus`, `remove_slackbus`, `turnedoff_*`, `set_regulated_bus`, the three SVC sites) the line is deleted, because the `tell_pv_changed()` / `tell_slack_participate_changed()` already there does the job.

**Three flags were wrong before this branch**, and your comments found all three:

| where | was | now |
|---|---|---|
| `GeneratorContainer::set_regulated_bus` | `tell_recompute_sbus()` | **removed** — `fillSbus` never reads `regulated_bus_id_`, and what it stamps for a regulating gen (active power only) is the same wherever it regulates |
| `GeneratorContainer::set_regulated_bus` | `tell_pv_changed()` unconditional | gated on `voltage_regulator_on_` — the field is inert otherwise |
| `ConverterStationContainer::_deactivate` / `_reactivate` | `tell_pv_changed()` for every station | gated on `voltage_regulator_on_` — `fillpv` skips the others |

**Are stations in Sbus?** Yes — `fillSbus_station` stamps every connected one (active power always, reactive too when it does not regulate). So `tell_recompute_sbus()` there is right and stays; the line is now commented to say so.

**Why no HVDC in `build_groups`?** Not a bug. A station has no regulated bus of its own — the container stores none, a regulating station always regulates the bus it stands on — so it can never be the *remote* controller that creates a group. It pins its own bus through `fillpv` like a local generator, and joins a group only when one already claims that bus, which is `build_controllers`' job and is keyed on what `build_groups` returns. Spelled out at both ends now.

**TODOs added** (top of the changelog) for the four inputs with no runtime setter at all, so a caller cannot change them on a live grid: a generator's and a station's `voltage_regulator_on_`, an SVC's `regulation_mode_` and `regulated_bus_id_`, and remote regulation *by* a station, which is not modelled. Each says what flag the setter must raise.

**Left open deliberately** — four threads where I did not do what was asked and you should judge: the two `change_v` sites (I kept the flag, narrowed; it is the only non-redundant use left) and the two HVDC-in-`build_groups` ones (I answered rather than changed code).

**Checks.** C++ 247 cases / 591,133 assertions green; targeted python 245 passed / 65 skipped; full python suite 1440 passed, the 99 failures all missing grid2op `data_test` fixtures in this container and identical on the base. Instruction counts unchanged and all 14 (grid, phase) traces still bit-identical to the pre-PR build. Both invalidation paths mutation-checked: dropping the setpoint term, and silencing `set_regulated_bus`, each make a specific test fail.

---

src/core/Utils.hpp
```diff
@@ -238,6 +259,53 @@ class AlgoControl final
...
+         * the bus set changed (`change_dimension_`), the bus LABELLING changed
+         * (`ybus_change_sparsity_pattern_`, which is what makes
+         * `LSGrid::init_converter_bus_id` run), the slack set moved
+         * (`slack_participate_changed_`, and the slack is not PV), or a bus changed
+         * class (`pv_changed_` / `pq_changed_`).
+         *
+         * Read by `LSGrid::_build_into_cache`, which is the only thing that rebuilds it.
+         */
+        [[nodiscard]] bool need_recompute_pv_pq() const noexcept {
+            return need_reset_solver_ || change_dimension_ || ybus_change_sparsity_pattern_ ||
```
BDonnot: Disagree with ybus_changed_sparsity_pattern. Make a test to resolve this issue.

src/core/Utils.hpp
```diff
@@ -238,6 +259,53 @@ class AlgoControl final
...
+                   slack_participate_changed_ || pv_changed_ || pq_changed_;
```
BDonnot, line 293: Disagree with slack_partcipate_changed, make another test

src/core/element_container/GeneratorContainer.hpp
```diff
@@ -114,6 +114,11 @@ class LS2G_API GeneratorContainer final: public OneSideContainer_PQ, public Iter
         void add_slackbus(int gen_id, real_type weight, DualAlgoControl & solver_control){
             // TODO DEBUG MODE
             if(weight <= 0.) throw std::runtime_error("GeneratorContainer::add_slackbus Cannot assign a negative (<=0) weight to the slack bus.");
+            // (taking the slack role does change who is a voltage controller --
+            // is_pseudo_off() answers false for a slack generator whatever its active
+            // power -- but tell_slack_participate_changed() above already rebuilds the
+            // pv/pq split, and the voltage-control plan is built around that split. See
+            // AlgoControl::need_recompute_voltage_control.)
```
BDonnot: That's the thing. A slack (not the ref one) has nothing to do with voltage.
So I don't see why it changes the voltage control cache.

src/tests/test_cache_reuse.cpp
```diff
@@ -813,10 +813,11 @@ TEST_CASE("a divergence recovers under every built-in AC algorithm", "[LSGrid][c
     // check_solution() is thrown into the sequence on purpose: it runs the same
```
BDonnot, line 813: Putting this comment here but it concerns the whole file. The test I suggested to resolve the debate whether set q voltage setpoint of q generator, or the slack weights or the ybus sparsity pattern would be in this file.

The idea is, eg for ybus sparsity pattern:
- find where it's activated (in this case disconnecting a line).
- run a powerflow on the base case.
- do the action identified at step 1
- run a second poweflow (the voltage cache should not be rebuilt)
- save the results (voltages at each bus)

Then tell the solver needs to be reset 
Re run the powerflow (cache will be fresh as if it were just built).

Compare the results with the one saved.

If that match then the flag (in this example change sparsity pattern) can be removed from "recompute_pv_pq"

BDonnot (commented):
A few tests to do

src/core/Utils.hpp
BDonnot: You were right. The test says so, and the term is **gone** in 18988f9.

I built your recipe as two `[pv_pq]`-tagged cases in `test_cache_reuse.cpp`, then built the predicate both ways and ran them against each:

```
dropping ybus_change_sparsity_pattern_  ->  41 assertions, all pass
dropping slack_participate_changed_     ->  2 cases, 1 fails, 2 assertions fail
```

So:

```cpp
bool need_recompute_pv_pq() const noexcept {
    return need_reset_solver_ || change_dimension_ ||
           slack_participate_changed_ || pv_changed_ || pq_changed_;
}
```

Why it was redundant, now that I had to look properly: this flag is only ever raised by a **branch-side** mutation — `TwoSidesContainer_rxh_A::_reactivate`, `OneSideContainer_forBranch::_reactivate` / `_change_bus`. Every one of those goes through `GenericContainer::_apply_and_track_buses`, which raises `change_dimension_` exactly when the mutation empties or fills a bus — which is exactly when the labelling moves. If no bus crossed, `init_converter_bus_id` recomputes `id_me_to_solver` to the same thing, and a branch is neither a voltage controller nor a slack, so the split it produces is identical. My "it re-labels the buses, and the split is expressed in that labelling" was true of the flag's *name* and not of what it can actually do.

Measured on its own (same tree, term restored, `topo` phase, answers bit-identical on all seven grids):

| grid | term kept | term dropped |
|---|---:|---:|
| case30 | 440,089 | 436,057 (**-0.9%**) |
| case118 | 1,470,391 | 1,444,534 (**-1.8%**) |
| case1354pegase | 20,163,544 | 20,013,614 (**-0.7%**) |
| case9241pegase | 182,113,949 | 181,393,356 (**-0.4%**) |

(only every other solve of `topo` raises the flag, so that is roughly half a split rebuild per solve.)

---

src/core/Utils.hpp
BDonnot: Same experiment, opposite verdict: this one is **load-bearing**, and the test now proves it rather than asserting it. `slack_participate_changed_` stays (18988f9).

`TEST_CASE("slack_participate_changed, on its own, DOES move the split", "[LSGrid][cache_reuse][pv_pq]")` — two sections, the slack moved to another generator (`update_slack_weights_by_id`) and a second generator joining it (`add_gen_slackbus`). Each asserts the term is raised **and the other five are not**, then solves warm, resets, solves cold and compares. With the term dropped from the predicate:

```
  the slack moves to another generator
  a second generator joins the slack
test cases:  2 |  1 passed | 1 failed
assertions: 37 | 35 passed | 2 failed
```

And the reason has nothing to do with voltage, which is where I think our disagreement was — you are right that a non-reference slack has nothing to do with voltage. It is that the split reads the slack set **directly**:

```cpp
// GeneratorContainer::fillpv
if(is_in_vect(bus_id_solver.cast_int(), slack_bus_id_solver.to_int_vector())) continue;  // slack bus is not PV
```

and the PQ loop right after it skips slack buses too. So a bus joining the slack set must leave `bus_pv`, and one leaving it must join `bus_pv` or `bus_pq`. Move the slack set and the split moves with it, whatever the machine does to voltage.

---

src/core/element_container/GeneratorContainer.hpp
BDonnot: Agreed on the premise, and the comment was leading with the wrong reason. Rewritten in 18988f9.

A non-reference slack has nothing to do with voltage — but it has everything to do with the **pv/pq split**, and that is what `tell_slack_participate_changed()` is for here. The split reads the slack set directly (`fillpv`: *"slack bus is not PV"*, and the PQ loop after it skips slack buses too), so a bus joining or leaving the slack set changes which list it is in. Since the split is layer 2 of the voltage-control plan, the plan is rebuilt as a consequence — not because the slack role means anything to voltage control. That is now what the comment says first, and it points at the `[pv_pq]` test that fails if the flag is dropped.

The voltage link is kept as the secondary paragraph it deserves, because it is real and non-obvious: `is_pseudo_off()` starts with `if (gen_slackbus_[gen_id]) return false;`, so a zero-P *slack* generator is a voltage controller where an ordinary zero-P one would not be. It needs no flag of its own — the same `tell_slack_participate_changed()` covers it.

```cpp
// Why `tell_slack_participate_changed()` and nothing about voltage: taking
// the slack role is an ACTIVE-power role, but the pv/pq split reads the
// slack set directly -- `fillpv` skips a bus that is in
// `slack_bus_id_solver` ("slack bus is not PV"), and the PQ loop right
// after it skips it too -- so a bus joining or leaving the slack set moves
// the split, whatever it does to voltage.
```

---

src/tests/test_cache_reuse.cpp
BDonnot: Built exactly this, in this file, plus the refinement you added afterwards ("a state where only one flag is raised independently of the others") — section 8 of `test_cache_reuse.cpp`, tag `[pv_pq]`, in 18988f9.

Your recipe verbatim, with the isolation step in front of it:

```cpp
struct PvPqTerms { bool reset, dimension, sparsity, slack_participate, pv, pq; };
```

* solve, so the cache describes the grid;
* apply the action, and **read all six terms** — the case asserts the one under test is raised and the other five are not, otherwise a neighbour does the work and the comparison proves nothing;
* solve WARM (on the cache the action left behind);
* `tell_solver_need_reset()`, solve COLD;
* the two answers must agree to 1e-9.

Choosing the action took a couple of tries, exactly as you suspected. `deactivate_powerline` was not enough on its own (it raises `ybus_some_coeffs_zero_` + `one_el_change_bus_`, not the sparsity flag), and a line whose end sits alone on a bus drags `change_dimension_` in with it. What works on the exotic grid is *reconnecting* line 3 — both its ends carry other elements, so the bus set does not move — and, as a second section, moving one end of trafo 0. For the slack: `update_slack_weights_by_id({1})` and `add_gen_slackbus(1, 1.0)`.

Then the decisive part, which is your last sentence: I built `need_recompute_pv_pq()` both ways and ran these cases against each.

```
=== dropped: no_sparsity ===
All tests passed (41 assertions in 2 test cases)
=== dropped: no_slack ===
  the slack moves to another generator
  a second generator joins the slack
test cases:  2 |  1 passed | 1 failed
assertions: 37 | 35 passed | 2 failed
```

So `ybus_change_sparsity_pattern_` is removed and `slack_participate_changed_` stays, and both cases stay in the suite: the first is now a live regression test (the warm path really does reuse the split), the second a guard that fails the day someone drops the term without checking. Full suite green: 249 cases, 591,174 assertions.

---

BDonnot (commented):
Ok

PR #192, opened 2026-09-07
Recover a stranded singleton voltage controller instead of diverging

BDonnot:
## Summary

- Fixes a `ContingencyAnalysis` / `SecurityAnalysis` `handle_disconnected_grid` bug: a contingency that strands a lone (single-controller) remote voltage regulator's own bus — e.g. a generator that reaches its regulated bus only through a transformer that trips — used to come back as a DIVERGENCE. Bus masking is a value-only Jacobian edit (no sparsity/dimension change, so the symbolic factorization is reused), but the `VoltageControl` extension's bordered voltage row survived untouched and kept demanding `Vm(regulated) == v_set` while the controller's own reactive-injection unknown lost its only coupling once its bus's P/Q rows were masked to identity — a structurally singular system.
- For a group with exactly one controller, that row is now repurposed (only when the mask strands the controller's own bus) into a plain `Q_c == 0` pin, using one extra Jacobian slot per such controller. The regulated bus then falls back to its own ordinary PQ equations, matching what a rebuilt (single-shot) topology already does once the disconnected controller drops out of its group.
- That extra slot is reserved **only** when it can ever be used: `NRAlgo` / `AlgorithmSelector` gained `set_may_mask_voltage_control(bool)`, wired by `BaseBatchSweep` exactly when `handle_disconnected_grid` is enabled for a given run (on the long-lived member algo, with a forced sparsity rebuild since it may already have sparsity built; and on every freshly-spawned per-thread algo in the multi-threaded path, no rebuild needed there). Plain `ac_pf`, `TimeSeriesCPP`, and a `ContingencyAnalysis`/`ScenarioSweep` that never turns `handle_disconnected_grid` on stay bit-for-bit as cheap as before this change, regardless of how many remote-regulating generators or SVCs the grid has.

## Scope

Deliberately narrow, matching an explicit request during review:
- Only a group with exactly one controller (`cnt == 1`) is covered. A group shared by more than one controller regulating the same bus is untouched — stranding one of several members stays exactly as it was before (it happens not to diverge either, for an unrelated reason: the surviving sharing-row coupling keeps the system non-singular; see the new "unaffected by the cnt==1 fallback" test).
- Stranding the *regulated* bus itself (rather than a controller's bus) also stays a DIVERGENCE — that case has no well-defined fallback under the "same sparsity every contingency" masking design, and even a rebuilt single-shot topology has no support for it either (it raises).
- A finer per-controller gate (reserving the slot only for controllers whose bus appears in *this run's actual* contingency set, using the `_li_masked` data `_select_ref_slack_and_masks` already computes) was considered and intentionally not implemented: it would need a new "contingency set changed since sparsity was built" invalidation path that doesn't exist today, and getting it wrong would silently reintroduce the same singular-Jacobian bug this PR fixes.

## Test plan

- [x] Full C++ suite built from scratch (Eigen `NR_SparseLU`, no KLU) against this branch's `dev_1.0.1` base: 252 test cases / 591203 assertions, all green.
- [x] New regression tests in `src/tests/test_batch_voltage_control.cpp`:
  - the fix itself: a stranded lone controller now converges and matches a manually-disconnected single-shot reference, with the `handle_disconnected_grid=False` case unchanged (still skipped outright)
  - the `cnt>1` scope boundary is pinned as unaffected
  - the fix also works through the multi-threaded batch path (exercises the per-thread `set_may_mask_voltage_control` wiring specifically), run 3x to rule out scheduling flakiness

https://claude.ai/code/session_01Dkbg9xyFV7Paj7ui2GQSyH

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dkbg9xyFV7Paj7ui2GQSyH)_

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1009,9 +1009,6 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
   the check is free. What remains on the powerflow path is the ``allow_*_cache_reuse`` switch plus
   a debug-only assertion, so a future third claimant is caught by the C++ suite (run under ASan,
   UBSan and valgrind in CI) at no cost in release wheels.
-
-[1.0.0] 2026-08-28
---------------------
 - [FIXED] ``ContingencyAnalysis`` / ``SecurityAnalysis``'s ``handle_disconnected_grid`` mode no
   longer reports a DIVERGENCE for a contingency that strands a lone (single-controller) remote
   voltage regulator's own bus while leaving the bus it regulates connected -- e.g. a generator
@@ -1042,6 +1039,9 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
   *regulated* bus itself (rather than a controller's bus) also stays a DIVERGENCE: unlike this
   case, that one has no well-defined fallback under the "same sparsity every contingency" masking
   design, and even a rebuilt single-shot topology has no support for it either (it raises).
+
+[1.0.0] 2026-08-28
+--------------------
 - [BREAKING] **solver cache reuse is now automatic and on by default**, per solver family.
   A powerflow reuses what the previous one of the same family built -- the compact bus labelling,
   ``Ybus`` / ``Sbus``, the PV / PQ split, the slack weights -- and only re-stamps what the grid
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1012,6 +1012,26 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
 
 [1.0.0] 2026-08-28
 --------------------
+- [FIXED] ``ContingencyAnalysis`` / ``SecurityAnalysis``'s ``handle_disconnected_grid`` mode no
+  longer reports a DIVERGENCE for a contingency that stands a lone (single-controller) remote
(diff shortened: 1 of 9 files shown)


CHANGELOG.rst
```diff
@@ -1012,6 +1012,36 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
...
+  regardless of how many remote-regulating generators or SVCs the grid has. A sloped SVC keeps
+  reserving its coupling slot unconditionally as before (needed for the slope itself, independent
+  of masking).
+  A group shared by more than one controller (``cnt > 1``) is untouched by this fix: stranding one
+  of several controllers regulating the same bus stays exactly as before (see
+  ``src/tests/test_batch_voltage_control.cpp`` for what that already does -- it happens not to
+  diverge, since the surviving sharing-row coupling keeps the system non-singular). Stranding the
+  *regulated* bus itself (rather than a controller's bus) also stays a DIVERGENCE: unlike this
+  case, that one has no well-defined fallback under the "same sparsity every contingency" masking
+  design, and even a rebuilt single-shot topology has no support for it either (it raises).
```
BDonnot, line 1041: Should ne move on 1.0.1, it missed the 1.0.0 deadline 😊

BDonnot (commented):
Need to modify Changelog

PR #193, opened 2026-09-07
Disconnect generators per row in ScenarioSweep, without re-analyzing the Jacobian

BDonnot:
Two commits: the feature, and a small unrelated `CLAUDE.md` that is easy to review separately.

---

# 1. Disconnect generators per row in ScenarioSweep, without re-analyzing the Jacobian

## What

`ScenarioSweep` gains a third contingency axis: `set_contingency_gens(mask)`, a `(n_simul, n_gen)` boolean mask disconnecting generators row by row, alongside the existing line and trafo masks. It does three things:

- takes the machine's active power out of that row's injection (its reactive setpoint too, when it does not regulate voltage);
- re-weights the distributed slack without it, renormalised;
- and, when the **last** generator regulating its own bus goes, turns that bus from PV to PQ for that row — its magnitude is solved for instead of held at the setpoint.

The lost MW is picked up by the slack. A caller wanting redispatch says so with `modify_gen_p`.

## Why it is not just "change pv/pq per row"

A sweep is fast precisely because its labelling never changes: `need_rebuild` (`NRAlgo.tpp:29-39`) is false after the first row, so the loop is **one `analyze` + one `factorize`, then only `refactorize`** (`NRAlgo.tpp:99-107`). Relabelling a bus per row would raise `has_pv_changed()` and buy a fresh symbolic factorization on every row — the one thing this must not cost.

So the labelling is not changed. Instead:

- **`Base::set_switchable_vm_buses`** reserves a `Vm` unknown + a `Q` equation for every bus that *can* flip, once, before the "n" warm-up solve builds the sparsity. J's pattern becomes the union over all rows, base case included.
- **`NRSystem::set_pv_pinned_buses`** then masks the `Q` row of the buses that are still PV in a given row to the identity — `J[q_row, vm_col] = 1`, rest of the row zero, residual zero — so `dVm == 0` and the magnitude holds at the setpoint.

Both halves reuse machinery that was already here. The reservation follows `free_vm_slack_buses_` line for line (`NRSystem.hpp:247-273`). The masking shares `masked_zero_pos_` / `masked_one_pos_` and the single `_recompute_mask_positions` pass with `set_masked_buses`, which `handle_disconnected_grid` has been driving per contingency all along — a pure value-level edit, sparsity untouched.

Measured on a 40-row generator N-1 sweep on case14:

```
nb_analyze=1  nb_factorize=1  nb_refactorize=115  nb_refactorize_failed=0
```

## A design note

This gives the concerned buses a `Vm` unknown plus a `Q` **equation**, not an explicit `Q` unknown. One row and one column per switchable bus instead of two; the generator's reactive output stays recoverable from the bus mismatch (`takes_q_residual_share`); and an explicit `Q_c` column is what *remote* control needs, where controller and controlled bus differ.

The slack **bus set** is untouched. The angle reference is a property of the batch, picked once, and moving it per row would raise `has_slack_participate_changed()` and cost the very re-analysis this avoids. Only the weights move.

The cost is one extra row and column per *switchable* bus, on every row of the sweep. A handful of candidate generators is negligible; masking every generator grows `dim(J)` by roughly the number of PV buses. Documented in `docs/scenario_sweep.rst`.

## Scope

Only generators regulating their **own** bus. A mask naming a remote controller, or a generator standing on a group-controlled bus (a remote generator, an SVC, an hvdc converter station), is refused by `compute()` with a message saying so: those live in the `VoltageControl` extension, whose own Jacobian columns and rows nothing here reserves or masks yet.

## Along the way

- `get_slack_weights_solver` and the new `get_slack_weights_solver_without` (the same weights, evaluated as if some generators were off) now share `_raw_slack_weights_solver`, so who participates and the disconnected-bus checks live in one place and the two cannot drift apart. The new one is `const`, which is what lets the per-row helpers run on several threads at once.
- `BaseBatchSolverSynch::get_linear_solver_stats()` exposes the batch's own solver counters. This is how the one-analysis claim above is actually asserted — the batch classes previously had no way to report them.

## Testing

New `lightsim2grid/tests/test_ScenarioSweep_gen_contingency.py` (11 tests, 16 subtests). The oracle throughout is a real one-off `ac_pf` on a copy of the grid with the generators actually deactivated, compared to 1e-8:

- generator N-1, one row per machine;
- a bus with two generators: one off keeps it PV, both off turns it PQ — and the magnitudes are asserted to agree in the first case and differ in the second, so the distinction is not silently vacuous;
- per-row injections combined with per-row outages;
- a combined line + generator row;
- the slack-participating generator;
- the `handle_disconnected_grid` masked path;
- 4-thread vs single-thread agreement;
- `nb_analyze == 1` across 12 rows — a regression here means a row is changing the pv/pq split;
- an all-`False` mask reproduces the plain sweep exactly (run over varying injections on purpose: an all-defaults sweep starts at its own answer and converges in zero iterations, which would compare two Jacobians neither of which was ever built);
- a remotely-regulating generator in the mask raises.

Regression runs: C++ suite 252/252 pass; the batch / Jacobian / solver-control Python suites pass (258 passed, 65 skipped). The full Python suite reports 1096 passed / 107 failed, and every one of those failures is this container missing grid2op test data (`data_test/test_case14.json`, `test_multi_chronics`), `pypowsybl`, or network access — none touches the changed code.

---

# 2. A `CLAUDE.md`, and un-ignoring it

Separate commit, unrelated to the feature — drop it if you would rather it went on its own.

Nothing in the repository said a commit needs a `Signed-off-by:` trailer, so nearly every PR from a Claude session went red on DCO for the same reason. (A PR template would not have helped: DCO validates the commit trailer, not the PR body.) `CLAUDE.md` now states the rule, including the part that makes it a rule rather than a formality — **a bot may not sign off**, so a Claude-written commit is authored by the human signing it and credits the model in `Co-Authored-By`. The two commits in the history signed off by `Claude <noreply@anthropic.com>` are called out as not to be copied.

This needed `.gitignore` line 311 dropped: `CLAUDE.md` was ignored (added in `4d37177`, among the venvs and local notes), and a session in a cloud container clones the repo fresh — so an ignored file simply is not there when it is needed.

The file also records the cold-start costs measured this session: the `eigen` / `SuiteSparse` / `Catch2` submodules are empty on a fresh clone and the build dies on `Eigen/Core`; the C++ tests configure from `src/tests`, not the top-level `CMakeLists.txt`; and a bare `pytest` run reports ~100 failures that are missing optional deps and absent grid2op test data rather than anything to do with the code.

https://claude.ai/code/session_016C8Y6Xp1bPzhUYnxYY7mYf

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -156,39 +156,18 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
 
 [1.0.1] 2026-xx-yy
 --------------------
-- [ADDED] ``ScenarioSweep.set_contingency_gens(mask)``: a third contingency axis, a
-  ``(n_simul, n_gen)`` boolean mask disconnecting generators row by row, alongside the
-  existing line and trafo masks. It takes the machine's active power out of that row's
-  injection (its reactive setpoint too when it does not regulate voltage), re-weights the
-  distributed slack without it, and -- when the LAST generator regulating its own bus goes
-  -- turns that bus from PV to PQ for the row. The slack **bus set** is untouched: the
-  angle reference is a property of the batch, picked once, and moving it per row would
-  raise ``has_slack_participate_changed()`` and cost the very re-analysis this is built to
-  avoid. Only generators regulating their own bus are in scope for now; a mask naming a
-  remote controller, or a generator standing on a group-controlled bus (a remote
-  generator, an SVC, an hvdc converter station), is refused by ``compute()`` with a message
-  saying so -- those live in the ``VoltageControl`` extension, whose own Jacobian columns
-  and rows nothing here reserves or masks yet.
-- [IMPROVED] the PV/PQ relabelling above costs **no extra symbolic factorization**, which
-  is the whole point: a sweep is fast because its labelling never changes, so one
-  ``analyze`` + one ``factorize`` serve every row and the rest are ``refactorize``. Rather
-  than re-analyzing per row, ``Base::set_switchable_vm_buses`` reserves a Vm unknown and a
-  Q equation for every bus that *can* flip -- so the sparsity is the union over the rows,
-  the "n" warm-up solve included -- and ``NRSystem::set_pv_pinned_buses`` then masks the Q
-  row of the buses that are still PV to the identity, freezing ``dVm`` at 0 so the
-  magnitude holds at the setpoint. Both reuse machinery that was already there: the
-  reservation follows ``free_vm_slack_buses_`` exactly, and the masking shares
-  ``masked_zero_pos_`` / ``masked_one_pos_`` and the single ``_recompute_mask_positions``
-  pass with ``set_masked_buses``, which ``handle_disconnected_grid`` has been driving per
-  contingency all along. Note the deliberate shape: a Vm unknown plus a Q *equation*, not
-  an explicit Q unknown -- the generator's reactive output is still recoverable from the
-  bus mismatch (``takes_q_residual_share``), and an explicit ``Q_c`` column is what
-  *remote* control needs, where controller and controlled bus differ. The cost is one row
-  and one column per switchable bus, on every row of the sweep.
+- [ADDED] ``ScenarioSweep.set_contingency_gens(mask)``: a per-step generator contingency mask,
+  next to the line and trafo ones. It removes the generator's injection from that step, re-weights
+  the distributed slack without it, and turns its bus PV -> PQ when it was the last machine
+  regulating it. Only generators regulating their own bus are supported for now.
+- [IMPROVED] the PV -> PQ relabelling above costs no extra symbolic factorization: every bus that
+  can change role is given a Vm unknown and a Q equation once, and each step masks the Q row of
+  the buses that are still PV. A sweep still pays one ``analyze`` + one ``factorize``.
 - [IMPROVED] ``GeneratorContainer::get_slack_weights_solver`` and the new
-  ``get_slack_weights_solver_without`` (same weights, evaluated as if some generators were
-  off) share ``_raw_slack_weights_solver``, so the rule for who participates and the
-  disconnected-bus checks live in one place and the two cannot drift apart.
+  ``get_slack_weights_solver_without`` share their implementation, so the rule for who
+  participates in the slack cannot drift between them.
+- [ADDED] ``get_linear_solver_stats()`` on the batch classes, to check a sweep really does reuse
+  its factorization.
 - [FIXED] a grid using voltage control that the selected algorithm cannot implement -- a
   generator or an hvdc converter station regulating a bus other than its own, several
   machines regulating one bus, or a voltage-mode SVC -- was **solved anyway, to a wrong
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -156,6 +156,39 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
 
 [1.0.1] 2026-xx-yy
 --------------------
+- [ADDED] ``ScenarioSweep.set_contingency_gens(mask)``: a third contingency axis, a
+  ``(n_simul, n_gen)`` boolean mask disconnecting generators row by row, alongside the
+  existing line and trafo masks. It takes the machine's active power out of that row's
+  injection (its reactive setpoint too when it does not regulate voltage), re-weights the
+  distributed slack without it, and -- when the LAST generator regulating its own bus goes
+  -- turns that bus from PV to PQ for the row. The slack **bus set** is untouched: the
+  angle reference is a property of the batch, picked once, and moving it per row would
+  raise ``has_slack_participate_changed()`` and cost the very re-analysis this is built to
+  avoid. Only generators regulating their own bus are in scope for now; a mask naming a
+  remote controller, or a generator standing on a group-controlled bus (a remote
+  generator, an SVC, an hvdc converter station), is refused by ``compute()`` with a message
+  saying so -- those live in the ``VoltageControl`` extension, whose own Jacobian columns
+  and rows nothing here reserves or masks yet.
+- [IMPROVED] the PV/PQ relabelling above costs **no extra symbolic factorization**, which
+  is the whole point: a sweep is fast because its labelling never changes, so one
+  ``analyze`` + one ``factorize`` serve every row and the rest are ``refactorize``. Rather
+  than re-analyzing per row, ``Base::set_switchable_vm_buses`` reserves a Vm unknown and a
+  Q equation for every bus that *can* flip -- so the sparsity is the union over the rows,
+  the "n" warm-up solve included -- and ``NRSystem::set_pv_pinned_buses`` then masks the Q
+  row of the buses that are still PV to the identity, freezing ``dVm`` at 0 so the
+  magnitude holds at the setpoint. Both reuse machinery that was already there: the
+  reservation follows ``free_vm_slack_buses_`` exactly, and the masking shares
+  ``masked_zero_pos_`` / ``masked_one_pos_`` and the single ``_recompute_mask_positions``
+  pass with ``set_masked_buses``, which ``handle_disconnected_grid`` has been driving per
+  contingency all along. Note the deliberate shape: a Vm unknown plus a Q *equation*, not
+  an explicit Q unknown -- the generator's reactive output is still recoverable from the
+  bus mismatch (``takes_q_residual_share``), and an explicit ``Q_c`` column is what
+  *remote* control needs, where controller and controlled bus differ. The cost is one row
+  and one column per switchable bus, on every row of the sweep.
+- [IMPROVED] ``GeneratorContainer::get_slack_weights_solver`` and the new
+  ``get_slack_weights_solver_without`` (same weights, evaluated as if some generators were
+  off) share ``_raw_slack_weights_solver``, so the rule for who participates and the
+  disconnected-bus checks live in one place and the two cannot drift apart.
 - [FIXED] a grid using voltage control that the selected algorithm cannot implement -- a
   generator or an hvdc converter station regulating a bus other than its own, several
   machines regulating one bus, or a voltage-mode SVC -- was **solved anyway, to a wrong
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -162,12 +162,15 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
   regulating it. Only generators regulating their own bus are supported for now.
 - [IMPROVED] the PV -> PQ relabelling above costs no extra symbolic factorization: every bus that
   can change role is given a Vm unknown and a Q equation once, and each step masks the Q row of
-  the buses that are still PV. A sweep still pays one ``analyze`` + one ``factorize``.
+  the buses that are still PV. Each algorithm still analyzes once (so once per thread when
+  multi-threaded). It is not free though: J gains a row and a column per bus that can change
+  role, on every step.
 - [IMPROVED] ``GeneratorContainer::get_slack_weights_solver`` and the new
   ``get_slack_weights_solver_without`` share their implementation, so the rule for who
   participates in the slack cannot drift between them.
-- [ADDED] ``get_linear_solver_stats()`` on the batch classes, to check a sweep really does reuse
-  its factorization.
+- [ADDED] ``get_linear_solver_stats()`` / ``get_linear_solver_stats_per_algo()`` on the batch
+  classes, to check a sweep really does reuse its factorization. The first sums over every
+  worker thread, the second keeps them apart.
 - [FIXED] a grid using voltage control that the selected algorithm cannot implement -- a
   generator or an hvdc converter station regulating a bus other than its own, several
   machines regulating one bus, or a voltage-mode SVC -- was **solved anyway, to a wrong
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,4 +1,6 @@
-# Working in this repository
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
 
 ## Commits: the DCO signoff is mandatory
 
@@ -33,7 +35,7 @@ Rules:
   followed by `git push --force-with-lease`. This is the case that actually bites: the
   mistake is usually noticed only after CI has run.
 
-## Building from a fresh clone
+## Build
 
 The vendored dependencies are git submodules and start empty — a build fails with
 `Eigen/Core: No such file or directory` until they are fetched:
@@ -42,7 +44,7 @@ The vendored dependencies are git submodules and start empty — a build fails w
 git submodule update --init --depth 1 eigen SuiteSparse Catch2   # Catch2 only for the C++ tests
 ```
 
-The build is driven by scikit-build-core. Installing its backend first makes repeated
+The build is driven by scikit-build-core. Installing the backend first makes repeated
 builds much faster than re-resolving it every time:
 
 ```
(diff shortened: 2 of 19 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,215 @@
...
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Commits: the DCO signoff is mandatory
+
+Every commit must end with a `Signed-off-by:` trailer naming **a human**, or the DCO check
+fails the pull request. This is the single most common reason a PR here is red.
+
+```
+Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
```
BDonnot: I checked and it is Assisted-by: LLM [TOOL1] [TOOL2]
See https://docs.kernel.org/process/coding-assistants.html

CLAUDE.md
```diff
@@ -0,0 +1,215 @@
...
+```
+git submodule update --init --depth 1 eigen SuiteSparse Catch2   # Catch2 only for the C++ tests
+```
+
+The build is driven by scikit-build-core. Installing the backend first makes repeated
+builds much faster than re-resolving it every time:
+
+```
+pip install "scikit-build-core>=0.5.0" pybind11
+pip install -e . --no-build-isolation
```
BDonnot: Build why --no-build-isolation? The -e is sufficient most of the time

CLAUDE.md
```diff
@@ -0,0 +1,215 @@
...
+this reason.
+
+## Tests
+
+Python, from `lightsim2grid/tests` — the tests import each other by bare module name, so the
+working directory matters:
+
+```
+cd lightsim2grid/tests
+python -m pytest .                                    # everything
```
BDonnot: Tests are mainly unittest (purest might work but that's not what is commonly used).
Tests requires Grid2Op to be installed from source, otherwise some will not pass

CLAUDE.md
```diff
@@ -0,0 +1,215 @@
...
+ctest --test-dir build_tests --output-on-failure
+ctest --test-dir build_tests -R "some test name"      # one test; they are named by sentence
+```
+
+## Architecture
+
+Three layers: a standalone C++ core library (`src/core/`, buildable and usable with no
+Python at all — see `docs/cpp_library.rst`), pybind11 bindings (`src/bindings/python/`),
+and the Python package (`lightsim2grid/`) whose headline product is `LightSimBackend`, a
+grid2op backend.
```
BDonnot: And other entry points : the algorithms (eg NR_KLU), the TimeSeries, the ScenarioSweep and the ContingencyAnalysis

CLAUDE.md
```diff
@@ -0,0 +1,215 @@
...
+### `LSGrid` is the hub
+
+`src/core/LSGrid.{hpp,cpp}` owns everything: the element containers, the grid↔solver bus
+labelling, and the construction of the solver input (`Ybus`, `Sbus`, the pv/pq split, the
+slack). `build_solver_input` / `build_dc_solver_input` produce a `SolverBusLayout` the
+algorithms consume; `compute_results` publishes flows and injections back onto the
+containers. It is a large file — start from the method you need rather than reading it
+through.
+
+### Two bus numbering spaces, and they are different C++ types
```
BDonnot: There are 3 bus numbering total.

You forget the "local" bus labelling, the one relative to a give substitation, between 1 and n_max_busbar_per_sub

CLAUDE.md
```diff
@@ -0,0 +1,215 @@
...
+  backwards: a row is an equation (a P or Q mismatch), a column is an unknown (a Δθ or Δ|V|).
+- **`build_J_sparsity`** (`NRSystem.tpp`) is entirely ledger-driven: one pass over the Ybus
+  nonzeros emits every dS-derived entry for all components at once. Give a bus a row/column
+  in the ledger and the sparsity follows automatically.
+- **Value-level row masking** keeps the sparsity fixed across scenarios:
+  `set_masked_buses` (P and Q rows → identity, for buses a contingency stranded) and
+  `set_pv_pinned_buses` (Q row only, for a bus that is PV in this scenario but PQ in
+  another). Both resolve to positions in `J_.valuePtr()` and rewrite values only, so the
+  symbolic factorization survives. This is the pattern to reuse for anything that changes a
+  bus's role per solve.
```
BDonnot, line 187: For example in batch algorithm (not needed in regular solve called one after the other)

CLAUDE.md
```diff
@@ -0,0 +1,215 @@
...
+**Deprecated names still in the tree**, kept working but not for new code: `GridModel` →
+`LSGrid`, `lightsim2grid.gridmodel` → `lightsim2grid.network`, `lightsim2grid.solver` →
+`lightsim2grid.algorithm`, `SecurityAnalysis` → `ContingencyAnalysis`.
+
+## Changelog
+
+`CHANGELOG.rst`, under the current development version at the top — **not** under `[TODO]`,
+which is a list of open problems rather than a release. Entries are prefixed `[ADDED]`,
+`[FIXED]`, `[IMPROVED]` and explain *why*, at length where the reasoning is not obvious from
+the diff. Match the surrounding style; a one-line entry is out of place here.
```
BDonnot: Please read the Changelog entries from before Claude. Today's Changelog is too verbose !

BDonnot:
All seven review comments addressed in `11923db`.

I could not reply on the threads themselves: the review is still **pending** (unsubmitted), and since this session authenticates as the same account GitHub refuses a second pending review to reply into. Submitting your review for you seemed wrong, so here is one comment instead — the threads are yours to resolve when you submit.

| # | comment | what changed |
|---|---|---|
| 1 | `Assisted-by:`, not `Co-Authored-By:` | Example and rule updated: `Assisted-by: Claude Code (claude-opus-5)`, naming tool and model, and stating the assistant is never an author. Kernel doc linked. I could not open docs.kernel.org (blocked by this container's egress proxy), so I followed your wording rather than guessing the syntax. |
| 2 | why `--no-build-isolation`? | Demoted. `pip install -e .` is now the stated norm; the isolation flag is a footnote for when you are iterating on C++ and do not want the backend re-resolved each time. |
| 3 | tests are unittest; grid2op must be from source | Both fixed, and the second one matters more than a doc nit — it is the actual explanation for the ~100 failures the file previously dismissed as "environmental". The commands are now `python -m unittest discover` / `... test_ScenarioSweep.TestScenarioSweepCPP.test_...`, checked by running them. |
| 4 | other entry points | Added: the algorithms (`NR_KLU` etc. via `lightsim2grid.algorithm`) and `TimeSeries` / `InjectionSweep` / `ContingencyAnalysis` / `ScenarioSweep`, described as standing on their own rather than as things behind the backend. |
| 5 | three bus labellings, not two | Corrected. `LocalBusId` was missing entirely — the **1-based** busbar within a substation, from 1 to `n_busbar_per_sub`, which is what a grid2op topology action speaks, converted via `SubstationContainer::local_to_gridmodel`. The 1-based-ness is called out, since the note in `SubstationContainer.hpp` records it going wrong before. |
| 6 | masking is for the batch algorithms | Scoped: noted that a plain solve, or solves run one after another, has no need of it. |
| 7 | changelog too verbose | The guidance was backwards — it told a future session to write *at length*, which is how the recent entries got the way they are. Now: one to four lines, ~20 words, with the `[0.13.x]` and earlier blocks named as the register to aim at, and the reasoning explicitly pushed to the commit message rather than the changelog. This change's own entry is cut from 408 words to 145. |

One thing left inconsistent, for you to decide: the three earlier commits on this branch still carry `Co-Authored-By:`, so the branch contradicts the file it adds. Normalising them means a force-push over commits you have already read, which did not seem worth doing unasked.

---

CHANGELOG.rst
```diff
@@ -156,6 +156,18 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
...
+  the distributed slack without it, and turns its bus PV -> PQ when it was the last machine
+  regulating it. Only generators regulating their own bus are supported for now.
+- [IMPROVED] the PV -> PQ relabelling above costs no extra symbolic factorization: every bus that
+  can change role is given a Vm unknown and a Q equation once, and each step masks the Q row of
+  the buses that are still PV. A sweep still pays one ``analyze`` + one ``factorize``.
+- [IMPROVED] ``GeneratorContainer::get_slack_weights_solver`` and the new
+  ``get_slack_weights_solver_without`` share their implementation, so the rule for who
+  participates in the slack cannot drift between them.
+- [ADDED] ``get_linear_solver_stats()`` on the batch classes, to check a sweep really does reuse
+  its factorization.
```
BDonnot: Does it work correctly even in case of multi threading ? Like one analyze per thread, one factorization per thread and then only refactorization ?

lightsim2grid/tests/test_ScenarioSweep_gen_contingency.py
```diff
@@ -0,0 +1,330 @@
...
+                                   rtol=1e-10, atol=1e-10)
+
+    def test_with_handle_disconnected_grid(self):
+        """The masked code path takes the generator contingencies too.
+
+        ``handle_disconnected_grid`` runs its own per-row loop (_run_range_masked), so a
+        generator contingency has to be applied there as well -- including on rows whose
+        line contingency also strands part of the grid, where the bus mask and the PV
+        pinning have to compose.
+        """
```
BDonnot, line 328: A similar test should be added for the interaction with the previous PR : disconnection of a transfo of a remote regulating gen (gen remote regulates the other side of the transfo).
Probably needs a custom grid for that

src/bindings/python/binding_batch.cpp
```diff
@@ -58,6 +58,13 @@ void bind_batch_sweep_common(py::class_<T> & cls)
         .def("amps_computation_time", &T::amps_computation_time, DocTimeSeries::amps_computation_time.c_str())
         .def("thread_init_time", &T::thread_init_time, DocTimeSeries::thread_init_time.c_str())
         .def("nb_solved", &T::nb_solved, DocTimeSeries::nb_solved.c_str())
+        .def("get_linear_solver_stats", &T::get_linear_solver_stats,
+             "Linear-solver counters of this batch's own algorithm (nb_analyze, "
+             "nb_factorize, nb_refactorize, ...). A batch keeps its labelling fixed "
+             "across rows, so nb_analyze should read 1 for the whole compute() however "
```
BDonnot: 1 per thread used

src/bindings/python/binding_batch.cpp
```diff
@@ -229,6 +236,25 @@ void bind_batch(py::module_& m) {
...
+             "Unlike the two branch masks this does not edit Ybus -- a generator has no "
+             "admittance. It removes the generator's active power (and, if it does not "
+             "regulate voltage, its reactive setpoint) from that step's injection, "
+             "re-weights the distributed slack without it, and -- when the LAST "
+             "generator regulating its own bus is taken out -- turns that bus from PV "
+             "to PQ for the step, so its voltage magnitude is solved for instead of "
+             "held at the setpoint. The lost MW is picked up by the slack; use "
+             "modify_gen_p to express a redispatch instead.\n\n"
+             "The PV/PQ relabelling costs no extra symbolic factorization: every bus "
+             "that can flip is given a voltage-magnitude unknown and a reactive "
```
BDonnot, line 260: Write that it's not cost free. J is larger

src/core/batch_algorithm/BaseBatchSolverSynch.hpp
```diff
@@ -178,6 +178,22 @@ class LS2G_API BaseBatchSolverSynch : protected BaseConstants
...
+         * built once and every row after the first only refactorizes. A test (or a
+         * profiling session) reading nb_analyze > 1 is looking at a row that changed
+         * the pv/pq split, which is a bug, not a tuning question.
+         *
+         * Single-threaded runs only tell the whole story: with nb_thread > 1 each
+         * worker owns its own algorithm (see _compute_threaded) and these are the
+         * member one's counters -- the "n" warm-up solve, plus whatever range the
+         * calling thread took.
+         */
+        LinearSolverStats get_linear_solver_stats() const {return _algo.get_linear_solver_stats();}
```
BDonnot: Not correct in case of multi threading. In this case reports only the stats of the main thread

CLAUDE.md
```diff
@@ -0,0 +1,236 @@
+# CLAUDE.md
```
BDonnot, line 1: Missing a ref to the different ways to init a LSgrid

CLAUDE.md
```diff
@@ -0,0 +1,236 @@
...
+|---|---|---|---|
+| `TimeSeries` | no | yes | the previous row's solution |
+| `InjectionSweep` | no | yes | the same seed |
+| `ContingencyAnalysis` | yes | no | the same seed |
+| `ScenarioSweep` | yes | yes | the same seed |
+
+Members that only make sense for some instantiations are SFINAE-gated on
+`YbusPolicy::supports_contingency` / `SbusPolicy::supports_vary` rather than split into
+subclasses. The performance premise of all four is that **the pv/pq labelling handed to the
+solver never changes between rows**, so the batch pays one `analyze` + one `factorize` and
```
BDonnot: We'll this isn't true anymore. And that's precisely what this pr is about.

It should be rephrased with something like:

The sparsity pattern of the Jacobian is fixed during all the elements in the batch.

And maybe have a note for some PV/PQ switch (this PR), for the disconnection of trafo with a single generator (the previous one) and for the disconnection of the smallest element if a contingency split the grid in multiple part.
I expect to land some other features along this line at some point.

src/core/batch_algorithm/BaseBatchSolverSynch.hpp
BDonnot: You were right, and documenting the limitation was the wrong fix — it under-reported the work and could not answer the question the accessor exists to answer.

Each worker's counters are now harvested after the joins into `_thread_solver_stats_` and summed in, so `get_linear_solver_stats()` covers the whole `compute()`. Added `get_linear_solver_stats_per_algo()` alongside, because the sum on its own cannot distinguish "every algorithm analyzed once" from "one algorithm analyzed four times" — and that distinction is the whole invariant.

---

CHANGELOG.rst
BDonnot: Yes — one analyze and one factorize **per algorithm**, and each worker owns its own, so multi-threaded it is one per worker plus the member algorithm that solves the "n" warm-up case. It works because `init_thread` hands each fresh algorithm the switchable set *before* its first `build_J_sparsity`, so every worker builds the same enlarged pattern and refactorizes from there.

That was reasoning, not measurement, when you asked — so it is now measured. `test_one_analysis_per_algorithm_when_threaded` runs 24 rows on 4 threads and asserts `nb_analyze <= 1` for each of the 5 algorithms individually, plus that the aggregate equals the sum. The changelog line is reworded to say "each algorithm still analyzes once (so once per thread when multi-threaded)" rather than implying one for the whole batch.

---

CLAUDE.md
BDonnot: Agreed, and it is the sharpest correction in the round — the file stated as an invariant the very thing this PR breaks.

Rephrased to: **the sparsity pattern of the Jacobian is fixed for every element of the batch**, with the explicit note that this does *not* forbid a row changing a bus's role, as long as it does so by rewriting values inside a pattern reserved up front. The three mechanisms are listed as you suggested — PV↔PQ switching (this PR), a trafo disconnection stranding the single generator regulating through it (#192), and masking out the smaller side when a contingency splits the grid — framed as examples of a pattern rather than a closed list, since you expect more along the same line.

What is still genuinely forbidden is spelled out separately: a different pv/pq vector or a different slack *set* handed to the solver per row, which raises `has_pv_changed()` / `has_slack_participate_changed()` and forces a fresh `analyze`.

---

lightsim2grid/tests/test_ScenarioSweep_gen_contingency.py
BDonnot: Added, and thanks for the pointer to #192 — it changed how I built it.

I had started hunting for a grid where a generator can regulate across a transformer and the far side stays connected, and ended up on pandapower `case118` (gen 24 / bus 62 / trafo 4), because on the grid2op case14 the only such configuration also islands the far bus, so both sides merely diverge and the comparison asserts nothing.

Looking at #192 instead: `test_voltage_control_batch.py` already has exactly the fixture — pandapower `case14` with generator 3 (bus 7) regulating bus 9 — so I dropped case118 and reused it, and the two files now describe the same grid. On it, trafo 4 (buses 6–8) sits on the control path and stays solvable, while trafo 3 (buses 6–7) puts the controller's own bus behind the disconnection.

Three tests, each against a one-off solve of the same case: the transformer on the control path; the same row disconnecting that transformer **and** a generator (the interaction you were really asking about); and the controller stranded behind trafo 3, asserting the batch and the one-off solve agree on whether it solves at all. The C++ side of the stranding itself is already covered by `test_batch_voltage_control.cpp`, so I did not duplicate it — what is new here is that the generator-contingency axis composes with it.

---

src/bindings/python/binding_batch.cpp
BDonnot: Done. The docstring said the relabelling costs no extra symbolic factorization and stopped there — true and misleading together. It now spells out that the reserved unknowns and equations make J bigger for **every** step, whether or not that step disconnects anything: one extra row and column per bus that can change role, negligible for a handful of candidate generators, roughly `n_pv` extra if you mask every generator on the grid — and every factorization and solve pays for it. The changelog entry says the same in one line.

---

src/bindings/python/binding_batch.cpp
BDonnot: Fixed — the binding now says the invariant is 1 analyze **per algorithm used**: 1 single-threaded, and with `nb_thread > 1` one per worker plus the member algorithm that solves the "n" warm-up case. It also points at the new `get_linear_solver_stats_per_algo()` for when the sum is not enough to tell what happened.

---

CLAUDE.md
BDonnot: Added at the top of the `LSGrid` section: a grid is never built by hand, it is loaded, and `lightsim2grid/network/` holds one converter per source — pandapower, pypowsybl/iidm, MATPOWER, PowerModels.jl — plus `load_binary` for the fast binary format, with pointers to `docs/network.rst` and `docs/binary_serialization.rst`. It also notes which path `LightSimBackend` takes depending on the environment, and that when a bug looks like bad input the converter is the place to look before the solver.

---

CLAUDE.md
BDonnot: Addressed in `11923db` — the masking bullet now says it is a *batch-algorithm* concern, and that a plain solve, or solves called one after another, has no need of it.

---

BDonnot (commented):
Ok

PR #194, opened 2026-09-08
Load a grid2op environment from a MATPOWER case file

BDonnot:
> **Based on #193.** The branch starts from that PR's head, so until #193 lands this diff shows its commits too. **The commit to review here is the last one**, `Load a grid2op environment from a MATPOWER case file`. Rebase or merge order is yours; nothing here touches the ScenarioSweep work.

# What

`LightSimBackend` gains a third `loader_method`, `"matpower"`, next to `"pandapower"` and `"pypowsybl"`: an environment can now ship its powergrid as the `grid.m` a MATPOWER case is distributed as (or the `grid.mat` matpower saves).

```python
import grid2op
from lightsim2grid import LightSimBackend

env = grid2op.make(path_to_env_with_a_grid_m,
                   backend=LightSimBackend(loader_method="matpower"))
```

`supported_grid_format` is `("m", "mat")`, so `grid2op.make` finds the file on its own. It goes straight through `init_from_matpower` and never builds a pandapower or a pypowsybl network on the way. `loader_kwargs` takes `grid` (an already parsed case, same contract as the pypowsybl loader's own `grid`), `n_busbar_per_sub` and `double_bus_per_sub`.

MATPOWER describes numbered buses and nothing else, so the choices the loader has to make are the same ones the pypowsybl loader makes with `use_buses_for_sub=True`:

- **substations**: one grid2op substation per matpower bus; how many busbar sections each gets is the caller's (the extra ones start empty and deactivated).
- **names**: grid2op's default ones (`sub_0`, `load_1_0`, `gen_5_3`, `0_4_1`), since matpower names nothing.
- **thermal limits**: left open for the environment's `config.py` — matpower's `RATE_A` is a branch MVA rating, and 0 in a fair share of the published cases.
- **generators**: several `mpc.gen` rows on one bus stay independent machines, as `init_from_matpower` already did.

# Along the way: where the substation ids come from

Writing this the obvious way would have made a **third** copy, in the backend, of "which substation is this element in". It is a property of the file, so it now lives where the file is read:

- `init_from_powermodels` — and with it `init_from_matpower` and `init_from_pf_delta` — sets `set_gen_to_subid` and friends on the grid it returns, as `init_from_pypowsybl` already did.
- `LightSimBackend` reads that back with one `_aux_read_subid_from_grid()`, **which the pypowsybl path now uses too**, replacing the block that re-derived the same values from the dataframes the loader hands back.

The value stored is the bus each element was **built** on, not `el.bus_id`: an element the source file declares out of service reports `bus_id == -1`, while the substation it belongs to is a property of the grid rather than of its status. That is why the `_aux_add_*` helpers now return their bus array instead of the backend reading it off the finished grid.

`init_from_pandapower` is deliberately left alone. pandapower has no substation notion of its own and which of its buses make one up is grid2op's pandapower backend's decision, not the file's — so there the backend stays the authority. Said explicitly in `docs/network.rst`.

`_aux_setup_right_after_grid_init` used `_orig_grid_pypowsybl is None` to mean "the loader did not lay the substations out, so set `n_sub` myself". That is now an explicit `_grid_laid_out_by_loader` flag: matpower lays them out too, and the old test would have had the backend overwrite a correct `n_sub`.

# One thing beyond the ask

A matpower bus whose `BASE_KV` is 0 — matpower's way of saying a case never leaves per unit, and several published ones do — was **refused outright** by `SubstationContainer`, so those cases could not be loaded at all, before this PR and independently of it. It is now read as 1 kV with a warning, so every voltage reported for it is numerically its per-unit value. Flagging it because it changes `init_from_matpower` behaviour, not just the new backend path; happy to split it out.

# Testing

`lightsim2grid/tests/case_14_matpower/` is a real environment. Its `grid.m` is the IEEE 14 bus case — the data `pandapower.networks.case14()` carries, with the published generator limits restored, the `BASE_KV` column filled in (a grid2op backend reports kV), and the reference-bus machine moved to the **last** `mpc.gen` row so that no test passes for the wrong reason. Plus a `config.py` with thermal limits, so the environment starts around rho = 0.67 rather than at 0.

`test_backend_matpower.py`, **74 tests**:

- the shape of the loaded grid, and its nominal voltages;
- that the substation ids come from the loader, survive an out-of-service element, and that a topology action lands on the bus they imply;
- AC and DC `runpf`, and agreement with a bare `init_from_matpower` + `ac_pf`;
- **the whole solution against pandapower** solving its own copy of `case14()` — bus magnitudes and angles, loads, generation, line and transformer flows, the shunt. An independent converter and an independent solver on the same grid, which is what makes this an oracle rather than a tautology;
- a `.mat` file, and each loader kwarg;
- a grid2op environment through `reset`, `step`, a topology action, a line disconnection, `copy` and a `Runner`;
- grid2op's own `AAATestBackendAPI` suite. One of its tests, `test_01load_grid`, is skipped: it asserts the shape of grid2op's `educ_case14_storage` (6 generators, 2 storage units), as its own docstring warns it does. Every other test of the suite runs.

`test_init_from_matpower.py` gains the loader-level half: the substation ids of every element type, including with several busbar sections, and the `BASE_KV` fallback.

Ran, from `lightsim2grid/tests`: `test_backend_matpower`, `test_init_from_matpower`, `test_LSGrid_pf_delta`, `test_backend_pypowsybl`, `test_init_from_pypowsybl`, `test_correct_substation_pypowsybl`, `test_storage_pypowsybl`, `test_hvdc_pypowsybl`, `test_voltage_control_pypowsybl`, `test_LSGrid_pandapower`, `test_check_grid`, `test_n_busbar_per_sub`, `test_binary_serialization`, `test_gridmodel_change_bus` — **493 tests, all passing**.

`CLAUDE.md` also records that the full python suite takes over two hours, so it is not a way to check a change here.

https://claude.ai/code/session_01C4L2EgRWRdUREGHrUVh23R

---
_Generated by [Claude Code](https://claude.ai/code/session_01C4L2EgRWRdUREGHrUVh23R)_

### DIFF
--- a/docs/network.rst
+++ b/docs/network.rst
@@ -37,20 +37,26 @@ An `LSGrid` can be built from several source formats, each with a dedicated ``in
 
 See the "Detailed documentation" section below for the full signature and caveats of each.
 
-A grid built by ``init_from_pypowsybl``, ``init_from_matpower``, ``init_from_powermodels``
-or ``init_from_pf_delta`` also knows **which substation / voltage level each of its elements
-belongs to** (``sub_id`` of an ``*Info`` object, set by the loader through
-``set_gen_to_subid`` and friends). That is a property of the source file, so the loader is
-where it is worked out -- and, unlike ``bus_id``, it stays available for an element the
-source declares out of service. It is what
+Whichever loader built it, the resulting ``LSGrid`` also knows **which substation / voltage
+level each of its elements belongs to** (``sub_id`` of an ``*Info`` object, set by the loader
+through ``set_gen_to_subid`` and friends). That is a property of the source file, so the
+loader is where it is worked out -- and, unlike ``bus_id``, it stays available for an element
+the source declares out of service. It is what
 :func:`lightsim2grid.network.LSGrid.update_topo` needs to turn "busbar 2 of my substation"
-into a global bus id, so such a grid is ready for grid2op-style topology actions as it comes.
+into a global bus id, so a loaded grid is ready for grid2op-style topology actions as it
+comes.
 
-``init_from_pandapower`` is the exception: pandapower has no substation notion of its own,
-and which of its buses make up a grid2op substation is decided by grid2op's own pandapower
-backend rather than by the file. There, it is
-:class:`lightsim2grid.lightSimBackend.LightSimBackend` that sets the substation ids, from
-what grid2op says.
+None of these formats has a busbar section within a bus, so the rule is the same everywhere:
+**one substation per source bus, everything on busbar section 1**, and ``n_busbar_per_sub``
+sections allocated per substation (the extra ones start empty and deactivated). A grid built
+from pandapower is, in this respect, indistinguishable from one built from any other source.
+
+.. note::
+    ``init_from_pandapower`` takes an ``init_subid`` argument for one caller only:
+    :class:`lightsim2grid.lightSimBackend.LightSimBackend` passes ``False``, because its
+    pandapower path is built on grid2op's own pandapower backend, which has been the one
+    deciding what a substation is for as long as it has existed. That is history, not
+    design -- nothing else should need the flag.
 
 .. note::
     If you want to use one of these formats as the powergrid of a **grid2op environment**,
--- a/docs/network.rst
+++ b/docs/network.rst
@@ -37,6 +37,28 @@ An `LSGrid` can be built from several source formats, each with a dedicated ``in
 
 See the "Detailed documentation" section below for the full signature and caveats of each.
 
+A grid built by ``init_from_pypowsybl``, ``init_from_matpower``, ``init_from_powermodels``
+or ``init_from_pf_delta`` also knows **which substation / voltage level each of its elements
+belongs to** (``sub_id`` of an ``*Info`` object, set by the loader through
+``set_gen_to_subid`` and friends). That is a property of the source file, so the loader is
+where it is worked out -- and, unlike ``bus_id``, it stays available for an element the
+source declares out of service. It is what
+:func:`lightsim2grid.network.LSGrid.update_topo` needs to turn "busbar 2 of my substation"
+into a global bus id, so such a grid is ready for grid2op-style topology actions as it comes.
+
+``init_from_pandapower`` is the exception: pandapower has no substation notion of its own,
+and which of its buses make up a grid2op substation is decided by grid2op's own pandapower
+backend rather than by the file. There, it is
+:class:`lightsim2grid.lightSimBackend.LightSimBackend` that sets the substation ids, from
+what grid2op says.
+
+.. note::
+    If you want to use one of these formats as the powergrid of a **grid2op environment**,
+    you do not call these functions yourself:
+    :class:`lightsim2grid.lightSimBackend.LightSimBackend` does it, through its
+    ``loader_method`` argument (``"pandapower"``, ``"pypowsybl"`` or ``"matpower"``). See
+    :ref:`lightsimbackend_matpower` for the MATPOWER case.
+
 For example, you can init it from a pandapower grid like (NOT RECOMMENDED, though sometimes needed):
 
 .. code-block:: python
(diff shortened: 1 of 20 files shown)


docs/network.rst
```diff
@@ -37,6 +37,28 @@ An `LSGrid` can be built from several source formats, each with a dedicated ``in
...
+A grid built by ``init_from_pypowsybl``, ``init_from_matpower``, ``init_from_powermodels``
+or ``init_from_pf_delta`` also knows **which substation / voltage level each of its elements
+belongs to** (``sub_id`` of an ``*Info`` object, set by the loader through
+``set_gen_to_subid`` and friends). That is a property of the source file, so the loader is
+where it is worked out -- and, unlike ``bus_id``, it stays available for an element the
+source declares out of service. It is what
+:func:`lightsim2grid.network.LSGrid.update_topo` needs to turn "busbar 2 of my substation"
+into a global bus id, so such a grid is ready for grid2op-style topology actions as it comes.
+
+``init_from_pandapower`` is the exception: pandapower has no substation notion of its own,
```
BDonnot: Neither does Matpower... Apply same method here: buses are substation, put everything on busbar 1 and add as many busbars per substation as given.

To not break LightsimBackend, have a flag, named like 'init_subid' true by default and set this to False when using the "init_from_pandapower" in LightsimBackend so that the pandapower backend is still responsible for these.

Beside "historical reasons" (explaining why a pandapower backend was used here) a grid init from pandapower should be indistinguishable from a grid init from another sources.
"Historical reasons" / backward compat is what forces this flag.

lightsim2grid/lightSimBackend.py
```diff
@@ -102,6 +102,12 @@ class LightSimBackend(Backend):
         # "grid"
     }
     
+    KEYS_MATPOWER_LOADER = {
+        "grid",
+        "double_bus_per_sub",
```
BDonnot: This is legacy. Remove it now that 1, 3 , 4 etc busbar per substation are supported by grid2op

lightsim2grid/lightSimBackend.py
```diff
@@ -1142,6 +1180,150 @@ def _load_grid_pypowsybl(self, path=None, filename=None):
...
+        self._aux_read_subid_from_grid()
+        
+        # the names: matpower has none, so use grid2op's default ones (and give them
+        # back to the grid, so a LSGrid error message names the same element grid2op does)
+        self.name_sub = np.array([f"sub_{i}" for i in range(self.n_sub)]).astype(str)
+        self.name_load = np.array([f"load_{sub_id}_{id_obj}"
+                                   for id_obj, sub_id in enumerate(self.load_to_subid)]).astype(str)
+        self.name_gen = np.array([f"gen_{sub_id}_{id_obj}"
+                                  for id_obj, sub_id in enumerate(self.gen_to_subid)]).astype(str)
+        self.name_line = np.array([f"{sub1}_{sub2}_{id_obj}" for id_obj, (sub1, sub2)
```
BDonnot: How about using grid2op default names, there is a function for that I think

lightsim2grid/network/from_matpower/_mpc_to_powermodels.py
```diff
@@ -32,13 +32,29 @@ def mpc_to_powermodels(bus, gen, branch, dcline, baseMVA) -> dict:
...
+                      "Set the BASE_KV column of the case if you want voltages in actual kV.")
+
     for i in range(bus.shape[0]):
         bus_i = int(bus[i, BUS_I])
         key = str(i + 1)
         network["bus"][key] = {
             "bus_i": bus_i,
             "bus_type": int(bus[i, BUS_TYPE]),
-            "base_kv": float(bus[i, BASE_KV]),
+            "base_kv": float(bus[i, BASE_KV]) if float(bus[i, BASE_KV]) > 0. else 1.0,
```
BDonnot: I strongly disagree with the implementation.

If all base_kv are 0. ok replace them all by 1 is correct. But if some buses have a base kV and not the other this is speculative. Data should be refused in this case of "some based kV missing, some base kV present"

BDonnot (commented):
Ok

PR #195, opened 2026-09-08
Continuation powerflow with per-element load / generation steering

BDonnot:
| | added | removed |
|---|---|---|
| **C++ (src/core, src/bindings)** | +979 | -0 |
| **Python (package)** | +384 | -0 |
| **Tests** | +922 | -0 |
| **Docs + changelog** | +187 | -0 |
| **Build / other** | +98 | -0 |
| total | +2570 | -0 |

Traces the PV curve from the grid's own injections towards a target state and stops at the voltage-collapse nose, with per-element steering vectors deciding which loads and generators move along the way.

```python
from lightsim2grid import run_cpf

res = run_cpf(grid, loading_factor=3.0)
print(f"collapse at {res.lam_max:.3f} of the requested increase")

# only the loads of one area grow, at their own rates
steering = np.zeros(len(grid.get_loads()))
steering[my_area_load_ids] = 1.0
res = run_cpf(grid, loading_factor=3.0, load_steering=steering)
```

## A batch algorithm, and why a sibling rather than a fifth instantiation

`ContinuationSweep` derives from `BaseBatchSolverSynch`, which is where the batch layer's real guarantee lives: `_finish_preprocessing` solves the base case with `tell_all_changed()` — building the Jacobian sparsity and paying the one symbolic factorization — and leaves the control in `tell_none_changed()`, so every corrector after it refactorizes numerically only. **A curve of several hundred points costs one `analyze`**, and both test suites assert it. It also means the `tell_none_changed()` discipline is never applied by hand here.

It is deliberately *not* a fifth `BaseBatchSweep` instantiation. What distinguishes a continuation is not which of (Ybus, Sbus) varies — in that taxonomy it is exactly `TimeSeries`' cell — but **who decides the next row**:

| | `BaseBatchSweep` | continuation |
|---|---|---|
| row count | known up front | unknown (adaptive step, nose) |
| a diverged solve | a failed row | a retry signal: halve the step, same point |
| threading | supported | impossible, strictly sequential |
| Ybus axis | a whole policy | unused |

Making it an instantiation would have meant a policy axis the other four ignore plus loosening the fixed-row-count invariant `ScenarioSweep`'s cross-axis lock relies on.

## What the algorithm gains

Two primitives, not an algorithm: `cpf_tangent` (one triangular solve reusing the standing factorization) and `cpf_predict`. Their right-hand side is built from the `NRLedger`, so it is correct for the augmented Jacobian with no offset arithmetic, and its sign is *derived* from the residual convention in `_residual_into` rather than assumed.

That claim is tested where it costs something, on the **exotic-elements fixture** (IEEE14 + a voltage-regulating SVC + three HVDC lines with angle droop enabled on one + a phase-shifting transformer + storage), in two steps:

1. **the curve is correct there** — every traced point matches an ordinary `ac_pf`, and the test first asserts the grid really does augment J so it cannot pass vacuously;
2. **the exotic elements are actually read** — moving one has to move the margin. An HVDC converter station's voltage setpoint shifts the nose 1.648670 → 1.641051; the transfer of the non-droop HVDC line shifts it to 1.644316 (and moves the droop-controlled line with it, via the angles it responds to). The nose is reproducible to ~1e-9, so the 1e-4 threshold is far outside anything numerical. Each mutated run is itself re-checked against powerflows on its own mutated grid, so a margin that moved has moved to another *correct* answer rather than because the run broke.

Step 1 alone would pass on a continuation that ignored the HVDC entirely — the corrector would ignore it just as consistently as the predictor. Step 2 is what closes that.

## Steering

The steering vectors needed no new element→bus mapping: the direction is built with `SbusPolicy::Vary`'s own scatter helpers applied to target-minus-base deltas, so per-bus load aggregation, disconnected elements and the `sn_mva` division all come from the code `TimeSeries` already exercises.

A coefficient in `[0, 1]` per element; 0 holds it at its base value, 1 moves it fully. **`gen_steering` defaults to all ones — slack machines included**, which is where this deliberately parts from MATPOWER.

MATPOWER requires a target case to be "same as base case w.r.t. Qg and slack Pg" because a slack bus has no active-power equation: its `Pg` is an *output*, so specifying a change there asks for something the solver ignores. In lightsim2grid the slack is by design a *generator*, and that premise only half transfers:

- **single slack** — it holds. Scaling the machine's `target_p` 219 → 438 MW leaves the solution bit-identical and its output at 232.39 MW. Excluding it is a no-op, not a correction.
- **distributed slack** — it fails. The same scaling moves V by 0.12 and the output 227.54 → 308.50 MW. Excluding participants is an arbitrary dispatch choice, and on a grid whose slack spans many machines it silently freezes most of the generation.

So one rule for every generator. On a single-slack grid the curve is unchanged either way (case14, k=4: `lam_max` 1.020084), which is the point. Note the slack's *output* follows the load regardless — with every other machine scaled by k it produces roughly k× its base power; holding its setpoint would never have held its production.

`gen_steering=0` holds generation fixed and lets the slack supply the whole increase — a legitimate but **different curve with a different nose**, so the two margins are not comparable.

## Two degenerate directions

- **Zero direction** (target == base) throws: nothing was asked.
- **A direction the slack absorbs entirely** — steering only the machine at a single slack bus — is *not* refused; it is a legal request that simply has nowhere to go. But it stops at the base case with `success=False` and an explanatory message, because before that it marched λ to 20.8 over a thousand points on which |V| changed by 3e-15 and reported success.

## MATPOWER alignment

Option names and defaults follow `cpf.*` (`step` 0.05, `step_min` 1e-4, `step_max` 0.2, `adapt_step` off using the error-based scheme with `adapt_step_damping` 0.7 / `adapt_step_tol` 1e-3, `nose_tol` 1e-5, base→target formulation with λ=0/1).

The one thing that cannot follow MATPOWER is the parameterisation: this is the natural one (`cpf.parameterization = 1`), so **the curve stops at the nose rather than rounding it**. Pseudo arc length — MATPOWER's default — needs a λ column and a parameterisation row inside the Jacobian, i.e. an `NRSystem` extension, and is left for a follow-up. Docs and docstring say so, and that the default will flip when it lands.

## Verification

Correctness oracles are independent of the continuation itself:

- **every traced point re-solved with a plain `ac_pf`** at the same injections, on a plain grid and on the exotic one — the tolerance is keyed on `tangent_lam`, which *is* the conditioning of the point, rather than excluding points by index;
- **nose vs. brute-force bisection** on the loading factor (case14): 1.020084 both ways, 2.3e-9 relative;
- **`nb_analyze == 1`** for a 278-point curve, extensions and all.

Three failure modes have their own tests because each would otherwise produce a plausible-looking curve rather than an error: an inverted load direction (loads are stamped into Sbus with a minus, so loading must *lower* the voltages), a direction that moves nothing, and an exotic element the continuation does not read.

32 Python tests and 14 Catch2 tests; the full C++ suite is 266/266 at both C++14 and C++26, no regressions. `test_jacobian`, `test_dist_slack_jacobian`, `test_multi_slack` and `test_pf_input_robustness` are clean. `test_ACPF` could not run in the dev container (no grid2op installed, so `LightSimBackend` is unimportable) — unrelated to this change, and covered by CI.

## Files

- `src/core/batch_algorithm/ContinuationSweep.{hpp,cpp}` — the algorithm
- `src/core/powerflow_algorithm/{BaseAlgo,NRAlgo,NRSystem}.{hpp,tpp}` — the three NR primitives (purely additive)
- `src/core/AlgorithmSelector.hpp` — forwarding, guarded like `get_J`
- `src/bindings/python/binding_batch.cpp` — `ContinuationSweepCPP`
- `lightsim2grid/continuationPowerflow.py` — the wrapper, `plot_pv_curve`
- `docs/continuation_powerflow.rst`, `CHANGELOG.rst`, tests
- `utils/pr_diff_stats.py` + `CLAUDE.md` — the table at the top of this body, and the note asking future PRs to lead with it

https://claude.ai/code/session_01P1QabyFxUXhXSyBgmUAteW

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -47,47 +47,22 @@ have to reason about and how much is tests, docs and changelog:
 ```
 | | added | removed |
 |---|---|---|
-| **C++ (src/core, src/bindings)** | +968 | -0 |
+| **C++ (src/core, src/bindings)** | +970 | -0 |
 | **Python (package)** | +384 | -0 |
-| **Tests** | +764 | -0 |
-| **Docs + changelog** | +160 | -0 |
-| total | +2277 | -0 |
+| **Tests** | +922 | -0 |
+| **Docs + changelog** | +212 | -0 |
+| total | +2489 | -0 |
 ```
 
-Buckets, in this order (drop a row that is empty):
-
-- **C++** — `src/core/**` and `src/bindings/**` (`.cpp` / `.hpp` / `.tpp`), tests excluded.
-- **Python** — `lightsim2grid/**.py`, tests excluded.
-- **Tests** — `src/tests/**` and `lightsim2grid/tests/**`, whatever the language.
-- **Docs + changelog** — `docs/**` and any `.rst` / `.md`, `CHANGELOG.rst` included.
-- **Build / other** — everything left (CMake, CI, fixtures), so the total is the real total.
-
-Generate it against the PR's BASE branch rather than counting by hand:
+`utils/pr_diff_stats.py` produces it — it holds the bucket definitions, so they stay in one
+place rather than being restated here:
 
 ```
-python3 - "$(git merge-base origin/<base> HEAD)" <<'EOF'
-import subprocess, sys, fnmatch
-out = subprocess.run(["git", "diff", "--numstat", sys.argv[1] + "...HEAD"],
-                     capture_output=True, text=True).stdout
-def bucket(p):
-    if fnmatch.fnmatch(p, "src/tests/*") or fnmatch.fnmatch(p, "lightsim2grid/tests/*"): return "tests"
-    if p.startswith("docs/") or p.endswith((".rst", ".md")): return "docs"
-    if p.startswith(("src/core/", "src/bindings/")) and p.endswith((".cpp", ".hpp", ".tpp", ".h")): return "cpp"
-    if p.startswith("lightsim2grid/") and p.endswith(".py"): return "python"
-    return "other"
-tot = {}
-for line in out.strip().splitlines():
-    a, d, path = line.split("\t")
-    if a == "-": continue          # binary file
-    e = tot.setdefault(bucket(path), [0, 0]); e[0] += int(a); e[1] += int(d)
-rows = [("cpp", "C++ (src/core, src/bindings)"), ("python", "Python (package)"),
-        ("tests", "Tests"), ("docs", "Docs + changelog"), ("other", "Build / other")]
-print("| | added | removed |\n|---|---|---|")
-for k, label in rows:
-    if k in tot: print(f"| **{label}** | +{tot[k][0]} | -{tot[k][1]} |")
-print(f"| total | +{sum(v[0] for v in tot.values())} | -{sum(v[1] for v in tot.values())} |")
-EOF
+python3 utils/pr_diff_stats.py                # against origin/dev_1.0.1 (the default)
+python3 utils/pr_diff_stats.py origin/main    # against another base branch
 ```
+
+Run it against the PR's own base branch, and paste the table as the first thing in the body.
 
 ## Build
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -37,6 +37,58 @@ Rules:
   followed by `git push --force-with-lease`. This is the case that actually bites: the
   mistake is usually noticed only after CI has run.
 
+## Pull requests: lead with the diff breakdown
+
+A PR here routinely runs to one or two thousand lines, which is daunting to open and tells
+a reviewer nothing about where the work actually is. **Start every PR body with a table
+splitting the diff by kind**, so the reader can see at a glance how much of it is code they
+have to reason about and how much is tests, docs and changelog:
+
+```
+| | added | removed |
+|---|---|---|
+| **C++ (src/core, src/bindings)** | +968 | -0 |
+| **Python (package)** | +384 | -0 |
+| **Tests** | +764 | -0 |
+| **Docs + changelog** | +160 | -0 |
+| total | +2277 | -0 |
+```
+
+Buckets, in this order (drop a row that is empty):
+
+- **C++** — `src/core/**` and `src/bindings/**` (`.cpp` / `.hpp` / `.tpp`), tests excluded.
+- **Python** — `lightsim2grid/**.py`, tests excluded.
+- **Tests** — `src/tests/**` and `lightsim2grid/tests/**`, whatever the language.
+- **Docs + changelog** — `docs/**` and any `.rst` / `.md`, `CHANGELOG.rst` included.
+- **Build / other** — everything left (CMake, CI, fixtures), so the total is the real total.
+
+Generate it against the PR's BASE branch rather than counting by hand:
+
+```
+python3 - "$(git merge-base origin/<base> HEAD)" <<'EOF'
+import subprocess, sys, fnmatch
+out = subprocess.run(["git", "diff", "--numstat", sys.argv[1] + "...HEAD"],
+                     capture_output=True, text=True).stdout
+def bucket(p):
+    if fnmatch.fnmatch(p, "src/tests/*") or fnmatch.fnmatch(p, "lightsim2grid/tests/*"): return "tests"
+    if p.startswith("docs/") or p.endswith((".rst", ".md")): return "docs"
+    if p.startswith(("src/core/", "src/bindings/")) and p.endswith((".cpp", ".hpp", ".tpp", ".h")): return "cpp"
+    if p.startswith("lightsim2grid/") and p.endswith(".py"): return "python"
+    return "other"
+tot = {}
+for line in out.strip().splitlines():
+    a, d, path = line.split("\t")
+    if a == "-": continue          # binary file
+    e = tot.setdefault(bucket(path), [0, 0]); e[0] += int(a); e[1] += int(d)
+rows = [("cpp", "C++ (src/core, src/bindings)"), ("python", "Python (package)"),
+        ("tests", "Tests"), ("docs", "Docs + changelog"), ("other", "Build / other")]
+print("| | added | removed |\n|---|---|---|")
+for k, label in rows:
+    if k in tot: print(f"| **{label}** | +{tot[k][0]} | -{tot[k][1]} |")
+print(f"| total | +{sum(v[0] for v in tot.values())} | -{sum(v[1] for v in tot.values())} |")
+EOF
+```
+
 ## Build
 
 The vendored dependencies are git submodules and start empty — a build fails with
--- a/src/core/batch_algorithm/ContinuationSweep.cpp
+++ b/src/core/batch_algorithm/ContinuationSweep.cpp
@@ -1,0 +1,339 @@
+// Copyright (c) 2020-2026, RTE (https://www.rte-france.com)
+// See AUTHORS.txt
+// This Source Code Form is subject to the terms of the Mozilla Public License, version 2.0.
+// If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+// you can obtain one at http://mozilla.org/MPL/2.0/.
+// SPDX-License-Identifier: MPL-2.0
+// This file is part of LightSim2grid, LightSim2grid implements a c++ backend targeting the Grid2Op platform.
+
+#include "ContinuationSweep.hpp"
+
+#include <algorithm>
+#include <cmath>
+#include <sstream>
+
+namespace ls2g {
+
+real_type ContinuationSweep::_checked_positive(real_type v, const char * name)
+{
+    if(!std::isfinite(v) || (v <= 0.)){
+        std::ostringstream exc_;
+        exc_ << "ContinuationSweep: " << name << " must be a finite, strictly positive "
+             << "number, got " << v << ".";
+        throw std::runtime_error(exc_.str());
+    }
+    return v;
+}
+
+void ContinuationSweep::set_max_steps(int v)
+{
+    if(v < 1){
+        std::ostringstream exc_;
+        exc_ << "ContinuationSweep: max_steps must be >= 1, got " << v << ".";
+        throw std::runtime_error(exc_.str());
+    }
+    _max_steps = v;
+}
+
+SbusPolicy::Vary::RealMat ContinuationSweep::_delta_row(const RealVect & target,
+                                                        const Eigen::Ref<const RealVect> & base,
+                                                        const char * axis_name)
+{
+    SbusPolicy::Vary::RealMat res = SbusPolicy::Vary::RealMat::Zero(1, base.size());
+    if(target.size() == 0) return res;  // axis never set: no contribution
+    if(target.size() != base.size()){
+        std::ostringstream exc_;
+        exc_ << "ContinuationSweep::set_target_" << axis_name << ": expected one value per "
+             << "element (" << base.size() << "), got " << target.size() << ".";
+        throw std::runtime_error(exc_.str());
+    }
+    res.row(0) = (target - base).transpose();
+    return res;
+}
+
+void ContinuationSweep::_build_direction(int nb_buses_solver,
+                                         const SolverBusIdVect & id_me_to_solver)
+{
+    const auto & generators = _grid_model.get_generators_as_data();
+    const auto & s_generators = _grid_model.get_static_generators_as_data();
+    const auto & loads = _grid_model.get_loads_as_data();
+    const char * algo_name = "ContinuationSweep";
+
+    const SbusPolicy::Vary::RealMat d_gen_p =
+        _delta_row(_target_gen_p, _grid_model.get_gen_target_p(), "gen_p");
+    const SbusPolicy::Vary::RealMat d_sgen_p =
+        _delta_row(_target_sgen_p, _grid_model.get_sgen_target_p(), "sgen_p");
+    const SbusPolicy::Vary::RealMat d_load_p =
+        _delta_row(_target_load_p, _grid_model.get_load_target_p(), "load_p");
+    const SbusPolicy::Vary::RealMat d_load_q =
+        _delta_row(_target_load_q, loads.get_target_q(), "load_q");
+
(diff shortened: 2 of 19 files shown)


CLAUDE.md
```diff
@@ -37,6 +37,58 @@ Rules:
...
+    if a == "-": continue          # binary file
+    e = tot.setdefault(bucket(path), [0, 0]); e[0] += int(a); e[1] += int(d)
+rows = [("cpp", "C++ (src/core, src/bindings)"), ("python", "Python (package)"),
+        ("tests", "Tests"), ("docs", "Docs + changelog"), ("other", "Build / other")]
+print("| | added | removed |\n|---|---|---|")
+for k, label in rows:
+    if k in tot: print(f"| **{label}** | +{tot[k][0]} | -{tot[k][1]} |")
+print(f"| total | +{sum(v[0] for v in tot.values())} | -{sum(v[1] for v in tot.values())} |")
+EOF
+```
```
BDonnot: In claude.md, are you sure ? Create this file somewhere and put a link to it there don't you think ?

src/core/batch_algorithm/ContinuationSweep.hpp
```diff
@@ -0,0 +1,282 @@
...
+--- Defaults ----------------------------------------------------------------------
+
+The option names and default values follow MATPOWER's `cpf.*` so that a user who
+knows `runcpf` is not faced with new concepts: step 0.05, step_min 1e-4, step_max
+0.2, adapt_step off, adapt_step_damping 0.7, adapt_step_tol 1e-3, nose_tol 1e-5,
+and "stop at the nose" unless a target lambda is given. The one place this cannot
+follow MATPOWER is the parameterisation, whose default there is pseudo arc length
+(see above).
+**/
+class LS2G_API ContinuationSweep : public BaseBatchSolverSynch
```
BDonnot: ContinuationSweep should be final I guess

src/core/batch_algorithm/ContinuationSweep.hpp
```diff
@@ -0,0 +1,282 @@
...
+        // One entry per grid element of the relevant type, in MW / MVAr, i.e. the
+        // same units and ordering as LSGrid::get_gen_target_p() and friends. An axis
+        // never set (or set empty) means "same as the base case", so it contributes
+        // nothing to the direction. Setting none of them at all is an error: the
+        // direction would be zero and the continuation would walk lambda along a
+        // curve that does not move.
+        void set_target_gen_p(const RealVect & values)  {_target_gen_p = values;}
+        void set_target_sgen_p(const RealVect & values) {_target_sgen_p = values;}
+        void set_target_load_p(const RealVect & values) {_target_load_p = values;}
+        void set_target_load_q(const RealVect & values) {_target_load_q = values;}
```
BDonnot: why not cont Eigen::Ref<const RealVect> & here ? (following the other part of the code)

src/core/batch_algorithm/ContinuationSweep.hpp
```diff
@@ -0,0 +1,282 @@
...
+        real_type get_step_max() const {return _step_max;}
+        void set_step_max(real_type v) {_step_max = _checked_positive(v, "step_max");}
+        bool get_adapt_step() const {return _adapt_step;}
+        void set_adapt_step(bool v) {_adapt_step = v;}
+        real_type get_adapt_step_damping() const {return _adapt_step_damping;}
+        void set_adapt_step_damping(real_type v) {_adapt_step_damping = _checked_positive(v, "adapt_step_damping");}
+        real_type get_adapt_step_tol() const {return _adapt_step_tol;}
+        void set_adapt_step_tol(real_type v) {_adapt_step_tol = _checked_positive(v, "adapt_step_tol");}
+        real_type get_nose_tol() const {return _nose_tol;}
+        void set_nose_tol(real_type v) {_nose_tol = _checked_positive(v, "nose_tol");}
```
BDonnot: v is a weird name for a real number. Prefer I don't know x for example

src/core/batch_algorithm/ContinuationSweep.hpp
```diff
@@ -0,0 +1,282 @@
...
+        // Number of traced points, base case included. get_voltages() has exactly
+        // this many rows (it is truncated at the end of compute()).
+        Eigen::Index nb_points() const {return _nb_points;}
+
+        // lambda at each traced point; lam(0) == 0 (the base case).
+        const RealVect & get_lam() const {return _lam;}
+        // the tangent's lambda component at each traced point, in (0, 1]. It tends to
+        // zero at the nose -- that is the collapse indicator, see the class comment.
+        // The last point has no tangent computed after it and reads 0.
+        const RealVect & get_tangent_lam() const {return _tangent_lam;}
```
BDonnot: why not Eigen:Ref<const RealVect> here ?

src/core/batch_algorithm/ContinuationSweep.cpp
```diff
@@ -0,0 +1,362 @@
...
+        // bus' P equation exists and takes the direction, but the slack-absorbed unknown
+        // cancels it one for one, so the machine's actual output is unchanged and no bus
+        // moves. So this is NOT refused. What is refused is pretending it went somewhere:
+        // continuing would march lambda to its target over hundreds of identical points
+        // and report a loading margin that is really an artefact of max_steps. Stop here
+        // instead, with the base case as the only point and a message saying why.
+        //
+        // Checked on the first predictor only: which unknowns a direction reaches is a
+        // property of it and of the topology, and neither changes along the curve.
+        if (it == 0 && V_pred == V_last) {
```
BDonnot: V is not int or bool or char, it's floating point (complex) and "==" should be avoided for comparing complex numbers. Use rather (V_pref - V_last).abs().max() < _tol_float (not sure about _tol_float)

src/core/batch_algorithm/ContinuationSweep.cpp
```diff
@@ -0,0 +1,362 @@
...
+                   "changes a real input, but not one the powerflow reads, since that bus' "
+                   "generation is whatever balances the grid rather than what was asked of "
+                   "it. There is no loading margin along such a direction, so nothing is "
+                   "traced past the base case; steer something whose power the powerflow "
+                   "actually has to route.";
+            break;
+        }
+
+        // ---- corrector ---------------------------------------------------------
+        CplxVect V = V_pred;
```
BDonnot: why all these copies of V, V_pred, V_last ? Are they all mandatory ?

BDonnot (commented):
fix merge conflicts (should be on changelog and cmakefilest so no issue) and adress the issues

CLAUDE.md
BDonnot: You're right — moved it to `utils/pr_diff_stats.py`, next to the other dev scripts, and CLAUDE.md now just links to it with the two usage lines. The bucket definitions live in the script rather than being restated in prose, so the two can't drift apart.

```
python3 utils/pr_diff_stats.py                # against origin/dev_1.0.1 (the default)
python3 utils/pr_diff_stats.py origin/main    # another base branch
```

---

src/core/batch_algorithm/ContinuationSweep.hpp
BDonnot: Agreed, done. Nothing derives from it and nothing should — the loop is the class.

---

src/core/batch_algorithm/ContinuationSweep.hpp
BDonnot: No good reason — done, the four `set_target_*` now take `const Eigen::Ref<const RealVect> &`.

---

src/core/batch_algorithm/ContinuationSweep.hpp
BDonnot: Fair — `v` reads as a voltage everywhere else in this file. Renamed to `x` for the numeric setters and `on` for the two booleans (`set_adapt_step`, `set_exact_tangent`), plus the `_checked_positive` parameter.

---

src/core/batch_algorithm/ContinuationSweep.hpp
BDonnot: Done — `get_lam()` and `get_tangent_lam()` now return `Eigen::Ref<const RealVect>`, and `get_direction_solver()` returns `Eigen::Ref<const CplxVect>` for the same reason.

---

src/core/batch_algorithm/ContinuationSweep.cpp
BDonnot: Right. It happened to be exact — both sides come out of the same `_reconstruct_V_into` on the same `(Va, Vm)`, so the bits match — but relying on that is fragile and reads wrong regardless.

Now:

```cpp
if (it == 0 &&
    (V_pred - V_last).cwiseAbs().maxCoeff() <= BaseConstants::_tol_equal_float) {
```

`_tol_equal_float` is the one this file already uses (for the `sn_mva != 1` test). It's also the more honest statement of what's being asked: not "bit-identical" but "moved nothing" — and there's no risk of it firing on a real direction, since this runs on the first predictor with a full `step`, where a genuine direction moves voltages by step-sized amounts.

---

src/core/batch_algorithm/ContinuationSweep.cpp
BDonnot: Two of them were not, and they were worse than copies — `V_corr` (then called `V`) and `Sbus` were *constructed* inside the loop, so every one of a curve's several hundred iterations allocated and freed two `nb_bus` vectors. Both are now declared once above the loop and assigned into, which reuses the storage.

The four that remain are each read after the next is written, so they can't be collapsed:

- `V_pred` — the predictor's output. Still needed *after* the corrector has overwritten `V_corr`, because the step adaptation is `‖corrected − predicted‖∞`.
- `V_corr` — `compute_one_powerflow` overwrites its argument with the solution, so it can't be `V_pred` itself.
- `V_last` — the last *converged* point. Must survive a corrector failure, since the retry re-seeds from it (`_restore_at`) and the next tangent is taken there.
- `z` — the tangent; already hoisted.

Comment on the declarations records that.

---

src/core/batch_algorithm/ContinuationSweep.hpp
```diff
@@ -0,0 +1,282 @@
...
+        // exactly as in an ordinary solve). Everything the four axes do NOT cover is
+        // identical in the base and the target state and therefore cancels in the
+        // difference -- which is why this needs no equivalent of
+        // SbusPolicy::Vary::constant_sbus_pu.
+        void _build_direction(int nb_buses_solver, const SolverBusIdVect & id_me_to_solver);
+
+        // one target axis as a 1-row matrix of DELTAS (target - base), or a 1 x n
+        // zero matrix when the axis was never set. `base` is the grid's own target
+        // vector for that element type.
+        static SbusPolicy::Vary::RealMat _delta_row(const RealVect & target,
```
BDonnot: Eigen::Ref<const> here too maybe ?

src/core/batch_algorithm/ContinuationSweep.hpp
```diff
@@ -0,0 +1,282 @@
...
+
+        static real_type _checked_positive(real_type x, const char * name);
+
+        // Re-establish the algorithm's state at the last converged point after a
+        // corrector diverged: the failed solve left (Va, Vm) and the standing
+        // factorization at some diverged iterate, and both feed the next tangent.
+        // Re-solving from V_last converges immediately -- and precisely BECAUSE it
+        // converges immediately it never enters the NR loop and so never
+        // refactorizes, which is why the refactorization is asked for explicitly.
+        bool _restore_at(const CplxVect & V_last, real_type lam_last,
```
BDonnot: Eigen::Ref<const XXX> here too maybe

src/core/batch_algorithm/ContinuationSweep.hpp
```diff
@@ -0,0 +1,282 @@
...
+                         int max_iter, real_type tol);
+
+        // MATPOWER's predictor-error step control (cpf.adapt_step):
+        //   err        = ||[theta(pv,pq); |V|(pq)]_corrected - [..]_predicted||_inf
+        //   step_scale = min(2, 1 + damping . (tol/err - 1))
+        // clipped to [step_min, step_max]. MATPOWER's error also carries a lambda
+        // term; here it is structurally zero (see the definition).
+        real_type _adapted_step(real_type step,
+                                const CplxVect & V_corr,
+                                const CplxVect & V_pred) const;
```
BDonnot: Eigen::Ref<const XXX> here too maybe

src/core/batch_algorithm/ContinuationSweep.hpp
BDonnot: Eigen::Ref<const XXX> here too maybe

BDonnot (commented):
some more Eigen::Ref to add maybe

src/core/batch_algorithm/ContinuationSweep.hpp
BDonnot: Done — `_delta_row`'s `target` is now `const Eigen::Ref<const RealVect> &`, matching `base` next to it (which already was, which made the asymmetry worse).

---

src/core/batch_algorithm/ContinuationSweep.hpp
BDonnot: Done — `_restore_at` takes `const Eigen::Ref<const CplxVect> & V_last`.

---

src/core/batch_algorithm/ContinuationSweep.hpp
BDonnot: Done — `_adapted_step` takes both `V_corr` and `V_pred` as `const Eigen::Ref<const CplxVect> &`.

That's all four of the remaining bare references; `grep` for `const RealVect &` / `const CplxVect &` in the two files now comes back empty. Every current call site passes a plain vector, so nothing copies.

---

src/core/batch_algorithm/ContinuationSweep.hpp
BDonnot: Same line as above — `_restore_at`'s `V_last` is now `const Eigen::Ref<const CplxVect> &`.

---

BDonnot (commented):
ok

PR #199, opened 2026-09-12
linear solvers: solve J^T x = b out of the factorization of J

BDonnot:
| | added | removed |
|---|---|---|
| **C++ (src/core, src/bindings)** | +114 | -1 |
| **Tests** | +220 | -0 |
| total | +338 | -1 |

First of four PRs towards `BatchCPUPowerFlow`, a batched differentiable AC powerflow for PyTorch over `ScenarioSweep`. This one adds only the linear-solver capability the adjoint needs; nothing here depends on the rest, and nothing else in the tree calls it yet.

## Why

Reverse-mode differentiation of a powerflow needs the adjoint system `J^T λ = x̄` at the converged point. Forming `J^T` and factorizing it costs a second symbolic analysis and a second numeric factorization per solve — which is what a GPU library has to do, its direct solver only going one way.

KLU does not: `klu_tsolve` walks the very `L` and `U` that `klu_factor` / `klu_refactor` already produced, backwards. Eigen's `SparseLU` has the same through its transpose view. So the adjoint of an already-factorized system costs one triangular solve, and the whole "build the transposed pattern, map J's values into it, analyze, factorize" machinery is unnecessary on CPU.

## What

- `solve_transpose(b)` on the four linear solvers, with `CAN_SOLVE_TRANSPOSE` saying which of them have it.
  - KLU → `klu_tsolve`, same `Symbolic` / `Numeric` as `klu_solve`.
  - SparseLU → `solver_.transpose().solve(b)`. Non-`const`, unlike `solve()`: Eigen's `SparseLU::transpose()` is a non-const member.
- NICSLU and CKTSO report the capability as **absent**. Both ship outside this repository and are not built in CI, so whether they can walk their factors transposed is unverified here — rather than claim it and be wrong, they return the new `ErrorType::NotImplemented` instead of silently answering `J x = b`. A plausible-looking solution to the wrong system is the one outcome worth ruling out. The comments say this is a not-implemented-here, not a library limitation, so whoever has the vendor docs can fill it in. A caller branches on the constant at compile time or on the error at runtime.
- `ErrorType::NotImplemented` is appended at the **end** of the enum, so no existing value renumbers; bound to Python alongside the others.
- `LinearSolverPolicy` counts and times transposed solves apart from plain ones (`nb_solve_transpose`, `timer_solve_transpose_`), so an adjoint's cost can be attributed rather than guessed. That measurement is what a later decision about keeping per-row factorizations will rest on.

Out-of-tree linear solvers plugged into `LinearSolverPolicy` are unaffected: both the new method and the new static member are instantiated only if used.

## Tests

New `src/tests/test_solve_transpose.cpp`, on matrices asymmetric in **both** pattern and values:

- `solve_transpose` against a dense LU of the explicitly transposed matrix, and the residual of the transposed system;
- the untransposed `solve` gives a *different* vector — without that assertion the test would pass on an implementation that dropped the transposition;
- the sequence the batch adjoint will use: one `analyze`, then new values into the same pattern, `refactorize`, transposed solve;
- KLU and SparseLU agreeing with each other and with the dense reference, before and after a refactorize;
- `LinearSolverPolicy` counting a transposed solve separately from a plain one.

## Verification

- Full C++ suite **277/277**, at the default C++26 and at a pinned C++14 build (`-std=c++14` confirmed in the generated flags). KLU built from the SuiteSparse submodule (v2.3.6), so the KLU paths really run.
- **Mutation check**: swapping `klu_tsolve` → `klu_solve` makes both KLU tests fail, so the test bites rather than passing by symmetry. Reverted and re-verified.
- `binding_enums.cpp` syntax-checked at C++14 against pybind11 and the vendored SuiteSparse headers.

No Python test touched: nothing in the Python API changes beyond the new enum value.

---

### DIFF
--- a/src/core/linear_solvers/KLUSolver.cpp
+++ b/src/core/linear_solvers/KLUSolver.cpp
@@ -12,9 +12,6 @@
 
 namespace ls2g {
 
-const bool KLULinearSolver::CAN_SOLVE_MAT = false;
-const bool KLULinearSolver::CAN_SOLVE_TRANSPOSE = true;
-
 ErrorType KLULinearSolver::reset(){
     // release both handles (their deleters use common_) before resetting common_
     numeric_.reset();
--- a/src/core/linear_solvers/KLUSolver.cpp
+++ b/src/core/linear_solvers/KLUSolver.cpp
@@ -13,6 +13,7 @@
 namespace ls2g {
 
 const bool KLULinearSolver::CAN_SOLVE_MAT = false;
+const bool KLULinearSolver::CAN_SOLVE_TRANSPOSE = true;
 
 ErrorType KLULinearSolver::reset(){
     // release both handles (their deleters use common_) before resetting common_
@@ -70,6 +71,20 @@ ErrorType KLULinearSolver::solve(Eigen::Ref<RealVect> b){
     int ok = klu_solve(symbolic_.get(), numeric_.get(), n, 1, &b(0), &common_);
     if(ok != 1){
         // std::cout << "\t KLU: klu_solve error" << std::endl;
+        return ErrorType::SolverSolve;
+    }
+    return ErrorType::NoError;
+}
+
+ErrorType KLULinearSolver::solve_transpose(Eigen::Ref<RealVect> b){
+    // klu_tsolve walks the very same L / U factors as klu_solve, backwards: J^T is
+    // never formed, and neither the symbolic analysis nor the numeric factorization
+    // is repeated. That is what makes an adjoint (a vector-Jacobian product, as
+    // reverse-mode differentiation of a powerflow needs) cost one triangular solve.
+    const int n = static_cast<int>(b.size());
+    int ok = klu_tsolve(symbolic_.get(), numeric_.get(), n, 1, &b(0), &common_);
+    if(ok != 1){
+        // std::cout << "\t KLU: klu_tsolve error" << std::endl;
         return ErrorType::SolverSolve;
     }
     return ErrorType::NoError;
--- a/src/core/linear_solvers/KLUSolver.hpp
+++ b/src/core/linear_solvers/KLUSolver.hpp
@@ -58,10 +58,10 @@ class LS2G_API KLULinearSolver final
         ErrorType solve_transpose(Eigen::Ref<RealVect> b);  // J^T x = b, out of the factorization of J
 
         // can this linear solver solve problem where RHS is a matrix
-        static const bool CAN_SOLVE_MAT;
+        static constexpr bool CAN_SOLVE_MAT = false;
 
         // can this linear solver solve J^T x = b out of the factorization of J
-        static const bool CAN_SOLVE_TRANSPOSE;
+        static constexpr bool CAN_SOLVE_TRANSPOSE = true;  // klu_tsolve
 
     private:
         // KLU frees its symbolic / numeric handles through a pair of functions that
--- a/src/core/linear_solvers/KLUSolver.hpp
+++ b/src/core/linear_solvers/KLUSolver.hpp
@@ -55,9 +55,13 @@ class LS2G_API KLULinearSolver final
         ErrorType factorize(const EigenRefConstRealSpMat & J); // numeric factorization (requires values)
         ErrorType refactorize(const EigenRefConstRealSpMat & J);  // re-numeric factorization, reuses symbolic
         ErrorType solve(Eigen::Ref<RealVect> b);
+        ErrorType solve_transpose(Eigen::Ref<RealVect> b);  // J^T x = b, out of the factorization of J
 
         // can this linear solver solve problem where RHS is a matrix
         static const bool CAN_SOLVE_MAT;
+
+        // can this linear solver solve J^T x = b out of the factorization of J
+        static const bool CAN_SOLVE_TRANSPOSE;
 
     private:
         // KLU frees its symbolic / numeric handles through a pair of functions that
@@ -132,9 +136,13 @@ class LS2G_API KLULinearSolver final
         ErrorType factorize(const EigenRefConstRealSpMat & /*J*/) { return ErrorType::NoError; }
         ErrorType refactorize(const EigenRefConstRealSpMat & /*J*/) { return ErrorType::NoError; }
         ErrorType solve(Eigen::Ref<RealVect> /*b*/) { return ErrorType::NoError; }
+        ErrorType solve_transpose(Eigen::Ref<RealVect> /*b*/) { return ErrorType::NoError; }
 
         // can this linear solver solve problem where RHS is a matrix
         static constexpr bool CAN_SOLVE_MAT = false;
+
+        // can this linear solver solve J^T x = b out of the factorization of J
+        static constexpr bool CAN_SOLVE_TRANSPOSE = true;
 
     private:
         // no copy allowed (matches the real KLULinearSolver)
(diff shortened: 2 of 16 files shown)


src/core/linear_solvers/KLUSolver.hpp
```diff
@@ -55,10 +55,14 @@ class LS2G_API KLULinearSolver final
         ErrorType factorize(const EigenRefConstRealSpMat & J); // numeric factorization (requires values)
         ErrorType refactorize(const EigenRefConstRealSpMat & J);  // re-numeric factorization, reuses symbolic
         ErrorType solve(Eigen::Ref<RealVect> b);
+        ErrorType solve_transpose(Eigen::Ref<RealVect> b);  // J^T x = b, out of the factorization of J
 
         // can this linear solver solve problem where RHS is a matrix
         static const bool CAN_SOLVE_MAT;
 
+        // can this linear solver solve J^T x = b out of the factorization of J
+        static const bool CAN_SOLVE_TRANSPOSE;
```
BDonnot: Const or constexpr ?

src/core/linear_solvers/KLUSolver.cpp
```diff
@@ -13,6 +13,7 @@
 namespace ls2g {
 
 const bool KLULinearSolver::CAN_SOLVE_MAT = false;
+const bool KLULinearSolver::CAN_SOLVE_TRANSPOSE = true;
```
BDonnot: Constexpr would be better

src/core/linear_solvers/LinearSolverPolicy.hpp
```diff
@@ -39,6 +39,10 @@ class LinearSolverPolicy
         // can this linear solver solve problem where RHS is a matrix
         static const bool CAN_SOLVE_MAT;
 
+        // can this linear solver solve J^T x = b using the factorization of J itself,
+        // without ever forming J^T (see solve_transpose)
+        static const bool CAN_SOLVE_TRANSPOSE;
```
BDonnot: Constexpr better

src/core/linear_solvers/LinearSolverPolicy.hpp
```diff
@@ -134,6 +153,9 @@ class LinearSolverPolicy
 template<class LinearSolver>
 const bool LinearSolverPolicy<LinearSolver>::CAN_SOLVE_MAT = LinearSolver::CAN_SOLVE_MAT;
 
+template<class LinearSolver>
+const bool LinearSolverPolicy<LinearSolver>::CAN_SOLVE_TRANSPOSE = LinearSolver::CAN_SOLVE_TRANSPOSE;
+
```
BDonnot: Constexpr here too

src/core/linear_solvers/NICSLUSolver.cpp
```diff
@@ -13,6 +13,13 @@
 namespace ls2g {
 
 const bool NICSLULinearSolver::CAN_SOLVE_MAT = false;
+// NICSLU's own solve entry point is wrapped here as `Solve(b, x)` only. Whether the
+// library can walk its factors transposed was not established against its headers (it
+// ships separately from this repository and is not built in CI), so lightsim2grid does
+// not claim the capability: a caller needing J^T x = b with NICSLU has to form the
+// transpose itself and factorize it. Implementing this is a contained follow-up for
+// anyone with the NICSLU documentation at hand.
+const bool NICSLULinearSolver::CAN_SOLVE_TRANSPOSE = false;
```
BDonnot: Constexpr

src/core/linear_solvers/NICSLUSolver.hpp
```diff
@@ -74,10 +74,15 @@ class LS2G_API NICSLULinearSolver final
...
         ErrorType solve(Eigen::Ref<RealVect> b);
+        // Not implemented: see the .cpp. Callers must honour CAN_SOLVE_TRANSPOSE.
+        ErrorType solve_transpose(Eigen::Ref<RealVect> b);
 
         // can this linear solver solve problem where RHS is a matrix
         static const bool CAN_SOLVE_MAT;
 
+        // can this linear solver solve J^T x = b out of the factorization of J
+        static const bool CAN_SOLVE_TRANSPOSE;
+
```
BDonnot: Constexpr

src/core/linear_solvers/SparseLUSolver.cpp
```diff
@@ -13,6 +13,7 @@
 namespace ls2g {
 
 const bool SparseLULinearSolver::CAN_SOLVE_MAT = true;
+const bool SparseLULinearSolver::CAN_SOLVE_TRANSPOSE = true;
```
BDonnot: Constexpr

BDonnot (commented):
Some constexpr instead of const would be better

src/core/linear_solvers/KLUSolver.hpp
BDonnot: Agreed — done in 12f9046, for both flags and on all four solvers plus the policy. The `.cpp` definitions are gone; the real classes now spell them exactly like their `_READ_THE_DOCS` stand-ins already did.

One subtlety worth recording, since it bit the test: before C++17 a `constexpr` static data member still needs an out-of-line definition if anything *odr-uses* it, and this project builds at C++14. Nothing outside these headers reads either flag, except the two checks I had written as `REQUIRE(...)` — and Catch2's expression decomposition binds its operands by reference, which is an odr-use. Those are `static_assert` now: a compile-time answer to a compile-time question, and no definition needed. Built and linked the whole test binary at C++14 as well as at the default standard to confirm.

`LinearSolverPolicy` mirrors the wrapped solver's flags with an in-class initializer, which keeps the property the out-of-line template definitions had: it is instantiated only if the member is used, so an out-of-tree linear solver declaring neither flag still works with the policy.

---

src/core/linear_solvers/KLUSolver.cpp
BDonnot: Done in 12f9046 — the definition moved into the header as `constexpr`. See the note on the KLUSolver.hpp thread for the C++14 odr-use detail.

---

src/core/linear_solvers/LinearSolverPolicy.hpp
BDonnot: Done in 12f9046 — both flags are now `static constexpr`, initialised in-class from the wrapped solver's.

---

src/core/linear_solvers/LinearSolverPolicy.hpp
BDonnot: Done in 12f9046 — these two out-of-line template definitions are gone entirely; the in-class `constexpr` initializer replaces them and keeps the same lazy-instantiation property.

---

src/core/linear_solvers/NICSLUSolver.cpp
BDonnot: Done in 12f9046 — the constant and the comment explaining why it is `false` both moved to the header.

---

src/core/linear_solvers/NICSLUSolver.hpp
BDonnot: Done in 12f9046 — `static constexpr bool CAN_SOLVE_MAT = false;` / `CAN_SOLVE_TRANSPOSE = false;` here, with the rationale alongside them.

---

src/core/linear_solvers/SparseLUSolver.cpp
BDonnot: Done in 12f9046 — both moved into SparseLUSolver.hpp as `constexpr` (`true` for both here).

---

PR #200, opened 2026-09-12
batch: keep each row's Jacobian, and solve the adjoint system of the whole sweep

BDonnot:
| | added | removed |
|---|---|---|
| **C++ (src/core, src/bindings)** | +740 | -6 |
| **Tests** | +503 | -0 |
| **Docs + changelog** | +5 | -0 |
| **Build / other** | +1 | -0 |
| total | +1249 | -6 |

Second of four PRs towards `BatchCPUPowerFlow`, building on #199's `solve_transpose`. This one makes a batch differentiable in C++; the pytorch class and the flows come next, and the `gen_v` gradient after that. Nothing existing changes behaviour: `keep_jacobian` is off by default and the per-row hook is a no-op when it is.

## Why

For a scalar loss `L(V_1..V_n)`, the implicit function theorem gives the gradient with respect to **every** injection of **every** row from one transposed solve per row. λ does not depend on which input is being differentiated, so the cost is independent of how many inputs there are — and of how wide `V` is.

Two properties of this repository's batch algorithms are what make that cheap, and they are why this lives here rather than in a generic autodiff layer:

- **one sparsity pattern for the whole batch** — a row may change a bus's role, but only by rewriting values inside a pattern reserved up front. So one pattern is snapshotted and each row is just `nnz_J` reals.
- **KLU answers the transposed system out of the factorization of J itself** (#199). No transposed copy, no second symbolic analysis — the machinery a GPU library needs here collapses to a flag.

## What

**`BatchAdjoint`** (new, `src/core/batch_algorithm/`): one pattern, one `(nb_rows × nnz_J)` **row-major** value buffer sized **once** before the row loop — everything it needs is known by then, the row count is locked and the warm-up solve has built the ledger — and a threaded `solve_JT` giving each worker its own solver and working matrix. A row is one memcpy, one refactorize, one triangular solve. Row-major so a row is contiguous and workers never share a cache line except at their range boundaries. It knows nothing about powerflows: it is a batch of same-pattern linear systems.

`xbar` is `(nb_rows, k · dim_J)` — k cotangents laid end to end, the layout `klu_tsolve`'s `nrhs` wants. k = 1 for a scalar loss; more only for a full Jacobian, and they cost one triangular solve each rather than one factorization each.

**On the sweep**: `keep_jacobian`, `solve_JT(xbar)`, `adjoint_row_ok()`, `adjoint_memory_bytes()`, `dim_J()`, and the ledger maps **re-keyed onto grid bus ids** so callers never have to reconstruct the solver's internal labelling. All four instantiations get them.

**What is deliberately not stored** is the numeric factorization. `klu_refactor` overwrites its `klu_numeric` in place and there is no way to clone one, so keeping a factorization per row would mean a full `klu_factor` — more expensive than a refactor — per row in the *forward*. Backward therefore pays one refactorize + one transposed solve per row, and the two are timed apart in `adjoint_solver_stats()` so that trade can be revisited on measurements rather than on belief.

## Three things the maths needs that the sweep alone would get wrong

- **Newton-Raphson stops on the Jacobian of its previous iterate** (it tests the residual before deciding it needs a new one). The adjoint is exact only at the solution, and the error is of the order of the last Newton step — which is not what the tolerance bounds. `refresh_J_at_solution()` re-evaluates it where it belongs, for one Jacobian fill and no factorization.
- **A row's Jacobian is captured while that row's own state is still installed** — its Ybus edits, its masked buses, its PV pinning. Refreshing after the loop restored its resting state would describe a system the row never solved.
- **The multiplier of an identity-frozen equation is dropped** (a bus a contingency stranded, a bus still PV in this row). Such a row of J is a single 1, so it feeds no other equation and the rest of λ is unaffected — but reporting it would put a gradient on an injection the solution does not depend on.

## Drive-by fix

`KLUSolver.hpp`, `NICSLUSolver.hpp` and `CKTSOSolver.hpp` closed their namespace **outside** their include guard, so including one twice in a translation unit emitted a stray `}`. Latent until now because nothing included them after `Solvers.hpp` did; `BatchAdjoint.hpp` does.

## Tests

Two levels, because this can be wrong in two independent ways.

- **`BatchAdjoint` alone**, against a dense LU of each row's explicitly transposed matrix — which owes nothing to powerflows. Plus: same answer at 1 and 4 threads on a row count that divides neither; a row the batch never solved comes back zero and says so; identity-pinned multipliers dropped and nothing else touched; a foreign pattern refused; k = 3 cotangents costing 3 solves and 1 analysis.
- **End to end on a grid**, against a central finite difference of a scalar loss — which owes nothing to Jacobians. A flipped implicit-function sign, a ledger read backwards, or a Jacobian captured one iterate early would each show up here. Covers load P, load Q (a PQ bus) and generator P (a PV bus, the other sign). The loss is linear in Re/Im rather than `|V|²`, whose angle cotangent is identically zero and would leave half the ledger untested.
- **The refresh has its own test**, because the gradient test cannot stand in for it: at a loose tolerance the two Jacobians differ measurably (and refreshing twice changes nothing), while at the tolerance the gradient test uses they agree well past its margin — verified by making the hook a no-op, which fails this test and not that one.

## Verification

- C++ suite **287/287**, at the default standard and at a pinned C++14 build.
- Python batch suite, all four instantiations: `test_InjectionSweep` (15), `test_ScenarioSweep` (21), `test_ScenarioSweep_gen_contingency` (16), `test_TimeSerie` (2), `test_timeseries_sbus` (5), `test_SecurityAnlysis_cpp` (12), `test_ContingencyAnalysis_split` (14), `test_ContingencyAnalysis_limit_violations` (37) — **122 tests, all pass**, against grid2op installed from source.
- Extension built and exercised from Python end to end: the adjoint gradient matches a finite difference to 10 significant digits, with `nb_analyze = 1` and one refactorize per row after the first.

---

### DIFF
--- a/src/core/batch_algorithm/BatchAdjoint.cpp
+++ b/src/core/batch_algorithm/BatchAdjoint.cpp
@@ -1,0 +1,243 @@
+// Copyright (c) 2026, RTE (https://www.rte-france.com)
+// See AUTHORS.txt
+// This Source Code Form is subject to the terms of the Mozilla Public License, version 2.0.
+// If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+// you can obtain one at http://mozilla.org/MPL/2.0/.
+// SPDX-License-Identifier: MPL-2.0
+// This file is part of LightSim2grid, LightSim2grid implements a c++ backend targeting the Grid2Op platform.
+
+#include "BatchAdjoint.hpp"
+
+#include <algorithm>
+#include <sstream>
+#include <stdexcept>
+#include <thread>
+
+namespace ls2g {
+
+std::unique_ptr<IAdjointLinearSolver> make_adjoint_linear_solver()
+{
+    #ifdef KLU_SOLVER_AVAILABLE
+    return std::unique_ptr<IAdjointLinearSolver>(new AdjointLinearSolver<KLULinearSolver>());
+    #else
+    return std::unique_ptr<IAdjointLinearSolver>(new AdjointLinearSolver<SparseLULinearSolver>());
+    #endif
+}
+
+void BatchAdjoint::allocate(Eigen::Index nb_rows, const Eigen::Ref<const Eigen::SparseMatrix<real_type> > & J)
+{
+    if(nb_rows <= 0){
+        clear();
+        return;
+    }
+    if(J.rows() != J.cols()){
+        std::ostringstream exc_;
+        exc_ << "BatchAdjoint::allocate: the Jacobian must be square, got " << J.rows()
+             << " x " << J.cols() << ".";
+        throw std::runtime_error(exc_.str());
+    }
+    // the pattern, values zeroed: a worker copies this to get a working matrix whose
+    // index arrays it never has to rebuild
+    pattern_ = J;
+    real_type * val = pattern_.valuePtr();
+    std::fill(val, val + pattern_.nonZeros(), static_cast<real_type>(0.));
+
+    J_values_ = RealMatRM::Zero(nb_rows, pattern_.nonZeros());
+    row_has_J_.assign(static_cast<size_t>(nb_rows), 0);
+    stats_ = LinearSolverStats();
+}
+
+void BatchAdjoint::clear()
+{
+    pattern_ = Eigen::SparseMatrix<real_type>();
+    J_values_ = RealMatRM();
+    row_has_J_.clear();
+    stats_ = LinearSolverStats();
+}
+
+void BatchAdjoint::store_row(Eigen::Index row, const Eigen::Ref<const Eigen::SparseMatrix<real_type> > & J)
+{
+    if(row < 0 || row >= J_values_.rows()){
+        std::ostringstream exc_;
+        exc_ << "BatchAdjoint::store_row: row " << row << " is out of the [0, "
+             << J_values_.rows() << "[ range this adjoint was allocated for.";
+        throw std::runtime_error(exc_.str());
+    }
+    // The whole premise of a batch is that this never fires (see the class comment).
+    // It is checked anyway, on every row: a pattern that drifted would not throw, it
+    // would quietly return a gradient computed from somebody else's matrix -- and the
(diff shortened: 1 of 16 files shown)


src/core/batch_algorithm/BatchAdjoint.cpp
```diff
@@ -0,0 +1,244 @@
...
+#include <sstream>
+#include <stdexcept>
+#include <thread>
+
+namespace ls2g {
+
+std::unique_ptr<IAdjointLinearSolver> make_adjoint_linear_solver()
+{
+    #ifdef KLU_SOLVER_AVAILABLE
+    return std::unique_ptr<IAdjointLinearSolver>(new AdjointLinearSolver<KLULinearSolver>());
```
BDonnot: std::make_unique() is better here

src/core/batch_algorithm/BatchAdjoint.cpp
```diff
@@ -0,0 +1,244 @@
...
+    #else
+    return std::unique_ptr<IAdjointLinearSolver>(new AdjointLinearSolver<SparseLULinearSolver>());
```
BDonnot: Here too and everywhere you call new I think

src/core/batch_algorithm/BatchAdjoint.cpp
BDonnot: Agreed — done in b78d721. `std::make_unique` is C++14, so it sits inside this project's floor, and it is what the rest of the core already uses (`BuiltinSolversRegistration.cpp`).

The derived `unique_ptr` converts to the base one on return, which is well defined here since `IAdjointLinearSolver` has a virtual destructor.

---

src/core/batch_algorithm/BatchAdjoint.cpp
BDonnot: Done in b78d721. These two were the only `new` in the PR — I grepped the three new files to check, and the rest of the diff allocates nothing by hand.

---

PR #201, opened 2026-09-12
differentiable: a batch of powerflows as a pytorch operation

BDonnot:
| | added | removed |
|---|---|---|
| **Python (package)** | +548 | -0 |
| **Tests** | +273 | -0 |
| **Docs + changelog** | +99 | -0 |
| **Build / other** | +3 | -0 |
| total | +923 | -0 |

Third of the series, on top of #199 (`solve_transpose`) and #200 (`BatchAdjoint`). This one is the layer an ML user actually touches; no C++ changes at all.

```python
pf = BatchCPUPowerFlow.init_from_grid(grid, nb_thread=8)
V = pf(load_p=load_p, line_status=status)          # complex (n_scenarios, n_bus)
pf.compute_flows(V).i_or_a.relu().sum().backward() # -> load_p.grad
```

`load_p`, `load_q`, `gen_p`, `sgen_p` carry gradients. `line_status` / `trafo_status` / `gen_status` are discrete (True = connected, grid2op's convention). `gen_v` is accepted as a fixed set-point.

## The one design choice worth reviewing

The injections reach the solver through the sweep's **own `modify_*` setters**, not as an assembled Sbus, and the element tensors are the autograd leaves. That is what keeps this layer thin: everything that makes the assembly subtle — a generator contingency re-weighting the distributed slack, a bus released from PV to PQ when its last regulating machine goes — stays in the C++ the forward already runs. Backward only undoes the last, affine step from a per-unit bus injection back to the elements on it (a load subtracts, a generator adds, both over `sn_mva`).

## Three things this layer owes the caller beyond calling C++

- **A row that did not converge is NaN, not zeros.** Zeros are a plausible-looking voltage and would quietly poison a loss; `converged()` says which rows they are.
- **A call's result depends only on its own arguments.** The sweep keeps whatever masks it was last given, so a call that drops one says so explicitly (an all-false mask), and dropping `gen_v` — which has no "nothing set" value to send — restarts the batch instead. Both directions are tested.
- **`gen_v` is rejected if it requires a gradient** rather than silently getting none: that term needs the dS/dVm column the Jacobian does not store for a voltage-fixed bus. It is the next PR.

`compute_flows` is `BaseBatchSolverSynch::_flows_of_row`'s formulas written in torch, so a loss on a current — an overload penalty, the usual reason to want any of this — reaches the injections.

torch stays **optional**: imported on first use, never at module scope, and a test pins that importing lightsim2grid (or this subpackage) does not pull it in.

## Tests

14, all against references that owe nothing to the thing under test.

- Every gradient is checked against a **central finite difference of the same loss**, re-running the batch: the plain case, one through the currents, one with lines tripped, one with a generator tripped, plus `gradcheck`.
- The torch flows are pinned against the **C++ flows on the same voltages**, so a wrong flow cannot masquerade as a wrong gradient.
- Plus: diverging rows (NaN, zero gradient), threads not changing the answer, the scenario count changing between calls, both call-to-call independence cases, and the input-validation errors.

The loss is linear in Re/Im rather than `|V|²`, whose angle cotangent is identically zero and would leave half the ledger untested.

**Mutation-checked**: flipping the `load_p` gradient sign fails 5 tests; removing the stale-mask reset fails exactly the independence test; removing the `gen_v` restart fails exactly its own test.

## Verification

- The 14 new tests pass, against torch 2.14 and grid2op installed from source.
- `test_ScenarioSweep` (21) and `test_InjectionSweep` (15) still pass — this PR is purely additive.

One caveat: the new tests use pandapower's case14 as their fixture and skip without pandapower. CI has it, but any job that doesn't will silently skip the whole file.

---

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -156,6 +156,9 @@ TODO: a "combine mode" axis for ``ScenarioSweepCPP`` choosing between the curren
 
 [1.0.1] 2026-xx-yy
 --------------------
+- [ADDED] ``lightsim2grid.differentiable.BatchCPUPowerFlow``: a batch of powerflows as a
+  pytorch operation, differentiable with respect to the injections through the adjoint of
+  :func:`solve_JT`. ``pip install lightsim2grid[torch]``; torch stays optional.
 - [ADDED] ``keep_jacobian`` and ``solve_JT`` on the batch algorithms: the adjoint of a whole
   sweep, one transposed solve per row, which is what reverse-mode differentiation (a pytorch
   backward) needs. See ``BatchAdjoint``.
--- a/docs/differentiable.rst
+++ b/docs/differentiable.rst
@@ -1,0 +1,95 @@
+Differentiable powerflow (pytorch)
+===================================
+
+.. _differentiable:
+
+A batch of powerflows, differentiable with respect to its injections, as a
+pytorch operation: a loss computed on the results of thousands of AC powerflows
+can be back-propagated to the loads and generators that produced them.
+
+.. code-block:: python
+
+    import torch
+    from lightsim2grid.network import init_from_pandapower
+    from lightsim2grid.differentiable import BatchCPUPowerFlow
+
+    grid = init_from_pandapower(some_pandapower_net)
+    pf = BatchCPUPowerFlow.init_from_grid(grid, nb_thread=8)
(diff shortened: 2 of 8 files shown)


BDonnot (commented):
Ok

PR #202, opened 2026-09-12
batch: keep the base case between two compute() calls, behind three nested cache levels

BDonnot:
| | added | removed |
|---|---|---|
| **C++ (src/core, src/bindings)** | +546 | -202 |
| **Tests** | +454 | -0 |
| **Docs + changelog** | +8 | -0 |
| total | +1008 | -202 |

Fourth of the series (after #199, #200, #201), but it is not about differentiation: it speeds up **every** batch class for anyone who calls `compute()` more than once on the same object.

## The waste

Before any row is solved, a batch establishes a base case: read the grid (admittance matrix, bus labelling, pv/pq split), walk the graph to settle what each contingency strands, solve one "n" powerflow, analyze and factorize the Jacobian. None of it depends on the injections — and every `compute()` redid all of it.

| grid | batch | before | after | preprocessing |
|---|---|---|---|---|
| case118 | 8 | 1.01 ms | 0.71 ms | 0.31 → 0.01 ms |
| case2869pegase | 8 | 66.1 ms | 53.9 ms | 11.5 → 0.11 ms |
| case9241pegase | 8 | 269 ms | 221 ms | 46.9 → 0.35 ms |
| case9241pegase | 64 | 1792 ms | 1688 ms | 46.3 → 1.19 ms |

It is a fixed per-call cost, so small batches on large grids gain most — the shape of a training loop. `nb_analyze` over 5 calls: 5 → 1 single-threaded, 9 → 5 at 4 threads.

## One cache, three nested levels

The first two commits kept the base case behind a set of ad-hoc booleans that every modifier had to remember to lower. Nothing said what a given modifier actually invalidates, so each new setter was a fresh judgement call — which is exactly how the segfault in the second commit happened. The third commit names the state instead:

| | drops | on top of |
|---|---|---|
| **L1** `clear_grid_results()` | what was read off the **grid**: Ybus / Bbus, the injections, the bus labelling, the pv/pq split, the slack and its weights | → L2 |
| **L2** `clear_batch_inputs()` | what was built for **this batch**: the per-row Ybus edit lists, what the graph walk settled, the masks, the reserved PV ↔ PQ structure, the "n" solve — **and the algorithm**, whose ledger, sparsity and factorization that solve built | → L3 |
| **L3** `clear_batch_outputs()` | what the last run **produced**: voltages, flows, convergence, limit violations, kept Jacobians, counters, timers | — |

Dropping one drops every level below it, and that nesting is the contract: a modifier names the highest level it invalidates and never enumerates the rest. "The base case may be kept" is then exactly *neither L1 nor L2 was dropped since the last `compute()`* — two booleans instead of five, raised where the corresponding tier is built. Each level is idempotent, which is what keeps `add_n1()` cheap in a loop: only the first of `n_line` calls has anything to do.

The mapping, in full:

- **L1** — `clear()`, `reuse_base_case = False`, `change_algorithm()`, `algo_config`
- **L2** — the contingency registrations (`set_contingency_lines/trafos/gens`, `add_*`, `remove_*`), `handle_disconnected_grid`, `nb_thread`, `init_from_n_powerflow`, `pick_reference_slack()`, a different row count
- **L3** — the injection setters, `violation_threshold`, `keep_jacobian`

Two deliberate mismatches, both noted in the code:

- `change_algorithm()` / `algo_config` are **L1 rather than the level they strictly need**. A different algorithm may be a different *family* (AC ↔ DC), and telling that case apart would buy a cache hit on a call nobody makes in a loop. They are no longer a `clear()` though — a level is not a registration, so the contingencies now survive one.
- `compute_limit_violations` is the converse: **L3 is all it invalidates**, but its full `clear()` is an explicitly tested contract (`test_setter_toggle_clears_and_raises_when_off`), so it stays as it was.

## Three latent bugs the mapping exposed

- `remove_n1()` invalidated **nothing at all** — it dropped the results and left the graph walk's verdicts describing a contingency set that no longer existed.
- `nb_thread` and `init_from_n_powerflow` likewise: the first changes what each worker's solver is warmed up from, the second what every row starts from.
- the registrations dropped the cache **before** validating their argument, so a rejected call left the object rebuilt for nothing.

## Why keeping anything is safe at all

Ownership, not a promise: the class holds its **own copy** of the grid, taken at construction, with no way to modify it. So everything that can change a base case goes through this class, and each of those now names its level. `invalidate_base_case()` stays as the public synonym for `clear_grid_results()`, for the one case this class cannot see — the `change_gridmodel()` still on that header's TODO list.

On by default, with a public `reuse_base_case` to turn off. That switch exists so a suspected caching problem can be told from a real one, which is why it is documented rather than internal.

## Two things worth reviewing

**The starting voltage is not part of what is kept.** A call may start anywhere, so `V` is remapped every time, doing what `LSGrid` does when it builds one: gather the caller's V onto the solver's buses, then snap every regulated bus to its target. Measured at **8.7 µs** on 9241 buses against the 47 ms saved, so there is no case for caching it. A test pins that a second call starting elsewhere converges where a *fresh* object started there would — not where the first call did. That distinction is observable because the reference slack keeps the angle it was started at.

**The algorithm belongs to L2, with the "n" solve.** This is the segfault of the second commit, made unrepresentable. An analysis and a factorization are not results: they are what that solve built and what the next batch would otherwise redo. Dropping them without also dropping the base case that would have used them is a solve against a default-constructed system — the exact failure `LSGrid::_pre_process_own_cache` warns about in its own comment, and the one the limit-violation suite hit. With the levels there is no call that can do it.

## Tests

11 new (7 + 4 for the levels). Most are the same comparison: a batch run twice on one object, against the same batch on a fresh one, with something done in between — a contingency registered, the thread count changed, results dropped, a different row count, a different starting voltage — and each either rebuilds or must not change the answer. The four new ones pin the hierarchy itself: that L3 keeps the base case and L2 does not, that L1 drops both, that a level is not a registration (`clear_grid_results()` keeps the contingencies, `clear()` does not), and that `init_from_n_powerflow` is L2.

## Verification

- C++ **298/298** (287 + 11 new).
- Every batch Python suite, since this changes behaviour for all existing callers: `test_InjectionSweep`, `test_ScenarioSweep`, `test_ScenarioSweep_gen_contingency`, `test_TimeSerie`, `test_timeseries_sbus`, `test_SecurityAnlysis`, `test_SecurityAnlysis_cpp`, `test_DCSecurityAnlysis`, `test_ContingencyAnalysis_split`, `test_ContingencyAnalysis_limit_violations`, `test_continuation_powerflow`, `test_hvdc_batch`, `test_voltage_control_batch`, `test_batch_algorithm_solver_selection`, `test_n1contingencyrewards`, `test_BatchCPUPowerFlow` — 236 tests, all pass.

## Not in here

With `nb_thread > 1`, each worker still builds and analyzes its own algorithm per call — only the member one stops re-analyzing. Keeping the workers alive across calls is the other half of this, worth a few percent more on the threaded path, and is its own change.


BDonnot (commented):
Ok

PR #203, opened 2026-09-13
batch: keep the worker algorithms of a threaded batch between calls

BDonnot:
| | added | removed |
|---|---|---|
| **C++ (src/core, src/bindings)** | +107 | -56 |
| **Tests** | +166 | -0 |
| total | +277 | -56 |

The other half of #202 — the part its own "Not in here" section named.

## The hole #202 left

Base-case reuse gave the **single-threaded** path a `compute()` that pays no setup: the member algorithm keeps its ledger, its Jacobian sparsity and its factorization, and every row refactorizes.

A **threaded** batch got none of it. Its rows are run by one algorithm per thread, not by the member one, and those were rebuilt on every call — a fresh `AlgorithmSelector` each, and a first row under `tell_all_changed()`, so `build_J_sparsity` + `analyze` + `factorize` instead of a refactorize. Which is the one thing a batch exists to pay once.

So a threaded batch re-analyzed `nb_thread` Jacobians per call while its single-threaded twin re-analyzed none. At small batch sizes that made threading a **net loss**: 4 rows of case9241pegase took 139 ms on one thread and 53 ms on four, and the four-thread call was spending 16 ms of that rebuilding what it already had.

## Measured

`nb_thread=4`, `batch=32`, arms interleaved over 7 rounds, `reuse_base_case` **on in both** — so this is the saving *on top of* #202, not the two added together.

| grid | before | after | |
|---|---|---|---|
| case118 | 0.88 ms | 0.70 ms | 1.26x |
| case1354pegase | 18.33 | 16.58 | 1.11x |
| case2869pegase | 55.52 | 49.81 | 1.11x |
| case9241pegase | 310.09 | 292.02 | 1.06x |

It is a **flat 4 ms (2869) / 18 ms (9241) per call, independent of the row count** — the signature of a fixed setup cost, and it matches the analyze+factorize per worker measured on its own. So it is worth a third of the call on the small batches a training loop makes, and a few percent on a large one:

| case9241pegase, T=4 | before | after | |
|---|---|---|---|
| N=4 | 52.89 ms | 36.77 | **1.44x** |
| N=8 | 89.76 | 72.15 | 1.24x |
| N=32 | 307.91 | 289.49 | 1.06x |
| N=64 | 607.44 | 591.50 | 1.03x |

End to end against `reuse_base_case` off (everything rebuilt, as before any of this): 63.55 → 49.74 ms on case2869pegase, 354.71 → 289.69 ms on case9241pegase.

## Why it is ~60 lines of C++

The workers are **L2** state, the same level as the member algorithm: each holds a ledger, a sparsity and a factorization built from exactly the batch inputs L2 holds. So `clear_batch_inputs()` drops them, and everything that reaches it — `change_algorithm`, `nb_thread`, a new row count, the contingency registrations — takes them out for free. The hierarchy #202 introduced is doing the work; this change is mostly *not* a new invalidation problem.

That is also why there is **no switch of its own**: `reuse_base_case` governs both, there being no state in which keeping one and rebuilding the other is the right answer. It's a one-line change if you'd rather have the separate knob — I used one while measuring, to A/B the two halves in one process, and folded it afterwards.

## Two things deliberately not kept

- **The per-thread Ybus copies** of the contingency path. A row edits its worker's copy and puts it back, so a kept copy would accumulate the rounding of every add/subtract round-trip of every row of every call. Rebuilt per call; it is a memcpy.
- **The workers of a call that threw.** Such a worker is left wherever the exception caught it — mid-row, Ybus edits possibly not restored, masking and PV pinning not put back — so the set is dropped and the next call builds fresh ones.

## The threads themselves are still spawned per call

I built a pool and measured it before dropping it. It is worth a further **~0.15 ms, fixed and grid-independent** — simply what spawning three `std::thread`s costs. Visible only where the whole call is about a millisecond (case118: 1.26x → 1.56x), inside the noise on case9241pegase. Not worth a thread lifetime to reason about.

| nb_thread=4, batch=32 | today | +algos kept | +threads kept |
|---|---|---|---|
| case118 | 0.88 ms | 0.70 | 0.56 |
| case2869pegase | 55.52 | 49.81 | 49.36 |
| case9241pegase | 310.09 | 292.02 | 289.12 |

## Correctness

**Voltages are bit-identical** to the previous behaviour — `max |dV| = 0.0e+00`, not "close" — at every grid size and every batch size measured, with the A/B done in one process so the two arms share a binary.

The 4 new C++ tests pin: that a threaded batch keeps its workers and still gives what a fresh object gives; that it gives what the *same* batch on one thread gives, for 2 and 3 threads; that L1 and L2 drop the workers while L3 does not, and so do a changed thread count and `reuse_base_case = False`; and that a single-threaded batch never claims to have kept workers.

The test that matters is the one that can tell "the workers survived" from "the flag was set": a kept worker's **refactorize counter carries over between calls**, where a rebuilt one is a fresh object and can only ever show the last call's. The analyze counters cannot do it — a rebuilt worker also reports exactly one, its own.

## Verification

- C++ **302/302** (298 + 4 new).
- Python **216 tests** over the 15 batch suites: `test_InjectionSweep`, `test_ScenarioSweep`, `test_ScenarioSweep_gen_contingency`, `test_TimeSerie`, `test_timeseries_sbus`, `test_SecurityAnlysis`, `test_SecurityAnlysis_cpp`, `test_DCSecurityAnlysis`, `test_ContingencyAnalysis_split`, `test_ContingencyAnalysis_limit_violations`, `test_BatchCPUPowerFlow`, `test_hvdc_batch`, `test_voltage_control_batch`, `test_continuation_powerflow`, `test_batch_algorithm_solver_selection` — all pass.

## Noticed, not fixed here

`ContingencyAnalysisCPP` is missing `bind_batch_sweep_common`, so from Python it has no `reuse_base_case`, `invalidate_base_case`, `base_case_was_reused`, `get_linear_solver_stats*`, nor any of the adjoint block (`keep_jacobian`, `solve_JT`, `dim_J`, the row/column maps). The C++ supports all of it — it gets this change too — but a Python caller cannot reach or turn off any of it. Happy to do that as its own PR.

---


BDonnot (commented):
Ok

PR #204, opened 2026-09-13
differentiable: the gen_v gradient, and one bus one magnitude

BDonnot:
| | added | removed |
|---|---|---|
| **C++ (src/core, src/bindings)** | +609 | -2 |
| **Python (package)** | +76 | -20 |
| **Tests** | +656 | -4 |
| **Docs + changelog** | +18 | -0 |
| total | +1359 | -26 |

Fifth of the series (#199, #200, #201, #202, #203): the last input of `BatchCPUPowerFlow` that carried no gradient. Chasing it turned up a silent bug that has nothing to do with differentiation, so the PR carries both.

> Three commits, readable in order: the gradient, then the contradiction it exposed, then the two follow-ups that came out of reviewing it.

---

# 1. Why lambda is not the answer for `gen_v`

An injection enters the Newton-Raphson system as a right-hand side: `dF/dS_spec = -I`, so **lambda IS its gradient**, and every existing gradient in this package is lambda read at the right row and scaled by `sn_mva`.

A voltage set-point is not an injection. It **fixes** the magnitude of the bus its generator regulates, and that magnitude is precisely what the system does not solve for. So `v` appears neither in `x` nor in `Sbus`, and its gradient needs a derivative the Jacobian deliberately does not store for such a bus.

| | |
|---|---|
| **indirect** | `-λᵀ dF/dv`, where `dF/dv_k` is the `dS/d\|V_k\|` column — `gen_v_indirect_grad()` |
| **direct** | the loss's own dependence on `V_k = v_k·e^{jθ_k}` at fixed unknowns: `Re(conj(V_k/\|V_k\|)·ḡV_k)` |

The split has a symmetry worth noticing: **the direct term is the very quantity a PQ bus contributes to `xbar`.** A bus whose magnitude is an *unknown* hands it to the adjoint; a bus whose magnitude is a *set-point* hands it to that set-point's gradient. So the direct half stays in the caller next to the `xbar` code it mirrors, and only the half that needs Ybus crosses into C++.

**No new storage.** I had planned a `(nb_steps, nnz_Y)` complex buffer for the per-row admittance matrices; it isn't needed. The row's edits are already kept as `ybus_policy_.li_coeffs`, so one scratch copy is patched and unpatched per row. And the self term needs no second walk of a column-major matrix: `S_b = V_b·conj(I_b)` makes it `S_b/|V_b|`, with one sparse product per row giving every `I_b`.

The **slack bus is in, not out**: its magnitude is fixed too, so a generator regulating it carries a gradient on the same terms as one on a PV bus.

---

# 2. One bus, one magnitude

Two elements regulating the **same** bus with **different** targets state two constraints a single magnitude cannot both satisfy. What happened to them was an accident of iteration order: `set_vm` writes each in turn, the last one visited wins, the others are silently discarded. Nothing makes the last machine of a busbar more authoritative than the first, and the answer would change if the elements were reordered.

`VoltageControlPlan` already refused exactly this — but only for a *group*, and a group forms around a **remote** regulator or an SVC. `build_groups` says so outright:

> *A purely local regulator therefore never lands here, which is what keeps the ordinary (possibly multi-generator) PV bus untouched.*

So the ordinary case — several machines on one busbar, each regulating the bus it stands on — was the one nobody checked.

**Two data sources, so two checks:**

| | where | what happens |
|---|---|---|
| the grid's own `target_vm_pu` | `LSGrid::_check_vm_targets_agree`, from **both** `_pre_process_own_cache` and `_build_foreign_cache`, and from `check_grid()` | throws, naming both elements, both values, the remedy |
| a batch's per-row `gen_v` | `BaseBatchSweep::_row_gen_v_conflicts` | that row is not solved — reported like any row skipped before the solver |

The two entry points are what make it hold for LSGrid's own powerflows **and** every batch class. The per-row one is a divergence rather than a throw because the grid never sees those values and one bad row must not take a sweep down — matching how an islanding contingency is handled. Both messages carry `conflicting voltage setpoints`, so one grep finds either.

⚠️ **Marked `[BREAKING]`**: a grid that "worked" before can now be refused. Checked against every grid in the suite — `ACPF`, `DCPF`, `LSGrid_pandapower`, `n1contingencyrewards`, `basic_backend_api` and the batch suites all pass, and not one of `test_LightSimBackend`'s 49 pre-existing errors (the documented missing grid2op test data) mentions a set-point.

**Two test fixtures changed, because they *were* the contradiction.** Two KCL cases moved a generator onto a converter station's bus (1.045 vs 1.0 pu) without touching the set-point, and passed because KCL balances whichever magnitude won. They now agree with the station — which is exactly what the SVC case three tests below already did: `change_v_gen(1, 1.0); // agree with the SVC's setpoint`. The repo already knew the invariant; two tests hadn't applied it.

---

# 3. The two follow-ups

### The check was not everywhere yet

`_apply_step_gen_v` writes only **generators**, so a per-row `gen_v` silently overwrote the target a co-located SVC or hvdc converter station had already applied — the same last-writer-wins one layer up, and worse because the row still *converged*, at a set-point nobody asked for. The per-row check now reads what those elements pin and refuses a row unless the generators on their bus ask for exactly that.

That limitation is now written down where it is met (the `modify_gen_v` comment, its binding docstring, the python module doc) and as a **TODO at the top of the changelog**: only **generator** set-points vary per row. There is no `modify_svc_v` and no `modify_hvdc_v`, so a generator sharing its regulated bus with either is fixed for the whole sweep and its `gen_v` gradient is zero.

### Tied set-points hold a share, not a partial each

Generators regulating one bus must be given the same value, so the loss is a function only **on the diagonal** `v₁ = … = vₙ`. Off it there is no value to compare against — such a row is refused, not solved differently. **The partial derivative of one set-point with the others held fixed does not exist**, and what these n numbers are is not a gradient in the usual sense. What exists is the derivative along the tie.

The first commit heaped it on whichever generator `set_vm` writes last. That has the right sum and two wrong properties: the answer depends on the order the generators sit in their container, and a plain gradient step moves one of them alone, leaving the set where the function is defined — so the next row is refused.

Each generator now carries **1/n** of its bus' derivative (`get_gen_v_share`, with the direct half weighted identically). Then a caller driving the group from one parameter — which is what the degree of freedom really looks like — recovers the derivative exactly through the chain rule; a caller stepping on all n keeps them equal; and nothing is privileged by ordering.

I considered NaN-ing them as "undefined", which is more literally honest and matches how this code NaNs a non-converged row, but it poisons the *correct* usage: a single tied leaf has a perfectly well-defined derivative, and NaN would destroy it. The non-existence of the partials is documented at `get_gen_v_share` instead.

---

# Verification

Every gradient is checked against a **central finite difference of the whole sweep** — something that knows nothing about Jacobians, ledgers or adjoint signs. Where the inputs are tied, the finite difference moves the group **together**, which is the only perturbation that stays where the function is defined.

**C++** (`test_gen_v_gradient.cpp`, 6 cases): the PV bus **and** the slack bus, with an assertion the difference is not trivially zero; three generators on one bus — the shares sum to the tied derivative, each holds the same piece; a row asking one bus for two magnitudes is not solved, and only that row; agreeing set-points are not a conflict, at float tolerance; a generator sharing its bus with an hvdc station cannot be moved, one alone on its bus can; a diverged row gets no gradient; DC raises.

Plus, in `test_voltage_control_reclassify.cpp`, the **local** counterpart of the existing remote-regulator case: two machines on one busbar disagreeing are refused, agreeing are fine.

**Python through torch**: every generator of case14 against a finite difference; the same under a line contingency, so the indirect half is taken on the *row's* matrix; `torch.autograd.gradcheck`; a row asking one bus for two magnitudes comes back not converged and NaN; and a grid deliberately doubling up a generator — case14 has one per bus and would have silently skipped the tied-gradient test.

- C++ **312/312** (306 + 6).
- Python: 19 `BatchCPUPowerFlow`, **no skips**, plus `ACPF`, `DCPF`, `LSGrid_pandapower`, `InjectionSweep`, `ScenarioSweep`, `ScenarioSweep_gen_contingency`, `TimeSerie`, `timeseries_sbus`, `SecurityAnlysis`, `SecurityAnlysis_cpp`, `DCSecurityAnlysis`, `ContingencyAnalysis_split`, `ContingencyAnalysis_limit_violations`, `voltage_control_batch`, `hvdc_batch`, `continuation_powerflow`, `n1contingencyrewards`, `basic_backend_api`, `Computers`.
- CI pre-flight: full C++14 configure/build/link + tests, libc++ with `_LIBCPP_REMOVE_TRANSITIVE_INCLUDES`, clang C++26 — all clean.

## Behaviour change

`forward(gen_v=...)` with `requires_grad=True` used to raise `NotImplementedError`; it now returns a gradient. The test asserting the old refusal is replaced by one asserting the set-point still pins the magnitude it should.

## Follow-up

#205 (stacked on this) fixes `ContingencyAnalysisCPP`, which never got `bind_batch_sweep_common` and so reaches none of this from python.


lightsim2grid/differentiable/_batch_cpu_power_flow.py
```diff
@@ -20,11 +20,41 @@
...
+whole sweep and its ``gen_v`` gradient is zero. See the TODO at the top of
+CHANGELOG.rst.
+
+``gen_v`` is differentiated differently from the other four, because it is not an
+injection: it FIXES the magnitude of the bus its generator regulates, which the
+Newton-Raphson therefore does not solve for. So lambda is not its gradient, and the
+two halves are put together by hand -- the ``dS/dVm`` column the Jacobian does not
+store for such a bus (``gen_v_indirect_grad``), plus the loss's own dependence on
+``V_k = v_k . exp(j.theta_k)`` at fixed unknowns. A generator whose set-point never
+reaches the solve -- disconnected, not regulating, overwritten by another generator
```
BDonnot: It cannot be overwritten now

lightsim2grid/differentiable/_batch_cpu_power_flow.py
```diff
@@ -20,11 +20,41 @@
...
+on the same bus, or regulating a bus the solver gives a magnitude to anyway -- gets
+a zero gradient, which is the truth: its set-point is a dead input.
```
BDonnot: Put it after that but you should have at least a warning for the rules when multiple voltage sources are regulating the same bus and how the gradient is split

lightsim2grid/tests/test_BatchCPUPowerFlow.py
```diff
@@ -37,6 +37,19 @@
     PP_AVAILABLE = False
 
 
+def _two_gens_on_one_bus():
+    """case14 with a second generator added on bus 1, regulating it as well. Both
+    set-points then aim at the same bus, and only the last one applied reaches the
+    solve (VoltageSourceContainer::set_vm is last-writer-wins)."""
```
BDonnot: Should not be the correct current behavior

src/core/LSGrid.cpp
```diff
@@ -1460,6 +1475,42 @@ CplxVect LSGrid::_pre_process_own_cache(
...
+        std::ostringstream exc_;
+        exc_ << "LSGrid: " << a.kind << " " << a.el_id << " and " << b.kind << " " << b.el_id
+             << " regulate the same bus (" << bus << ") with conflicting voltage setpoints ("
+             << a.target_vm << " vs " << b.target_vm << " pu). A bus has one magnitude, so "
+                "these two set-points cannot both hold; the powerflow would silently apply "
+                "whichever element it happened to write last. Give them the same target, or "
+                "turn one of the regulators off.";
+        throw std::runtime_error(exc_.str());
+    }
+}
```
BDonnot: 1) can this be refactored in the place where voltages is handled ?
2) this is behind "algo_controler.has_v_changed()" and not done each call?

src/core/LSGrid.cpp
```diff
@@ -1494,6 +1545,11 @@ CplxVect LSGrid::_build_foreign_cache(
             "powerflow, call pre_process_solver / pre_process_dc_solver instead.");
     }
 
+    // Contradictory voltage set-points are the grid's, not the caller's, so they are
+    // refused here rather than left for each batch class to notice (AC only -- a DC
+    // powerflow solves for no magnitude, so nothing there can contradict anything).
+    if(std::is_same<MatScalar, cplx_type>::value) _check_vm_targets_agree();
```
BDonnot: Wrong here, VM are seeded from generators in DC mode too !

BDonnot:
## ⚠️ I deleted the pending review by accident — reproducing it here

Replying inline was blocked by "user_id can only have one pending review per pull request", and I called `delete_pending` assuming the pending review was a stale one of my own. It was **BDonnot's**, and deleting it took its five comments with it. My mistake, and not recoverable through the API — so here are all five verbatim, with what I did about each. Sorry for the noise.

---

**1.** `lightsim2grid/differentiable/_batch_cpu_power_flow.py:55` — *"It cannot be overwritten now"*

Correct, and stale from before the gradient model changed. The line said a generator "overwritten by another generator on the same bus" gets a zero gradient. That stopped being true when tied set-points became a shared `1/n` — they are not overwritten, they must agree, and each carries a share. Rewritten.

**2.** `lightsim2grid/differentiable/_batch_cpu_power_flow.py:57` — *"Put it after that but you should have at least a warning for the rules when multiple voltage sources are regulating the same bus and how the gradient is split"*

Done. The rules now sit **after** the explanation of how `gen_v` is differentiated, as a `.. warning::` block in two halves: *what you may pass* (generators on one bus must agree; one sharing its bus with a voltage-mode SVC or an hvdc converter station cannot be moved at all, since only generator set-points vary per row) and *what the gradient means* (the partials do not exist, each generator holds `1/n` of the derivative along the tie, and what that implies for driving the group from one parameter versus stepping on all of them).

**3.** `lightsim2grid/tests/test_BatchCPUPowerFlow.py:42-43` — *"Should not be the correct current behavior"*

Correct — same staleness. The `_two_gens_on_one_bus` docstring still said "only the last one applied reaches the solve (`set_vm` is last-writer-wins)". Rewritten to say they must agree and share the bus' derivative.

**4.** `src/core/LSGrid.cpp:1479-1512` — *"1) can this be refactored in the place where voltages is handled? 2) this is behind `algo_controler.has_v_changed()` and not done each call?"*

Both done, and both were right.

It now sits inside `_build_into_cache`'s `init_pv_vm_targets` block, immediately above the three `set_vm` calls whose last-writer-wins behaviour *is* the problem — one site instead of the two entry points, still covering this grid's own powerflows and every batch class, and naturally skipped by `check_solution`, which must take the caller's voltage as given.

And it no longer runs per solve: only when the cache is being rebuilt anyway, or when `has_v_changed()` / `has_pv_changed()` says something moved a set-point or changed who regulates what. An injection-only step — a batch row, a grid2op episode — pays nothing. New test `a set-point moved BETWEEN two solves is still caught` covers the case that gating could have swallowed.

**5.** `src/core/LSGrid.cpp:1548-1551` — *"Wrong here, VM are seeded from generators in DC mode too!"*

Right, and this was the real bug of the five. DC seeds |V| from the generators through that same block, and `_dc_base_vm_solver_` echoes it straight back as the result's magnitude — so two set-points on one bus are just as silently resolved there. The family gate is gone, and `a DC powerflow refuses the contradiction too` pins it.

---

All five in 1fed91f. C++ **314/314** (312 + 2 new); Python re-run across `BatchCPUPowerFlow`, `ACPF`, `DCPF`, `LSGrid_pandapower`, `ScenarioSweep`, `ScenarioSweep_gen_contingency`, `voltage_control_batch`, `hvdc_batch`, `DCSecurityAnlysis`, `ContingencyAnalysis_limit_violations`, `n1contingencyrewards`, `basic_backend_api`, `continuation_powerflow`.

---

src/core/LSGrid.cpp
```diff
@@ -1460,6 +1475,42 @@ CplxVect LSGrid::_pre_process_own_cache(
     return V;
 }
 
+
+void LSGrid::_check_vm_targets_agree() const
```
BDonnot, line 1487: This could also be heavily optimized.... 

Allocate a vector of float (or eigen vect) of the size of the active bus in the grid.
Fill it with -1 (or whatever negative, q VM target cannot be negative)
Then walks to every container that can regulate voltage.

For each of them:
- if the corresponding value of this vector of VM is -1, put the value of VM for this elements 
- if this value is not -1 and agrees with the VM of this elements, good
- if this value is not -1 and disagrees : conflicts 

No fancy things, just simple fast O(n elements) algorithm

BDonnot (commented):
Ok

src/core/LSGrid.cpp
BDonnot: Done exactly that, in c13c7bf.

One `std::vector<real_type>` sized by the buses, filled with `-1.`; each regulating element, visited in the order `set_vm` visits them, either finds its bus unclaimed and writes its target there, or finds one and only has to agree with it. O(elements) plus the fill, one allocation, no map, no intermediate list. The `-1` sentinel is doing exactly the job you describe — a target is never negative, so it needs no second array to mark "unclaimed".

The triples existed only to name **both** elements in the message. That now costs a second walk, taken only once a conflict is found, on a path that ends in a throw — so it costs nothing that is ever measured. `collect_vm_targets` and the `VmTarget` struct are gone with them; `for_each_vm_target` hands `(element, bus, target, kind)` to a callback, which is all any caller wanted.

Message unchanged, still naming both:

```
LSGrid: generator 0 and generator 4 regulate the same bus (1) with conflicting
voltage setpoints (1.045 vs 1.075 pu). ...
```

Worth noting: since `init_from_pandapower` calls `check_grid()` itself, a grid like that is now refused **as it loads**, not at the first solve.

C++ 314/314; Python re-run across `BatchCPUPowerFlow`, `ACPF`, `DCPF`, `LSGrid_pandapower`, `voltage_control_batch`, `hvdc_batch`, `n1contingencyrewards`.

---

PR #205, opened 2026-09-13
bindings: ContingencyAnalysisCPP had drifted out of the shared interface

BDonnot:
| | added | removed |
|---|---|---|
| **C++ (src/core, src/bindings)** | +129 | -128 |
| **Tests** | +244 | -0 |
| total | +377 | -128 |

> ### ⚠️ This also carries #204's last commit, which #204 merged without
>
> #204 was merged at `1fed91f`, one commit before `c13c7bf` landed on its branch — the single-pass rewrite of `_check_vm_targets_agree` asked for in [this review](https://github.com/Grid2op/lightsim2grid/pull/204#discussion_r3999814153). So that commit is **not in `dev_1.0.1`**, and it reaches it through here (it was already in this branch's history, since this was stacked on #204). That is the `+129 / -128` in the C++ row; the binding change itself is `+68 / -47`.
>
> It is the version you specified: one `std::vector<real_type>` sized by the buses, filled with `-1.`, each regulating element either claiming its bus or agreeing with what it finds. O(elements) plus the fill, one allocation, no map, no intermediate list. `collect_vm_targets` and the `VmTarget` struct are gone; naming both elements in the message costs a second walk, taken only after a conflict is found, on a path that ends in a throw.
>
> Say the word if you would rather have it as its own PR instead.

---

## The bug is an absence

Every batch class is the same C++ template, but only three of the four shared a binder: `ContingencyAnalysisCPP` spelled its python interface out by hand. So every feature added to `bind_batch_sweep_common` silently skipped it, and by now it was missing — **from python only** —

- **base-case reuse**: `reuse_base_case`, `invalidate_base_case`, `base_case_was_reused`, `thread_algos_were_reused` (#202, #203)
- `get_linear_solver_stats`, `get_linear_solver_stats_per_algo`
- **the whole reverse-mode block**: `keep_jacobian`, `solve_JT`, `dim_J`, the four bus → row/column maps, `adjoint_row_ok`, `adjoint_solver_stats`, `adjoint_memory_bytes`, and #204's two `gen_v` accessors (#200, #204)

None of it was a C++ limitation. `ContingencyAnalysis` is `BaseBatchSweep` like the others and has had all of it working the whole time — reachable from C++ (its own cases in `test_base_case_reuse.cpp` call `base_case_was_reused()`) and from nowhere else. An N-1 study is the most-used batch class in this repo, and it was the one that could not turn the cache off to chase a suspected caching bug, nor differentiate at all.

## Fixing the cause, not the symptoms

Adding the missing names one by one would leave the mechanism in place for the next feature. So the shared half moves into `bind_batch_shared()`, which both call sites now use.

What stays hand-written for `ContingencyAnalysisCPP` is what genuinely differs: the contingency-list API, the limit violations, and its own wording for the properties it shares (its `init_from_n_powerflow` talks about line disconnections, not scenarios) — so no existing docstring changes.

## The guard

The first new test compares the two interfaces and insists the only names missing are the **six the C++ really gates off**:

```python
VARY_ONLY = {"modify_gen_p", "modify_sgen_p", "modify_load_p", "modify_load_q",
             "modify_gen_v", "get_status"}
```

`SbusPolicy::NOOP` — a contingency analysis varies the topology, not the injections, so it has no per-row setter and no aggregate status. Drift now fails a test instead of going unnoticed across releases.

## The rest check it works, not that it is callable

- the base case is kept between two `compute()` calls, and `add_n1()` drops it — with the rebuilt result matching a fresh object;
- `reuse_base_case = False` re-analyzes on every call;
- **the adjoint against a central finite difference**: scale every load by `1 ± ε`, re-run the whole analysis, and compare to `Σ −λ[p_row[bus]]·p/sn_mva` summed over contingencies and loads;
- **the `gen_v` gradient the same way**: shift every set-point by `± ε` and compare to `gen_v_indirect_grad` plus its direct half, with the dead columns asserted zero.

Both references re-run the analysis on a perturbed grid and know nothing about Jacobians or adjoint signs.

## Verification

- C++ **314/314**.
- Python: 6 new, plus `test_BatchCPUPowerFlow`, `test_SecurityAnlysis`, `test_SecurityAnlysis_cpp`, `test_DCSecurityAnlysis`, `test_ContingencyAnalysis_split`, `test_ContingencyAnalysis_limit_violations`, `test_InjectionSweep`, `test_ScenarioSweep`, `test_TimeSerie`, `test_ACPF`, `test_n1contingencyrewards` — all pass.

## Noticed, not fixed

The three cache levels themselves (`clear_grid_results` / `clear_batch_inputs` / `clear_batch_outputs`) are not bound for any class. `reuse_base_case` and `invalidate_base_case` cover what a caller actually needs, so I left them out rather than widen this — say if you'd rather have them.


BDonnot (commented):
Ok

PR #206, opened 2026-09-13
Report the physical limits a converged row leaves (reactive capability, hvdc droop, distributed slack)

BDonnot:
| | added | removed |
|---|---|---|
| **C++ (src/core, src/bindings)** | +1614 | -23 |
| **Python (package)** | +368 | -30 |
| **Tests** | +1911 | -2 |
| **Docs + changelog** | +278 | -3 |
| total | +4171 | -58 |

## What this adds

An opt-in, per-row check on the four batch algorithms: `compute_physical_violations`. Every converged row reports the limits whose violation means the row is **not a state the grid can reach at all** — as opposed to the operational limits `compute_limit_violations` already reports (a voltage band, a thermal rating: states the grid *does* reach and should not sit in).

Three checks, each the condition a PowSyBl OpenLoadFlow outer loop acts on. **Detection only** in all three: no bus is switched PV → PQ, no droop is clamped, no machine leaves the slack distribution, no row is re-solved.

| check | reported as | OLF loop |
|---|---|---|
| the **reactive capability** of a bus whose voltage is held by machines — did it need more than the *sum* of what its voltage-regulating generators, hvdc converter stations and voltage-mode SVCs can produce? | `LOW_Q` / `HIGH_Q` on the `BUS` | `ReactiveLimits` |
| the **active power** of an angle-droop ("AC emulation") hvdc line still in the linear regime — did `p0 + k·(θ₁ − θ₂)` leave `pmax_1to2_mw` / `pmax_2to1_mw`? | `HIGH_P` on the `HVDC`, `side` naming the direction | `HvdcAcEmulationLimits` |
| the **active power** of a generator carrying the distributed slack — did `target_p + its share of the imbalance` leave `min_p_mw` / `max_p_mw`? | `LOW_P` / `HIGH_P` on the `GENERATOR` | `DistributedSlack` |

Available on all four instantiations, unlike `compute_limit_violations` (contingency-only): a time series walking a load curve is exactly as likely to ask a bus for reactive power it does not have as a contingency is.

## Why each one is a *physical* limit

A voltage-regulating machine has no reactive setpoint: its Q is solved for, and lightsim2grid never clamps it. `status_droop` is an **input** of the solve, so nothing saturates an hvdc droop on its own. And the distributed slack is solved **inside the Jacobian** (`MultiSlack`), by fixed participation factors that know nothing about limits. In all three cases the row can converge describing equipment doing something it cannot do — the solution is the answer to a question with a constraint left out, not a state the grid reaches.

## Why reactive is per bus and active is per machine

Not an inconsistency — the two splits are different in kind.

The reactive power a *bus* needs is a fact about the solution. How it is divided between several machines on that bus is not: the solver never decides it, `LSGrid::_split_q_residual_per_bus` does, by a sharing convention (proportional to each machine's range). A per-machine reactive check would therefore report the convention — two 20 MVAr machines covering 30 MVAr together is feasible, and the same solution read machine by machine shows both 5 MVAr "over". `test_batch_physical_violations.cpp`'s *"a bus is fine while one of its machines is not"* builds exactly that case, checks against `ac_pf` that a per-machine reading **would** fire, and requires that nothing be reported.

The active split is not a convention: each generator's participation factor is input data the caller chose (`add_gen_slackbus`), and the solver distributes by it. So `target_p + share` is that machine's own power and its own limits are what to compare it against. *"each machine is judged at its own share of the imbalance"* pins this with a 1:3 participation, where any other split agrees with neither machine.

## Why the report grew a category

Three kinds of statement were being carried by one `LimitViolation` type, and they must not be acted on the same way. `ViolationCategory` is derived from `violation_type` (a pure function plus a `category` property) so the two cannot disagree:

| | | |
|---|---|---|
| `OPERATIONAL` | `LOW_VOLTAGE`, `HIGH_VOLTAGE`, `CURRENT` | a limit an operator chose, which the grid *can* leave: a reachable state, just not one to sit in |
| `PHYSICAL` | `LOW_Q`, `HIGH_Q`, `LOW_P`, `HIGH_P` | a limit of the equipment, which nothing can leave: the converged solution is **not** a state the grid can reach. A statement about the model's assumptions, not about how the grid is operated |
| `SOLVER` | `NOT_SIMULATED`, `DIVERGENCE` | not a limit at all. A divergence does not distinguish "no solution exists" from "this algorithm did not find one", which is why it cannot sit with the other two |

On the python side `run()` returns these in a **separate** `physical_violations` field rather than mixed into `limit_violations` — same reasoning. The field is last in both dataclasses and defaults to an empty list, so positional construction and callers that never enable the check are unaffected.

## How each value is obtained

A batch keeps the voltages and drops everything else, so each check re-derives its quantity per row, from state that has to be read **while the row's own state is still installed on the algorithm** — next to `_maybe_store_jacobian` and for the same reason.

**Reactive.** `q_bus = imag(mis_bus(b)) + Σ Q_c of the controllers standing on b`. `mis_bus` is the algorithm's own mismatch (`BaseAlgo::get_bus_mismatch`), which carries `−i·Q_c` at every `VoltageControl` controller's bus; adding those back restores the raw reactive residual, and the raw residual *is* what the machines pinning that bus produced. One expression covers an ordinary PV bus and a bus a control group holds. Reading the algorithm's own mismatch is what makes it cheap (no extra sparse product) and what makes it correct: it already nets out the hvdc droop flows and the controllers' own Q, which a raw `Ybus·V` would charge to the generators twice. A voltage-mode SVC contributes `b·|V|²·sn_mva` at the solved voltage — sign convention included: a positive susceptance is capacitive and *produces*, so `[0, b]` can only produce and `[−b, 0]` only absorb.

**HVDC.** `p = p0 + k·(θ₁ − θ₂)` from the converged angles, compared against the *sending* end's own maximum in each direction. Needs nothing but `get_Va()`, so it works in DC too. A line the caller has already saturated (`status_droop ≠ 0`) is skipped: its flow is pinned at the very limit this would compare against.

**Generator.** `target_p + node_mismatch(bus)·w_g / raw_w(bus)`, mirroring `GeneratorContainer::set_p_slack`. The subtlety here — and the reason the plan carries a `participants` list next to the reportable `gens` — is that **two per-bus weight vectors are in play and they are not the same one**: the vector handed to the solver is *normalized* (which is what makes `slack_absorbed·w(bus)` the share that bus absorbed), while the denominator `set_p_slack` divides by is the *raw* per-bus sum. Only the normalized one reaches a batch row, so the raw total is recovered per row as `w_norm(bus)·total_raw_w` — which is also what keeps a row that disconnects a participating machine right. A first version used the normalized vector directly and was a clean factor of two off; the `ac_pf` pin caught it. Works in DC as well, where the imbalance is shared out after the solve (`_fill_bus_mismatch_dc`).

Every reported value in every test is pinned against a single-shot `ac_pf` (or `dc_pf`) on the same grid.

## New input: optional generator active power limits

The model stored no `min_p` / `max_p`. They are added on the same terms as a branch's thermal rating — **optional, never read by the powerflow**, NaN where a machine has none: `LSGrid::set_gen_p_limits`, `GenInfo.min_p_mw` / `max_p_mw`, read from pandapower's `min_p_mw` / `max_p_mw` and pypowsybl's `min_p` / `max_p`. A grid without them still gets the other two checks and reports nothing for this one.

They are part of the generator's serialized state, so **`BINARY_FORMAT_VERSION` 6 → 7**: a file written by an earlier version no longer loads, and the committed fixture is regenerated. That is the one breaking change in this PR. It also retires half of an existing `[TODO]`: `is_pseudo_off` wanted a stored `min_p` to match OLF's `generatorsWithZeroMwTargetAreNotStarted`, and now has one.

## What is refused, and what is not

An AC algorithm that publishes no per-bus mismatch (every built-in family does — NR, fast-decoupled *and* Gauss-Seidel; a plugin must opt in) raises at `compute()` rather than reporting "no violation". **DC is not refused**: a DC powerflow has no reactive power at all, so that check is not applicable rather than missing, and the two active-power ones still run.

One cache subtlety worth a reviewer's eye: the base-case report is **level 2, not 3**. A `compute()` that reuses the base case does not re-solve it, so re-deriving the "n" report there would report the previous call's last row as the base case. It is kept with the base case, and both setters drop it. There is a test that fails without this.

## Scope

Opt in, so nothing changes unless asked for. With the flag off the only touched behaviour is a few added clears; the enums gained values, no existing one was renumbered, and physical violations never land in `get_violations()`.

## Tests

- **C++**: 338/338 (25 new cases in `test_batch_physical_violations.cpp`). One machine, two with asymmetric ranges, remote regulation through the controller list, an hvdc station and an SVC contributing to a bus, a droop in both directions and one read through a shifted bus labelling, a contingency that moves the flows, a skipped row, a row that disconnects a machine, a second `compute()` on a reused base case, Gauss-Seidel against Newton-Raphson, and for the slack: each machine at its own share, a non-participant, a machine with no limit, a per-row set-point, a contingency, and DC.
- **Python**: `test_physical_violations.py` (25 tests) pins the same values through the bindings and covers the wrapper contract and the pandapower columns. Re-run green: `test_ContingencyAnalysis_limit_violations`, `test_TimeSerie`, `test_InjectionSweep`, `test_ScenarioSweep`, `test_ScenarioSweep_gen_contingency`, `test_dc_dist_slack`, `test_multi_slack`, `test_multiple_slack`, `test_dist_slack_jacobian` (108 tests), plus `test_binary_serialization` / `test_pickleable` for the format bump.
- **Not run locally**: the pypowsybl path — `pypowsybl` is not installed in this environment, so `min_p` / `max_p` reaching `GenInfo` through `_aux_add_generators` is covered by CI only.

## Also in here

`docs/dev_notes/outer_loop_checks_missing_data.md` — a status note, not built documentation. Loop by loop, what a post-solve "would this outer loop have fired?" check needs versus what the `LSGrid` holds. The short version: the result side is complete everywhere, the three physical loops are now done, and every remaining gap is control data lost **in the converters** — the RTC/PST tap range, regulating flag, target and deadband, and the shunt sections never reach the grid model, because the tap is folded into the pi-model at load time. It proposes the minimal addition (a pure-input control descriptor per element, no `fillYbus`, no `AlgoControl` flag) and notes that those three checks will report as *operational*, unlike these.

https://claude.ai/code/session_01B3fKmS6YbHhHEqnz6pwksE


docs/dev_notes/outer_loop_checks_missing_data.md
```diff
@@ -0,0 +1,183 @@
...
+What is still missing around it:
+
+- **The PQ → PV direction does not exist.** OLF's loop also switches a bus *back* to PV
+  when its voltage recrosses the set-point on the right side. lightsim2grid never pins a
+  machine at a limit, so there is no such state to detect. This only becomes meaningful
+  the day the limit is *enforced*.
+- **A bus whose reactive power is not the generators' alone is skipped**, because the sum
+  of the generators' limits is then not the bus' capability. Two families put it there:
+  a voltage-regulating hvdc converter station (its capability IS a `[min_q, max_q]` in
+  MVAr — `get_station_q_range_mvar` — so extending the sum to it is easy and is the
```
BDonnot: So I did understand correctly. For HVDC this has to be done too.

docs/dev_notes/outer_loop_checks_missing_data.md
```diff
@@ -0,0 +1,183 @@
...
+  natural next step), and a voltage-mode SVC, whose capability is a susceptance range
+  marked *"stored, NEVER enforced"* in `SvcContainer.hpp` (turning `b_min`/`b_max` into
```
BDonnot: B is the result, bmin and bmax are know (I think) so this part Should also be in this PR.

docs/dev_notes/outer_loop_checks_missing_data.md
```diff
@@ -0,0 +1,227 @@
...
+sign down in the tests — a symmetric one cannot.
+
+**The question is asked per bus, not per machine**, and that is not a shortcut. The
+reactive power a bus needs is a fact about the solution; how it is divided between several
+machines standing on that bus is not — the solver never decides it,
+`LSGrid::_split_q_residual_per_bus` does, by a sharing convention (proportional to each
+machine's reactive range). A per-machine check would therefore report the convention: two
+20 MVAr machines covering 30 MVAr together is feasible, while the same solution read
+machine by machine can show both 5 MVAr "over". So the check compares a bus' reactive
+power against the **sum** of its machines' `[min_q, max_q]`.
```
BDonnot, line 68: To be honest (add it to the TODO list, in the changelog, not implement it now) for voltage controlers that are solved (eg generator remote controling or hvdc station sharing voltage control with another things) it could also be checked directly at the element level (and not only at the bus). But this is a refinment for later.

docs/dev_notes/outer_loop_checks_missing_data.md
```diff
@@ -0,0 +1,227 @@
...
+of the built documentation (`docs/*.rst`).
+
+Reference for the outer loops: PowSyBl OpenLoadFlow,
+`src/main/java/com/powsybl/openloadflow/lf/outerloop/config/DefaultAcOuterLoopConfig.java`
+(the registration order *is* the nesting order, innermost first) and the loop classes in
+`ac/outerloop/`.
+
+## The short version
+
+| OLF outer loop | Result side | Control side | Post-solve check possible? |
```
BDonnot, line 17: I did explicitly say to exclude the distributed slack, but by doing so there is another possible "non physical" issue that could arise here: with the distribution of the slack in the jacobian, some generator p can end up bellow pmin or above pmax.
This should also be implemented here because it's a physical violation solved with outer loops.

BDonnot (commented):
Need to add a pmin / pmax check too

docs/dev_notes/outer_loop_checks_missing_data.md
BDonnot: Done in d03f57e: an HVDC converter station's `min_q_mvar` / `max_q_mvar` now count towards its bus' capability, exactly like a generator's, and a bus is no longer skipped because it carries one. Same in `BusQCheck.hpp`'s plan (`station_ids`, as `(hvdc_id, side)` pairs), and a station whose Q the solver solved for is read off `get_controller_q()` rather than out of the mismatch.

Then, on your later comment, 3912c94 added the *active* side too: an angle-droop line still in the linear regime is checked against `pmax_1to2_mw` / `pmax_2to1_mw` (`HvdcPCheck.hpp`), reported as `HIGH_P` on the `HVDC` with `side` naming the direction.

---

docs/dev_notes/outer_loop_checks_missing_data.md
BDonnot: You were right, they were already there. Done in d03f57e: a voltage-mode SVC's `[b_min, b_max]` contributes `b·|V|²·sn_mva` at the solved voltage to its bus' capability.

Worth noting the sign convention, since it is what the asymmetric test pins down: a positive susceptance is capacitive and *produces* reactive power (a shunt `jb` injects `+b·|V|²` in generator convention), so `[0, b]` is a machine that can only produce and `[-b, 0]` one that can only absorb. A symmetric range would not have caught getting that backwards.

---

docs/dev_notes/outer_loop_checks_missing_data.md
BDonnot: Left as a `[TODO]` rather than implemented, as you asked (bc63017, top of `CHANGELOG.rst`):

> `compute_physical_violations` checks a voltage controller's reactive capability per BUS, which is right for the machines that pin it locally but says nothing per element where the solver DID solve each one: a generator regulating a remote bus, and the hvdc stations sharing one control group, get their own reactive output back (`get_controller_q`), so each could be compared against its own `min_q_mvar` / `max_q_mvar` on top of the bus-level check. Report both, without making the bus-level one per machine: the split between machines the solver did NOT solve for stays a convention (`_split_q_residual_per_bus`).

The plan already carries what that would need — `BusQEntry::ctrl_pos` is each controller's position in the `get_controller_q()` order — so it is an addition to the same loop rather than a new pass.

---

docs/dev_notes/outer_loop_checks_missing_data.md
BDonnot: Implemented in bc63017 — `GenPCheck.hpp`, reported as `LOW_P` / `HIGH_P` on the `GENERATOR`, next to the two existing checks.

Three things worth a reviewer's eye:

**It is per machine, where the reactive check is per bus, and that asymmetry is deliberate.** A bus' reactive power is split between its machines by a *convention* (`_split_q_residual_per_bus`, proportional to each one's range), so a per-machine reactive check would report that convention. The active split is not a convention: each generator's participation factor is input data you chose (`add_gen_slackbus`), so `target_p + share` is that machine's own power and its own limits are the right thing to compare it against. The value mirrors `GeneratorContainer::set_p_slack`, and every test pins it against `ac_pf`.

**The limits did not exist in the model.** They are added on the same terms as a branch's thermal rating — optional, never read by the powerflow, NaN where a machine has none (`LSGrid::set_gen_p_limits`, `GenInfo.min_p_mw` / `max_p_mw`), read from pandapower's `min_p_mw` / `max_p_mw` and pypowsybl's `min_p` / `max_p`. A grid without them still gets the other two checks and reports nothing for this one. They are part of the generator's serialized state, hence **`BINARY_FORMAT_VERSION` 6 → 7** and a regenerated fixture — that is the one breaking bit of this commit.

**The subtlety that bit me**, and the reason there is a `participants` list next to the reportable `gens` in the plan: there are two per-bus weight vectors and they are not the same one. The vector handed to the solver is *normalized* (which is what makes `slack_absorbed · w(bus)` the share that bus absorbed), while the denominator `set_p_slack` divides a machine's factor by is the *raw* per-bus sum. Only the normalized one reaches a batch row, so the raw total is recovered per row as `w_norm(bus) · total_raw_w` — which is also what keeps a row that disconnects a participating machine right. A first version used the normalized vector directly and was a clean factor of two off; the `ac_pf` pin caught it.

Also note this one works in DC (there the imbalance is shared out after the solve, `_fill_bus_mismatch_dc`), so a DC batch now reports both active-power checks.

`docs/dev_notes/outer_loop_checks_missing_data.md`'s `DistributedSlack` row is updated accordingly: it was the odd one of the three, because its control side was never missing — the participation factors are input data and the solver uses them. What was missing was the machines' own active power limits.

---

BDonnot (commented):
ok

PR #207, opened 2026-09-18
Fix derivatives

BDonnot:
Fix some issues by testing the batch algorithm against finite differences


lightsim2grid/network/from_pypowsybl/_aux_battery_apc.py
```diff
@@ -0,0 +1,176 @@
...
+  (default, and for a NaN, :data:`_OLF_DEFAULT_DROOP`) and ``min_target_p`` /
+  ``max_target_p`` (default ``min_p`` / ``max_p``); ``participation_factor`` is not used
+  by this balance type;
+* a battery is excluded when ``|target_p| < 1e-4`` MW (generatorsWithZeroMwTargetAreNotStarted),
+  ``max_p > 10000`` MW (plausibleActivePowerLimit), ``target_p`` lies outside
+  ``[min_target_p, max_target_p]``, that range is degenerate, or ``droop == 0``; a
+  *charging* battery (``target_p < 0``) does participate;
+* its key is ``max_p / droop`` (``max_p`` itself, not ``max_target_p``).
+
+Measured against OpenLoadFlow on a real RTE snapshot, battery by battery (shares x12.4,
```
BDonnot: remove RTE please

lightsim2grid/network/from_pypowsybl/_aux_battery_apc.py
```diff
@@ -0,0 +1,176 @@
...
+x1, x0.144 and 0 for the ``droop = 0`` / ``participate = false`` ones).
+
+pypowsybl up to 1.16.1 does not expose that extension on a battery
+(``get_extensions("activePowerControl")`` only lists generators; fixed by pypowsybl PR
+#1276, not released yet), so :func:`battery_active_power_control` can fall back to reading
+it off an XIIDM export of the network.
+"""
+
+import io
+import xml.etree.ElementTree as ET
```
BDonnot: xml becomes an hard dependency then ? From memory it's rather heavy :-/ Is there any solution outside of it ? Or maybe it's already a dependency of pypowsybl (it would surprise me)

Maybe lazy incluse it if needed only. With a try / catch failing back to a reasonable default behavior if not installed.

lightsim2grid/network/compare_lsgrid.py
```diff
@@ -67,6 +67,8 @@
 ATTR_STORAGES_INPUT = [
     "target_p_mw",
     "target_q_mvar",
+    "is_slack",
+    "slack_weight",
 ]
```
BDonnot, line 72: I put it here because I don't know where to have it otherwise:

- in the previous merge, there was also capability to detect elements of the slack outside their pmin / pmax. Is this physical detection present for battery too ?
- another question, still about the previous (already merged) feature, are the pmin / pmax of generator read when the grid is init from Matpower/ powermodels and pypowsybl ?

lightsim2grid/network/from_pypowsybl/_aux_battery_apc.py
```diff
@@ -0,0 +1,176 @@
...
+from packaging import version
+
+from ._aux_common import PYPOWSYBL_VER
+# OpenLoadFlow's `POWER_EPSILON_SI` and default `plausibleActivePowerLimit`, the ones the
+# bake uses for the generators (_olf_bake imports this module lazily, for that reason)
+from ._olf_bake import _ZERO_P_TOL, _MAX_PLAUSIBLE_ACTIVE_POWER_MW
+
+# OpenLoadFlow's hardcoded fallback droop (`AbstractLfGenerator.DEFAULT_DROOP`, "why not"),
+# for a unit whose `activePowerControl` extension does not set its own
+_OLF_DEFAULT_DROOP = 4.0
```
BDonnot: Avoid redeclaring this type of constants. The day it changes it will be error prone (and hard to debug). Have them all in an util file that is imported when such constants are needed.

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -53,10 +53,28 @@
...
   comparisons are False.
+* For generators, every one of the voltage-control rules above and below is
+  subordinate to what the reference solve shows OLF actually did: a generator
+  whose regulated bus sits exactly at its ``target_v`` (to 1e-8 pu, see
+  ``_TARGET_V_HELD_TOL_PU``; OLF pins a PV magnitude to machine precision) was
+  voltage-controlled and is never frozen, whatever the rules say; a regulating
+  generator whose target was NOT held was not controlling (switched, not
+  started, discarded, merged, below OLF's ``generatorVoltageControlMinNominalVoltage``,
+  ...) and is frozen to its realized Q in a final catch-all pass
+  (:func:`_bake_generator_voltage_control_not_held`). On a real 7k-bus RTE
```
BDonnot: Remove rte and use a more generic term "real grid snapshots" for example

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -178,6 +204,37 @@ def _bound_at_qlimit(df: pd.DataFrame, regulating: pd.Series):
...
 
 
+def _hit_qlimit(df: pd.DataFrame, q_gen: pd.Series):
+    """The reactive limit (generator convention, same limits as
+    :func:`_bound_at_qlimit`) each row's realized output ``q_gen`` sits at: the
+    nearer of the two, a missing limit counting as infinitely far.
+
+    A unit OLF switched at its limit injects exactly that limit, but the ``q`` it
+    reports is not always it: OLF spreads the reactive target of a bus over the
+    generators of that bus again when writing the results (on an RTE snapshot, two
```
BDonnot: Remove rte name

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -178,6 +204,37 @@ def _bound_at_qlimit(df: pd.DataFrame, regulating: pd.Series):
...
+    units of one bus at 1.22366 and 1.22242 MVAr reported 1.22304 and 1.22242, the
```
BDonnot: Don't put such precise numbers, explain the behavior in the doc.

It's not a memory...

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -178,6 +204,37 @@ def _bound_at_qlimit(df: pd.DataFrame, regulating: pd.Series):
...
+    return pd.Series(np.where(to_min < to_max, qmin, qmax), index=df.index)
+
+
+def _baked_q_at_limit(df: pd.DataFrame, q_gen: pd.Series):
+    """The fixed reactive output to bake for a unit frozen at its Q limit: the limit
+    (see :func:`_hit_qlimit`) when the reported ``q`` is within ``_Q_LIMIT_TOL_ABS`` of
+    it, the reported ``q`` otherwise.
+
+    Only a report that close is the misreport of a limit injection (2.9e-4 and
+    6.2e-4 MVAr on RTE snapshots, where the bus balance is off by exactly that much).
```
BDonnot: Remove rte and the values

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -178,6 +204,37 @@ def _bound_at_qlimit(df: pd.DataFrame, regulating: pd.Series):
...
+    The relative tolerance of :func:`_bound_at_qlimit` also catches units a visible
+    distance from their limit -- settled 0.18 MVAr inside it, or injecting a
+    ``target_q`` 0.11 MVAr beyond it -- and those inject what they report (their bus
```
BDonnot: Same here... Don't put number, they are good for memory but not for a documentation

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -336,8 +411,84 @@ def _bake_taps_and_sections(network, keep_only_main_comp=True):
 # parameter (AbstractLfGenerator.checkActivePowerControl).
 _MAX_PLAUSIBLE_ACTIVE_POWER_MW = 10000.0
 
+# Tolerance (pu of the regulated bus nominal voltage) under which the reference
+# solve is deemed to have HELD a generator's target voltage. OLF eliminates the
+# voltage-magnitude unknown of a PV bus (the magnitude *is* the target), so a
+# generator that actually took part in voltage control leaves |V - target_v| at
+# machine precision (~1e-16 pu measured on a 7k-bus RTE snapshot), whatever the
```
BDonnot: Rte name !!!

And update your memory with this (no explicit mention of rte name, of rte assets) in the documentation.

And because we are human and not robot, no mention of exact numbers like difference between a reference and sometimes else (the 1.34759 or something you wrote earlier) It's not useful.

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -521,45 +690,253 @@ def _bake_remote_control_bus_conflicts(network, keep_only_main_comp=True):
...
+    upd["target_q"] = -gen["q"].to_numpy()[mask]
+    upd["voltage_regulator_on"] = False
+    network.update_generators(upd)
+
+
+# Tolerance (MVAr) under which a non-regulating generator's realized reactive output is
+# deemed to be its target_q: a PQ injection is reproduced to the digit, whatever the
+# Newton-Raphson tolerance, and the clamped values seen on RTE snapshots go down to
+# ~3e-3 MVAr.
+_TARGET_Q_TOL_MVAR = 1e-6
```
BDonnot: Constants at the top of the file please

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -521,45 +690,253 @@ def _bake_remote_control_bus_conflicts(network, keep_only_main_comp=True):
...
+
+
+def _bake_generator_target_q_forced_in_limits(network, keep_only_main_comp=True):
+    """Write the realized reactive output into ``target_q`` for every connected,
+    non-regulating generator whose reference solve did not inject its ``target_q``.
+
+    OLF's default ``forceTargetQInReactiveLimits`` clamps a PQ generator's target into
+    its reactive capability at its active power (on RTE snapshots: units at 0 MW whose
```
BDonnot: Rte again

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -521,45 +690,253 @@ def _bake_remote_control_bus_conflicts(network, keep_only_main_comp=True):
...
+    curve starts above 0 MVAr, with ``target_q = 0``). The loop-free parameters turn the
+    reactive limits off, which turns the clamp off with them, and lightsim2grid never
+    clamps: without this both inject the raw target (7.6e-4 pu on the 6 kV side on one
```
BDonnot: Typical things that should not be here

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -521,45 +690,253 @@ def _bake_remote_control_bus_conflicts(network, keep_only_main_comp=True):
...
+    q_gen = -gen["q"]  # result column is load convention; target_q is generator convention
+    moved = (~gen["voltage_regulator_on"].astype(bool) & q_gen.notna()
+             & ((q_gen - gen["target_q"]).abs() > _TARGET_Q_TOL_MVAR))
+    if not moved.any():
+        return
+    network.update_generators(pd.DataFrame({"target_q": q_gen[moved].to_numpy()}, index=gen.index[moved]))
+
+
+def _bake_battery_voltage_control(network, keep_only_main_comp=True):
+    """Same result-driven decision as for generators, for the batteries carrying
```
BDonnot: Can both function be merged together then ?

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -521,45 +690,253 @@ def _bake_remote_control_bus_conflicts(network, keep_only_main_comp=True):
...
+
+
+def _extrapolate_curve_limits(network, df: pd.DataFrame):
+    """Return ``df`` with ``min_q_at_p`` / ``max_q_at_p`` replaced, for every row whose
+    realized active power (``-p``) lies outside the P range of its reactive capability
+    curve, by the linear extrapolation of the curve's end segment on that side.
+
+    That is what OLF evaluates with its default ``extrapolateReactiveLimits``, where
+    pypowsybl's columns clamp to the end point. On RTE snapshots, units running below
+    their curve (85.8 MW on a 217-947 MW curve, 2.3 MW on an 8.35-30 MW one) were
```
BDonnot: No no no !

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -726,7 +1104,20 @@ def _bake_active_power_control_participation(network, gen):
...
+      sharing key, caps every generator that would cross ``max_target_p`` (for a
+      positive mismatch; ``min_target_p`` for a negative one) at that limit and
+      redistributes the rest over the others, so a generator whose realized
+      dispatch sits at the limit on the side the mismatch pushes it took no share
+      at all. That capping is a dynamic effect no static weight reproduces; the
+      sign of the reference mismatch -- the total realized minus dispatched
+      generation -- is known here, so the generators OLF capped are excluded, and
+      the static distribution lightsim2grid then derives from the flag is the one
+      OLF effectively used (on a real 7k-bus RTE snapshot the 41 units at their
+      ``max_p`` carried 42 % of the raw key). This only holds for a mismatch of
```
BDonnot: Mention of rte

And 42% should not happen here

lightsim2grid/network/from_pypowsybl/_olf_bake.py
```diff
@@ -53,10 +53,28 @@
   0 MW) is left alone.
 * PV -> PQ reactive-limit switches (generators, VSC converter stations): only
   the units whose realized reactive power actually sits at a Q limit are frozen
-  to fixed-Q at that value, with voltage regulation switched off. Units still
```
BDonnot, line 56: Putting it at the top of the changes here, but does not really concern it:


Add in the Todo, on the Changelog, the fact that :
1) this file should be split into comprehensible "sub files" to facilitate the debugging 
2) there should be, opt-in options, off by default, the possibility to get the changes made to the grid. Like "gen xxx removed from slack", "batt yyy assigned PQ" etc in a json format for example

lightsim2grid/tests/test_olf_bake.py
```diff
@@ -424,6 +437,186 @@ def test_olf_baked_reporter_shows_no_action(self):
...
+        import pandas as pd
+        df = pd.DataFrame({"min_q": [-10.0, -10.0, np.nan, -4.0], "max_q": [5.0, 5.0, 3.0, np.nan]},
+                          index=["at_max", "at_min", "no_min", "no_max"])
+        q_gen = pd.Series([4.9995, -9.998, 2.999, -3.99], index=df.index)
+        hit = _hit_qlimit(df, q_gen)
+        self.assertEqual(hit.to_dict(), {"at_max": 5.0, "at_min": -10.0, "no_min": 3.0, "no_max": -4.0})
+
+    def test_olf_switched_unit_baked_at_its_limit_not_reported_q(self):
+        """A unit switched at its Q limit injects that limit, but OLF does not always
+        report it (on RTE snapshots it re-splits a bus' reactive target among the units
```
BDonnot: Rte...

src/core/VoltageControlData.hpp
```diff
@@ -47,7 +47,10 @@ struct LS2G_API VoltageControlSolverData
     // `elem_id` carries the hvdc LINE id for both, the kind saying which end. A
     // station is a controller only when a group regulates its bus -- on its own it
     // pins that bus through the ordinary PV path, like a local generator.
-    enum Kind { GEN = 0, SVC = 1, HVDC_SIDE_1 = 2, HVDC_SIDE_2 = 3 };
+    // STORAGE is never a controller of the plan (a storage unit only ever pins its
+    // own bus, see StorageContainer); the tag exists for LSGrid's reactive-residual
+    // bookkeeping (QShare), which uses the same kind space.
+    enum Kind { GEN = 0, SVC = 1, HVDC_SIDE_1 = 2, HVDC_SIDE_2 = 3, STORAGE = 4 };
```
BDonnot, line 57: Are storage units properly added to the physical violation qmin / qmax check ?

src/core/element_container/SlackParticipation.hpp
```diff
@@ -0,0 +1,223 @@
+// Copyright (c) 2026, RTE (https://www.rte-france.com)
+// See AUTHORS.txt
+// This Source Code Form is subject to the terms of the Mozilla Public License, version 2.0.
+// If a copy of the Mozilla Public License, version 2.0 was not distributed with this file,
+// you can obtain one at http://mozilla.org/MPL/2.0/.
+// SPDX-License-Identifier: MPL-2.0
+// This file is part of LightSim2grid, LightSim2grid implements a c++ backend targeting the Grid2Op platform.
+
+#ifndef LS2G_SLACK_PARTICIPATION_H
+#define LS2G_SLACK_PARTICIPATION_H
```
BDonnot, line 10: I put this comment here, why not...

Add a Todo in the changelog for the NR behavior when multiple slacks are present.
We should now make the difference between "participate in the distributed slack" which is pure active things and has nothing to do with voltages and "is reference slack" which is a much stronger say : theta = known, VM = known, p and q unknown.
Today an element participating to the distributed slack is necessarily PV which does not make any sense (a load could participate in the distributed slack...)

Let's not adress this yet, but have a todo around this.

BDonnot (commented):
Some changes

lightsim2grid/network/from_pypowsybl/_aux_battery_apc.py
BDonnot: Not a new dependency: `xml.etree.ElementTree` is in the Python standard library, so it is already there with the interpreter, and neither pypowsybl nor anything else has to ship it. Its C accelerator (`_elementtree`) is built into CPython.

That said, the concern about paying for it on every import is fair, so it is now imported **inside** `_apc_from_xiidm` — the only function that needs it, and only on the fallback path (pypowsybl <= 1.16.1, `source="auto"`). The import is wrapped in a `try` / `except ImportError` that warns and returns `{}`, which makes every battery fall back to OpenLoadFlow's defaults, so a Python built without `_elementtree` still loads the grid.

Done in c16eb3f (branch `claude/pr-207-comments-defow7`).

---

lightsim2grid/network/from_pypowsybl/_aux_battery_apc.py
BDonnot: Agreed — and worse than a duplicate, `_aux_battery_apc.py` was importing two of them back from `_olf_bake.py`, which is what forced the lazy-import dance between the two modules.

They now all live in one new module, `lightsim2grid/network/from_pypowsybl/_olf_const.py`, in two groups: OpenLoadFlow's own constants mirrored here (`_ZERO_P_TOL`, `_MIN_REACTIVE_RANGE_MVAR`, `_MIN_PLAUSIBLE_TARGET_V_PU` / `_MAX_PLAUSIBLE_TARGET_V_PU`, `_MIN_NOMINAL_V_FOR_TARGET_V_CHECK_KV`, `_MAX_PLAUSIBLE_ACTIVE_POWER_MW`, `_OLF_DEFAULT_DROOP`) and the tolerances used to read a reference solve back (`_Q_LIMIT_TOL_ABS` / `_Q_LIMIT_TOL_REL`, `_TARGET_V_HELD_TOL_PU`, `_TARGET_Q_TOL_MVAR`). `_olf_bake.py` has no module-level constant left at all, `_aux_battery_apc.py` and `_aux_add_slack.py` import from `_olf_const` directly, and the circular import is gone with them.

Done in c16eb3f (branch `claude/pr-207-comments-defow7`).

---

lightsim2grid/network/compare_lsgrid.py
BDonnot: Two questions, two different answers.

**1. Is the pmin / pmax detection present for batteries too? No, and it cannot be yet.**

`GenPCheck` (the active-power third of `compute_physical_violations`) only walks the generators, and the reason is one level down: `StorageContainer` carries no active power limits at all — there is no `min_p_mw` / `max_p_mw` on a storage unit, no `set_storage_p_limits`, nothing in `StateRes`. So a battery the distributed slack pushed past what it can deliver goes unreported, even though this PR does let it take a share of the slack.

Closing that gap is a real piece of work (container state + `BINARY_FORMAT_VERSION` bump + bindings + the converters + `GenPCheck`), so I have not widened this PR with it — it is written up as a TODO in the changelog instead, with what it takes. Say the word if you would rather have it here.

The *reactive* side is fine: a voltage-regulating storage unit's `[min_q, max_q]` does count towards its bus' capability (`BusQCheck`), tested both in `test_physical_violations.py::TestStorageCapabilityFromPython` and in `test_batch_physical_violations.cpp`.

**2. Are the generators' pmin / pmax read by the other converters? pypowsybl yes, MATPOWER / PowerModels no — now fixed.**

* pandapower: yes, `_aux_add_gen.py` reads `min_p_mw` / `max_p_mw`.
* pypowsybl: yes, `_aux_add_generators.py` reads `min_p` / `max_p` (iidm always carries them).
* PowerModels / MATPOWER: **no**. `from_powermodels/_aux_add_gen.py` read `qmin` / `qmax` and stopped there, so a grid loaded from a `.m` file could not report an active-power violation at all — even though `_mpc_to_powermodels.py` was already carrying `PMIN` / `PMAX` through into the dict, unused.

  Fixed in c16eb3f: the PowerModels converter now calls `set_gen_p_limits`, which covers MATPOWER since it goes through the same dict. A missing or non-finite limit becomes NaN, i.e. "no limit for that one", same convention as the other two. Tested in `test_init_from_matpower.py::test_generator_active_power_limits_are_read`.

---

src/core/VoltageControlData.hpp
BDonnot: Yes — a voltage-regulating storage unit's reactive range does count towards its bus' capability.

The `STORAGE` tag not being a controller of the voltage-control *plan* is what made this look doubtful, and the two are independent: `BusQCheck::build_bus_q_plan` builds each bus' `[q_min, q_max]` from everything that holds that bus, whatever path it holds it through, and it walks the storage units explicitly (`is_local_voltage_controller`, `get_min_q` / `get_max_q` summed into the bus like a local generator's). Tested in `test_physical_violations.py::TestStorageCapabilityFromPython::test_limits_of_a_regulating_storage_unit` and in `test_batch_physical_violations.cpp` ("a voltage-regulating storage unit's reactive capability counts towards its bus'").

Since the question came up right here, the comment on `Kind` now says it rather than leaving it to be re-derived (c16eb3f).

The *active* side is the one that is genuinely missing — see [my answer on the `compare_lsgrid.py` thread](https://github.com/Grid2op/lightsim2grid/pull/207#discussion_r4052302978): a storage unit has no `min_p_mw` / `max_p_mw` at all, so `GenPCheck` has nothing to compare against. TODO added.

---

lightsim2grid/network/from_pypowsybl/_olf_bake.py
BDonnot: Partly — the decision merges, the plumbing does not.

What was genuinely duplicated is now shared, as two helpers:

* `_resolve_regulated_bus(network, own_bus, regulated_element_id)` — which bus an element regulates, falling back to its own bus when it is local or the remote element cannot be resolved. The battery function had its own inline copy of this, right down to the `RuntimeError` fallback.
* `_target_v_held(network, reg_bus, target_v)` — did the reference solve hold that bus at the target (`_TARGET_V_HELD_TOL_PU`). This is *the* decision, and it is about the regulated bus, not about what is holding it, so there was no reason for two copies. `_generator_target_v_held` is now a one-liner over it.

That took ~30 lines out of `_bake_battery_voltage_control`.

What I did not merge into one function is the read/write around it, because a battery's voltage control is not in the same place as a generator's: the flag, the target voltage and the regulated element live in the `voltageRegulation` **extension** frame rather than in columns of `get_batteries()`, the reactive output comes from `get_batteries()`, and switching the regulation off is `update_extensions` + `update_batteries` rather than a single `update_generators`. Folding those into one function means a `kind` argument and a branch on it at every access — more code than the two thin wrappers, and harder to read. The docstring now says this explicitly so the next reader does not ask the same question.

Done in c16eb3f. `test_olf_bake.py` (27 tests) and `test_storage_voltage_control.py` (8) pass.

---

lightsim2grid/network/from_pypowsybl/_olf_bake.py
BDonnot: Both removed here, and everywhere else in the diff — see the [summary comment](https://github.com/Grid2op/lightsim2grid/pull/207#issuecomment-placeholder) for the full list.

On "update your memory with this": the durable place for it in this repo is `CLAUDE.md`, so it is now a section there, *Comments and docstrings: no RTE, no measured numbers* —

* no mention of RTE, of an RTE grid, snapshot or asset; say "a real grid snapshot" / "real grid snapshots" / "a large real grid". The copyright header is the one place the name belongs.
* no exact numbers carried over from an investigation ("30 generators", "42 % of the raw key", "18.820306 MVAr"). They date, a reader cannot check them, and they describe one snapshot rather than the behaviour — describe the behaviour instead.
* the exception, spelled out so it is not over-applied: a number a reader can act on stays — a tolerance the code actually uses, a constant mirrored from OpenLoadFlow, the values a test fixture sets. The rule is about measurements, not about arithmetic.

Done in c16eb3f.

---

lightsim2grid/network/from_pypowsybl/_olf_bake.py
BDonnot: Both added at the top of the `[TODO]` section of `CHANGELOG.rst` (c16eb3f):

1. **Split `_olf_bake.py`.** Along what each rule bakes — taps and sections, generator voltage control, reactive-limit switches, SVCs, active power and slack participation — so a failing bake points at one readable module rather than at a thousand-line file.
2. **An opt-in report of what the bake changed**, off by default: "gen xxx removed from the slack participants", "batt yyy frozen to PQ", "tap zzz fixed at position n", in json, each with the rule that decided it. Today the bake rewrites the grid silently and a caller cannot tell which decision produced the state it is handed.

Two more went in while going through the review, from the other threads: separating slack *participation* from the *reference* slack ([here](https://github.com/Grid2op/lightsim2grid/pull/207#discussion_r4052405425)), and active-power limits for storage units ([here](https://github.com/Grid2op/lightsim2grid/pull/207#discussion_r4052302978)).

---

src/core/element_container/SlackParticipation.hpp
BDonnot: Agreed, and not addressed — TODO added, in two places (c16eb3f):

* `CHANGELOG.rst`, top of `[TODO]`: "participates in the DISTRIBUTED SLACK" is a purely active statement — this element takes a share of the power imbalance, by a participation factor — and says nothing about voltage, unlike "is the REFERENCE slack" (theta known, |V| known, P and Q unknown). Today a participant is necessarily PV, which does not follow: a load could take a share without controlling any voltage. Separating the two roles is what would allow it.
* a shorter note in the class docstring of `SlackParticipation` itself, pointing at the changelog — that is where someone adding a participant will be looking.

---

BDonnot:
## Review worked through — on branch `claude/pr-207-comments-defow7`

All 20 threads are addressed in **c16eb3f**, pushed to `claude/pr-207-comments-defow7` (branched off this PR's head, `eccddec`), so it merges into `fix_derivatives` cleanly. I did not push to your branch.

| | added | removed |
|---|---|---|
| **C++ (src/core)** | +11 | -1 |
| **Python (package)** | +150 | -122 |
| **Tests** | +10 | -3 |
| **Docs + changelog** | +48 | -1 |
| total | +296 | -162 |

### RTE names and measured numbers — 11 threads

Every `RTE` introduced by this PR is gone, replaced by "a real grid snapshot" / "real grid snapshots", and every number carried over from an investigation with it. What each rule *does* is said instead:

| where | was | now |
|---|---|---|
| `_aux_battery_apc.py:25` | "a real RTE snapshot ... shares x12.4, x1, x0.144" | which units the rule keeps, and that the `droop = 0` ones take no part |
| `_aux_add_slack.py:58` | "7k-bus RTE snapshot ... to 1e-13" | "real grid snapshots ... to solver precision" |
| `_olf_bake.py:71` | "30 generators ... and ~100 they left regulating" | what the catch-all decides, in both directions |
| `_olf_bake.py:214-215` | "two units at 1.22366 and 1.22242 reported 1.22304 and 1.22242" | two units of a bus reported a hair from what they injected, the bus no longer balancing |
| `_olf_bake.py:229-232` | "2.9e-4 and 6.2e-4 MVAr ... 0.18 MVAr inside it ... 0.11 MVAr beyond" | the bus balance being off by the discrepancy is what identifies a misreport |
| `_olf_bake.py:418` | "~1e-16 pu ... at least ~1e-7 pu away" | machine precision vs. orders of magnitude further; the threshold sits between them |
| `_olf_bake.py:734-737` | "RTE snapshots ... 7.6e-4 pu on the 6 kV side" | the case it covers, and that the low-voltage side is visibly off without it |
| `_olf_bake.py:825-826` | "85.8 MW on a 217-947 MW curve ... 18.820306 / -13.686818 MVAr" | the extrapolated limit sits far enough from the clamped one to bake the wrong value |
| `_olf_bake.py:909` | "2e-3 pu on an RTE snapshot" | removed |
| `_olf_bake.py:1117-1118` | "the 41 units at their `max_p` carried 42 %" | the capped units carry enough of the raw key that ignoring them skews the distribution |
| `test_olf_bake.py:499` | "on RTE snapshots" | "on real grid snapshots" |

The numbers that stay are the ones a reader can act on: the tolerances the code uses, the constants mirrored from OpenLoadFlow, the values the test fixtures set.

The rule itself is now written down in `CLAUDE.md` (*Comments and docstrings: no RTE, no measured numbers*), so it does not have to be re-discovered next time.

### The rest

* **Constants** — all of them, OLF-mirrored and reading-back tolerances alike, moved into one module `from_pypowsybl/_olf_const.py`. `_olf_bake.py` has no module-level constant left (that also answers "constants at the top of the file"), and the circular import between `_olf_bake` and `_aux_battery_apc` goes with them.
* **`xml`** — stdlib, not a new dependency, but now imported inside `_apc_from_xiidm` only, behind a `try` / `except ImportError` falling back to OpenLoadFlow's defaults.
* **Merging the two voltage-control functions** — the decision merged (`_resolve_regulated_bus`, `_target_v_held`), the read/write around it did not, because a battery's regulation lives in an extension frame. Reasons in the docstring and [on the thread](https://github.com/Grid2op/lightsim2grid/pull/207#discussion_r4052349729).
* **Storage units and the physical checks** — reactive: already covered (`BusQCheck` counts a regulating storage unit's range in its bus'). Active: **not** covered, and cannot be — a storage unit has no `min_p_mw` / `max_p_mw` at all. TODO added rather than widening this PR.
* **Generator `pmin` / `pmax` by converter** — pandapower yes, pypowsybl yes, PowerModels / MATPOWER **no**. Fixed: the PowerModels converter now calls `set_gen_p_limits`, which covers MATPOWER too, with a test.
* **TODOs** — four in `CHANGELOG.rst`: separating distributed-slack participation from the reference slack (+ a note in `SlackParticipation.hpp`), splitting `_olf_bake.py`, the opt-in json report of what the bake changed, and storage active-power limits.

### Tests

Built from source and run locally: `test_olf_bake` (27), `test_storage_voltage_control` (8), `test_storage_distributed_slack` (12), `test_init_from_matpower` (27), `test_voltage_control_pypowsybl` (20), `test_physical_violations` (26) — all pass, except 4 in `test_physical_violations` that error on `import grid2op`, which is not installed in this container and is unrelated to the change.

---

lightsim2grid/network/from_pypowsybl/_olf_bake.py
BDonnot: (the summary link above should be https://github.com/Grid2op/lightsim2grid/pull/207#issuecomment-5739901840 — the comment did not exist yet when I wrote that reply)

---

BDonnot (commented):
Ok

PR #209, opened 2026-09-19
dist slack: read the first guess off the residual, and keep reading it in the FDPF

BDonnot:
| | added | removed |
|---|---|---|
| **C++ (src/core, src/bindings)** | +214 | -2 |
| **Tests** | +203 | -0 |
| **Docs + changelog** | +8 | -0 |
| total | +425 | -2 |

Both AC families seeded the distributed slack with `Re(sum(Sbus))` and neither of them
should have. The Newton-Raphson paid an iteration for it; the fast-decoupled algorithms
never moved off the seed at all and converged to the wrong distribution.

## The identity both fixes rest on

The sum of the per-bus active mismatches IS the grid's global active balance

```
sum_b Re(mis(b)) = sum_b Re(Scomp(b) - Sbus(b) - ...) + sa . W
```

with `W` the total participation weight. It is **affine in `sa` and in nothing else**,
whatever the rest of the right-hand side is made of, so one sum over the P rows gives the
`sa` that closes the balance at the starting voltages -- exactly, not approximately. And
because `sa` enters `mis` only through `+ sa . slack_weights`, shifting it by `delta`
shifts each slack bus' P row by `-delta . weight` and leaves every other row alone: the
residual is **corrected in place**, never re-evaluated. No second `Ybus . V` in either
family.

`Re(sum(Sbus))` is the answer to that same equation at a flat start of a lossless grid
whose entire right-hand side is Sbus, and nowhere else. That is the whole bug, twice.

## 1. Newton-Raphson: an iteration spent on the seed

`NRAlgo::compute_pf` now calls `NRSystem::calibrate_slack_absorbed` on the residual it has
just evaluated, before the first convergence test. Two things follow:

* a voltage that already meets the KCL -- another powerflow's output, the previous row of
  a time series -- is converged, where it used to cost one iteration for the slack alone;
* anything on the right-hand side that does **not** go through Sbus is accounted for. The
  angle-droop hvdc flows are the case in the tree today (`HvdcLineContainer` deliberately
  does not stamp them in AC); the exotic elements added since are the reason to stop
  guessing this way rather than to guess better.

Measured, every generator participating, flat start then re-solved from its own output
(SparseLU, tol 1e-8):

| | cold | warm (own V) | ‖dV‖ warm |
|---|---|---|---|
| case14 | 4 -> 4 | 1 -> **0** | 6.9e-16 -> 2.2e-16 |
| case118 | 4 -> 4 | 1 -> **0** | 1.7e-14 -> 2.3e-16 |
| case300 | 5 -> 5 | 1 -> **0** | 4.8e-13 -> 2.3e-16 |
| case1354pegase | 5 -> 5 | 1 -> **0** | 1.4e-13 -> 2.3e-16 |
| case9241pegase | 6 -> 6 | 1 -> **0** | 6.1e-13 -> 2.5e-16 |

A/B'd with two full builds. Cold starts are unchanged -- at `V = 1` the two guesses differ
only by the shunt and charging power drawn there -- and a warm re-solve now returns its
seed rather than one Newton step away from it.

## 2. Fast-decoupled: the seed WAS the answer

Worse than a bad guess. `slack_absorbed_` was set once and never written again; the
reference the signature already carried -- `real_type & slack_absorbed`, passed down into
`has_converged` -- was a reference to something no line assigned.

So every participant produced `Sbus(b) - w_b . sum(Sbus)` instead of
`Sbus(b) - w_b . (sum(Sbus) - losses)`, and the difference landed on the reference bus,
the one bus the fast-decoupled split writes no P equation for -- which is why nothing ever
complained. On a case14 with two slacks (0.75 / 0.25):

```
NR         sa = -0.036000   gen_p = [40, 2.70, 0, 0, 233.2933]
FDPF_XB    sa =  0.133933   gen_p = [40, 6.39, 0, 0, 229.0450]   <- 4.25 MW out, dV 6.2e-3 pu
pandapower (distributed_slack=True) = [40, 2.68, 0, 0, 233.2879]
```

B' and B'' cannot take the unknown -- they are constant matrices, which is the point of the
family, and the row it needs is the derivative of the losses, which is not. It is an outer
loop instead: `BaseFDPFAlgo::calibrate_slack_absorbed` re-solves the scalar balance at
every mismatch this solve evaluates. At the fixed point every P row of pvpq is under
tolerance and the sum over all buses is zero, so the reference slack's own mismatch --
minus the sum of the others -- is under tolerance too, up to their number.

Against Newton-Raphson, flat start, tol 1e-9, three slack buses (XB / BX):

| | max \|dV\| | max \|d gen_p\| | iters |
|---|---|---|---|
| case14 | 2.6e-13 / 5.9e-13 | 1.4e-10 / 8.2e-11 MW | 10 / 14 |
| case30 | 4.0e-13 / 6.5e-13 | 1.5e-11 / 7.8e-11 MW | 16 / 10 |
| case118 | 4.4e-13 / 1.2e-13 | 2.2e-10 / 1.6e-10 MW | 15 / 12 |
| case300 | 6.4e-13 / 1.2e-12 | 3.0e-10 / 6.1e-10 MW | 30 / 29 |
| case1354pegase | 1.3e-13 / 4.3e-14 | 7.6e-11 / 7.0e-11 MW | 15 / 19 |
| case9241pegase | 5.0e-13 / 4.7e-13 | 8.0e-10 / 5.8e-10 MW | 30 / 24 |

Iteration counts are within one of the single-slack runs, so the outer loop rides along
with the inner one rather than adding a stage.

**Skipped entirely when the grid has one slack bus** (`dist_slack_`): `sa` then shifts
`mis_bus_` at the reference bus alone, which owns no equation, and whose published
injection is read back as `mis.real() - sa . weight` whatever `sa` is. That path is
bit-for-bit what it was, and does not pay the reduction.

## Consistency of what is published

`LSGrid::_fill_bus_mismatch_ac` reads `mis.real() - slack_absorbed * slack_weights` back
off the algorithm to say what the generators produced, so the two must move together. Both
fixes shift the per-bus mismatch buffer with the state -- which matters in the NR precisely
because a solve that converges at iteration 0 never evaluates another residual.

## Masking and degenerate weights

A masked bus contributes 0 to the sum (its row is already pinned), which is what an island
a contingency stranded should contribute; a participation with no weight left to distribute
to leaves the seed alone. The batch sweeps zero and renormalise a masked bus' weight
themselves (`BaseBatchSweep::_masked_slack_weights`), and the NR restores the pinned rows
after the correction anyway.

## Tests

C++ **349/349**, four new cases:

* `test_powerflow_algorithm.cpp` -- the calibration recovers the converged `slack_absorbed`
  off a converged voltage and zeroes the residual; a dist-slack solve started from its own
  solution takes no iteration (reports `1 == 0` without the fix); the same with an
  angle-droop hvdc line, which is the "not in Sbus" case; a no-op on the single-slack
  system.
* `test_fdpf_algorithm.cpp` -- FDPF distributes the slack the way Newton-Raphson does
  (fails without the fix, by 5e-6 to 1e-5 pu per bus), and a single slack leaves the FDPF
  untouched.

Python, run individually: `fdpf`, `multiple_slack`, `multi_slack`, `dist_slack_jacobian`,
`dc_dist_slack`, `dist_slack_backend`, `NewtonPF`, `ACPF`, `SparseLUSolver`, `TimeSerie`,
`InjectionSweep`, `ScenarioSweep`, `ScenarioSweep_gen_contingency`, `SecurityAnlysis_cpp`,
`ContingencyAnalysis_split`, `cache_reuse`, `case118`, `solver_control` (204). All pass.

## Unrelated, pre-existing, not investigated here

On `case2848rte` both FDPF flavours land somewhere other than the Newton-Raphson (74 and
60 iterations; magnitudes agree to ~3e-3, the complex voltages do not). It happens with
**one** slack bus as well as with three -- i.e. on the path this PR does not touch.

https://claude.ai/code/session_01PfeWNrs4TYY57WzLyCdxza

---
_Generated by [Claude Code](https://claude.ai/code/session_01PfeWNrs4TYY57WzLyCdxza)_


src/core/powerflow_algorithm/BaseFDPFAlgo.hpp
```diff
@@ -142,6 +144,46 @@ class BaseFDPFAlgo final: public BaseAlgo
...
+         * minus the sum of the others -- is under tolerance too, up to their number.
+         */
+        void calibrate_slack_absorbed(real_type & slack_absorbed,
+                                      const Eigen::Ref<const RealVect> & slack_weights)
+        {
+            const real_type total_w = slack_weights.sum();
+            // nothing to distribute to (no weight left after a re-weighting, a
+            // degenerate input): leave the state as it is. Same 1e-12 floor as the
+            // batch sweeps' own re-weighting.
+            if(std::abs(total_w) <= static_cast<real_type>(1e-12)) return;
```
BDonnot: There is a constant "tol_eq_flaot" or something in this repo. Why not use it ?

src/core/powerflow_algorithm/NRSystem.hpp
```diff
@@ -463,6 +463,54 @@ class LS2G_API MultiSlack   // distributed-slack extension
...
+        real_type absorb_balance(real_type sum_p, Eigen::Ref<RealVect> res)
+        {
+            if (my_size_ == 0 || slack_weights_.size() == 0) return static_cast<real_type>(0.);
+            real_type total_w = static_cast<real_type>(0.);
+            for (int k = 0; k < my_size_; ++k) total_w += slack_weights_(slack_buses_[k]);
+            // a degenerate participation (every weight zero, or a sweep row that
+            // masked the last participant out) leaves the guess as it was: there
+            // is no bus to distribute to. Same 1e-12 floor as the batch sweeps'
+            // own re-weighting (BaseBatchSweep::_masked_slack_weights).
+            if (std::abs(total_w) <= static_cast<real_type>(1e-12)) return static_cast<real_type>(0.);
```
BDonnot: Use tol_ea_float instead of re defining such constants

src/core/powerflow_algorithm/NRSystem.tpp
```diff
@@ -460,6 +460,49 @@ inline RealVect NRSystem<Base, Rest...>::mismatch() const
...
+    // MultiSlack the slack ones), and a row index is handed out once, so this
+    // sums each bus' active mismatch exactly once. A masked bus contributes 0:
+    // _residual_into already forced its row to zero, which is what we want --
+    // an island the contingency stranded is not part of the balance.
+    real_type sum_p = static_cast<real_type>(0.);
+    const std::vector<int>& p_rows = ledger_.p_rows();
+    for (size_t k = 0; k < p_rows.size(); ++k) sum_p += res(p_rows[k]);
+
+    const real_type delta = ms->absorb_balance(sum_p, res);
+    if (delta == static_cast<real_type>(0.)) return false;
```
BDonnot: Comparison of float with == is had practice

src/core/powerflow_algorithm/BaseFDPFAlgo.hpp
BDonnot: Done in ebf4f83, here and in `NRSystem.hpp` — `_tol_equal_float` in both.

Checked that widening the band from 1e-12 to 1e-7 is safe here rather than assuming it: the weights this guard reads are normalised before they reach either algorithm (`LSGrid::pre_process_solver`, `slack_raw_weights / slack_raw_weights.sum()`), so the sum is 1 on a live grid and exactly 0 on a row whose participants a contingency masked out. Nothing realistic sits in between for the wider band to catch, and refusing to divide by a weight that small is the safer half anyway — it would amplify the imbalance rather than place it.

---

src/core/powerflow_algorithm/NRSystem.hpp
BDonnot: Done in ebf4f83 — same change as the FDPF one, `_tol_equal_float` here too.

The comment that pointed at `BaseBatchSweep::_masked_slack_weights` as the precedent for 1e-12 is gone with it; that one is now the odd one out rather than the reference (see the note on the PR about the other places).

---

src/core/powerflow_algorithm/NRSystem.tpp
BDonnot: Agreed, and removed rather than loosened into a tolerance in ebf4f83.

The `==` was carrying a decision — "the component did nothing, so leave the per-bus mismatch alone" — in the value of a quotient, which is the wrong place for it. `absorb_balance` now returns a `bool` and takes the mismatch buffer as an argument, so the two shifts that must agree (the residual row and `mis`) happen in one loop instead of being split across the caller, and there is no comparison left to make.

That also dropped a pass over the buses: the old correction wrote `mis.array() += delta * slack_weights.array()` across the whole grid to change the handful of entries that are not zero there. It now walks the slack buses — bit-identical, since adding `0.` to a double is exact, and O(nb_slack) instead of O(nb_bus).

---

BDonnot:
Swept the tree for the same two patterns. All three review points are fixed in ebf4f83; the rest is a report, **nothing below is touched by this PR**.

### The 1e-12 weight guard exists elsewhere

Same guard, same meaning, and on the **normalised** weights — so the swap would be as safe there as it is here:

- `src/core/batch_algorithm/BaseBatchSweep.hpp:1340`, `_masked_slack_weights`:
  `if(new_sum > 1e-12 && orig_sum > 1e-12) scratch *= (orig_sum / new_sum);`
  It re-weights `active_layout().slack_weights`, which is already `slack_raw_weights / sum()`.

Same guard, but on **raw** weights, where I would not change it without your call:

- `src/core/batch_algorithm/GenPCheck.hpp:134` — `if(std::abs(weight) < 1e-12) continue;` on `get_gen_slack_weight(gen_id)`
- `GenPCheck.hpp:247` — `if(!(std::abs(total_raw_w) > 1e-12)) return;`
- `GenPCheck.hpp:260` — `if(std::abs(bus_raw_w) > 1e-12){ ... }`, guarding the division

A raw participation weight is whatever the caller set and has no fixed scale, unlike the normalised vector the solver sees. `_tol_equal_float` there would swallow a legitimately small weight rather than a degenerate one, so it is a different judgement from the one made in this PR — worth making deliberately, in its own change.

Not the same thing, left alone: `BaseBatchSweep.hpp:1367`, `std::abs(it.value()) > 1e-8` in the connectivity walk. That asks whether a Ybus coefficient is a real connection — an admittance threshold, not a float-equals-zero test.

### Float equality

The one this PR introduced was the only place comparing a **computed** value. The others are exact sentinels and read as intentional:

- `NRAlgo.tpp:143` — `if (coeff != 1.) F *= coeff;`. The NoScaling policy returns the literal `1.`; a false negative costs one redundant multiply, nothing else.
- `HvdcLineContainer.cpp:241` and `:303` — `r_ohm_(hvdc_id) != 0.`, an input read back ("was a resistance given").
- `HvdcLineContainer.cpp:306` — `r_prime == 0.`. Computed, but it is a product of finite factors, so it is exactly zero only when `r_ohm_` is.

Happy to do the `BaseBatchSweep.hpp:1340` one here and leave `GenPCheck.hpp` for you to decide, or all of them, or none — say which.

---

BDonnot (commented):
Ok

PR #210, opened 2026-09-20
Report the batteries the distributed slack pushes past their limits

BDonnot:
| | added | removed |
|---|---|---|
| **C++ (src/core, src/bindings)** | +310 | -86 |
| **Python (package)** | +46 | -9 |
| **Tests** | +443 | -2 |
| **Docs + changelog** | +9 | -6 |
| total | +808 | -103 |

Implements the `[TODO]` entry on storage active power limits (removed from the changelog by this PR).

## Why

A storage unit takes a share of the distributed slack under the same rule as a generator (`SlackParticipation`, which `init_from_pypowsybl` drives from OpenLoadFlow's `activePowerControl` rule), and the slack is solved *inside* the Jacobian by participation factors that know nothing about limits. So a battery's converged active power is `target + its share of the imbalance` and can land beyond what the machine can deliver — exactly the statement `GenPCheck` already makes about a generator, and the condition OpenLoadFlow's `DistributedSlack` outer loop acts on. Until now there was nothing to compare it against.

## What

- `StorageContainer` gains the same optional `p_min_mw_` / `p_max_mw_` pair the generators have had since binary format 7: `LSGrid::set_storage_p_limits`, `StorageInfo.min_p_mw` / `max_p_mw`, NaN meaning "no limit for that one", two empty vectors to drop them again.
- `_aux_add_storage` reads an IIDM battery's `min_p` / `max_p` into it.
- New `ViolationElementType::STORAGE`; `compute_physical_violations` reports `LOW_P` / `HIGH_P` on it, category `PHYSICAL`. Detection only, as for the generators — nothing is clamped and nothing is re-solved.

### The convention

The limits, and the reported `value`, are in the **generator** convention (what the unit can inject) — matching the container's own `min_q_mvar` / `max_q_mvar` and IIDM's own `min_p` / `max_p`, and therefore the opposite way round from the load-convention `target_p_mw` / `res_p_mw` this library stores for a storage unit. That trap is documented at every entry point and pinned by a test in each suite (`test_the_load_convention_is_not_used`, "the limits are read in the GENERATOR convention").

### Not a second check

The two families share one slack, so the share a machine takes is a fraction of the RAW participation of its bus, both families included. `GenPCheck` therefore walks generators and storage units together rather than gaining a sibling.

## Bug fixed along the way

That same reasoning uncovers a real bug in the existing generator check: it recovers a bus' raw participation as `w_norm(bus) * total_raw_weight`, but summed the **generators only**, while `bus_slack_weight_` sums both families. A generator sharing the slack with a battery was reported at roughly twice the share it actually took. Covered by `test_a_generator_sharing_with_a_battery_keeps_its_own_share`, which pins both values against `ac_pf`.

Also: `GeneratorContainer::init_full` now drops the limits it replaces, as the storage one does. Re-initialising with a different number of machines used to leave a vector sized for the previous ones.

## Breaking

`BINARY_FORMAT_VERSION` 9 -> 10: a storage unit's optional active power limits are part of its state, so a file written by an earlier version no longer loads. The committed fixture was regenerated (`case14_sandbox_format10.lsb`).

## Testing

- C++ (Catch2): 352/352 pass, including 3 new `[storage]` cases. Headers syntax-checked at C++14, 17, 20 and 23.
- Python: `test_physical_violations` (8 new cases), `test_storage_pypowsybl` (3 new), `test_binary_serialization`, `test_LSGrid_pypowsybl`, `test_init_from_pypowsybl`, `test_olf_bake`, `test_ContingencyAnalysis_limit_violations`, `test_storage_distributed_slack`, `test_pickleable` — all pass. `test_deepcopy` errors on a missing `gymnasium` in this environment, unrelated to the change. The full suite was not run (it takes hours; that is CI's job).

## Deliberately out of scope

The pandapower and PowerModels storage converters. pandapower's `net.storage` min/max are in the *load* convention, so threading them through needs a negate-and-swap plus a decision that did not belong in a change scoped to pypowsybl. Easy follow-up.

https://claude.ai/code/session_018i88HyUF2hi4uVQEiFoBnW

---
_Generated by [Claude Code](https://claude.ai/code/session_018i88HyUF2hi4uVQEiFoBnW)_


BDonnot (commented):
ok

PR #212, opened 2026-09-22
Update line / trafo physical parameters in place, and keep the analysis

BDonnot:
Add LSGrid.update_powerlines_parameters(r, x, h_or, h_ex) and LSGrid.update_trafos_parameters(r, x, b): replace the series impedance and the charging admittance of ALL the powerlines / transformers (pu) without touching the topology, statuses, tap ratio or phase shift. The shared implementation is BranchContainer::update_physical_parameters, which recomputes the model coefficients (_update_model_coeffs) and calls tell_recompute_ybus. A transformer also resets its neutral base_r / base_x and flags the DC Sbus.

Any change flagged only with tell_recompute_ybus (these methods, but also change_ratio_trafo and change_shift_trafo) used to run the full rebuild path of the solvers, symbolic analysis included, although only the values of Ybus had changed:

- NRAlgo: need_recompute_ybus is no longer part of need_rebuild. The first iteration only refactorizes, as it already did for an injection change.
- BaseDCAlgo: need_recompute_ybus sets need_refactor_ instead of need_factorize_, unless the hvdc droop data changed.

A change of the sparsity pattern must be flagged by whoever makes it (ybus_change_sparsity_pattern, ybus_some_coeffs_zero, ...). This is checked by an assert in debug builds only (NRAlgo, BaseDCAlgo): in a release build an unflagged pattern change is not detected.

Tests: the two methods are added to the cache-reuse sweep (AC and DC, against a grid that never reuses anything); a test on get_linear_solver_stats checks that nb_analyze does not move after a parameter update, and a control checks that it does move when the sparsity pattern changes; a test in test_lsgrid compares an updated grid with one initialized with the same values.

Assisted-by: Claude Code (claude-sonnet-5)


BDonnot (commented):
ok

PR #213, opened 2026-09-22
Opt-in: keep the starting |V| at buses a voltage-control group regulates

BDonnot:
| | added | removed |
|---|---|---|
| **C++ (src/core, src/bindings)** | +174 | -11 |
| **Python (package)** | +54 | -0 |
| **Tests** | +174 | -0 |
| **Docs + changelog** | +14 | -0 |
| **Build / other** | +100 | -11 |
| total | +516 | -22 |

## Why

Before every solve, each regulated bus has its voltage magnitude set to its target. For a bus a voltage-control **group** regulates (remote regulation, a voltage-mode SVC, or shared control with a remote member) that magnitude is itself a Newton-Raphson unknown, held by the group's voltage row. Setting it up front is therefore only a starting-point choice, and one the step damping cannot soften.

On a real grid snapshot, a few small units remote-regulate a stiff 225 kV bus whose neighbours start at 1.0 pu. Setting that bus to its target made every `MaxVoltageChange` step tiny: lightsim2grid needed **57** iterations from a DC seed where OpenLoadFlow, with the same damping, needs **6**. OLF starts every |V| at 1 and solves `V − target = 0` as an equation.

## What

`LSGrid.set_keep_vinit_at_group_controlled_buses(True)` / `get_...`, **off by default**. With the option off, every code path is unchanged.

- With it on, the buses a group regulates keep the caller's magnitude, and the voltage row brings them to target, damped with the rest of the step. The **solution is the same**. On the snapshot above: 8 iterations from the plain `dc_pf` seed, 6 from a flat start, identical to OLF to 1e-5 pu.
- Ordinary PV buses, whose magnitude is fixed, are still set to their target.
- Implementation: `VoltageControlPlan::group_controlled_solver_buses` and a small `seed_vm_keeping` helper. It runs the existing `set_vm` calls unchanged, then puts the saved value back at those buses. It is applied in `LSGrid::_build_into_cache` (`ac_pf`, `dc_pf`, `build_solver_input`) and at the four batch seeding sites (n solve, grid-cache reuse, per-row `gen_v`, DC magnitude reconstruction).
- The option is copied with the grid, so a batch inherits it. The batch property `keep_vinit_at_group_controlled_buses` (C++ bindings and the `TimeSerie` / `ContingencyAnalysis` / `ScenarioSweep` wrappers) changes the batch's own copy and drops its kept base case.
- It is not part of `get_state` or the binary format, like `compute_results_`.

## Also in this PR

- `examples/dist_slack_algorithm` no longer compiled against the core. `NRAlgo` needs `LinearSolverPolicy<LinearSolver>`.
- The CI job `test_plugin_against_installed` now builds and smoke-tests all three plugin examples (`external_algorithm`, `lm_algorithm`, `dist_slack_algorithm`), not just the first. `dist_slack_algorithm` gets a `test_plugin.py` matching the other two. `lm_algorithm` and `dist_slack_algorithm` have not been built on Windows or macOS before, so this run is their first there.

## Tests

- New `test_keep_vinit_group_controlled.py` (10 tests):
  - default value and `copy()`;
  - the starting magnitude, observed through `dc_pf` and through a DC contingency analysis;
  - PV buses unaffected;
  - identical AC / ContingencyAnalysis / TimeSeries results with the option on or off;
  - inheritance by the batches;
  - the kept base case dropped when the option changes.
- Rerun, all passing: `test_voltage_control_batch`, `test_voltage_control_remote_gen`, `test_voltage_control_svc`, `test_voltage_control_reclassify`, `test_remote_control_disconnected_target`, `test_ContingencyAnalysis_bindings`, `test_InjectionSweep`, `test_ScenarioSweep`, `test_timeseries_sbus`.
- C++ Catch2 suite: 352/352.

The branch was cut before `efb7c29` and the merge of #212 on `dev_1.1.1`. It merges cleanly (four files auto-merge), but the local test runs above were done without those two commits.


BDonnot (commented):
ok

